PR #35, opened 2015-12-28
Added extension version detection

nfsmaniac:
Read description to b6bb335


gera2ld:
Inline script should not be used in this project, so I add the version detect script in the `render` part. Thanks for your idea.

PR #145, opened 2017-06-22
Fix multi-line options in side menu of Dashboard

bershan2:
This issue popped up in Russian translation of Violentmonkey, so screenshots are in Russian:
violentmonkey

The problem arises from incorrectly used `line-height` to set the distance between options: `line-height` makes it look like the two-line options are in fact different options. `padding-top` and `padding-bottom` give the same distance between the menus, but keep default distance between lines of the same menu.

### DIFF
--- a/src/options/style.css
+++ b/src/options/style.css
@@ -147,7 +147,8 @@ aside img {
 }
 .sidemenu > a {
   display: block;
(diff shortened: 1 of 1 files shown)


gera2ld:
Thanks

PR #150, opened 2017-07-01
[WIP] Icons on buttons in scripts list

bershan2:
# PLEASE DO NOT MERGE THIS IN YET!
**I opened this pull request for discussion** instead of opening an issue to (1) keep discussion and code changes in one place and (2) avoid opening a duplicate issue that would point to a pull request that I would open later.

Hi, I think it is a great idea to use icons on the buttons instead of text. However, I think this should be done consistently. I tried to replace the buttons on the items in the list of scripts, but I'm not sure which icons you want to see there. Thus I have a few questions:
- Which icons do you want to see there?
  - "Edit" button:
     Do you want to use the Cog (cog.svg)? I feel like a [pencil](http://www.flaticon.com/free-icon/pencil_25688#term=pencil&page=1&position=2) or a [pair of brackets](http://www.flaticon.com/free-icon/code_25185) from the same icon set would look better...
 - "Disable"/"Enable" button:
    Do you want the cross (remove.svg) as "Disabled" and check mark (check.svg) as "Enabled", like on the popup interface? Also, 
 - "Remove" button:
    If cross (remove.svg) is taken for "Disabled", there is no icon for "Remove". How about a [trashcan](http://www.flaticon.com/free-icon/trash_25214#term=trash&page=1&position=1) from the same icon set?

Currently the commit is in its very early stages, so there is nothing interesting.

### DIFF
--- a/src/options/views/script-item.vue
+++ b/src/options/views/script-item.vue
@@ -2,8 +2,10 @@
   <div class="script" :class="{disabled:!script.enabled}" draggable="true" @dragstart.prevent="onDragStart">
     <img class="script-icon" :src="safeIcon">
     <div class="script-info flex">
-      <a class="script-name ellipsis" target="_blank" :href="homepageURL"
-      v-text="script.custom.name || getLocaleString('name')"></a>
+      <a class="script-name ellipsis" v-text="script.custom.name || getLocaleString('name')"></a>
+      <a class="script-support" v-if="homepageURL" target=_blank :href="homepageURL">
+        <icon name="question"></icon>
+      </a>
       <a class="script-support" v-if="script.meta.supportURL" target=_blank :href="script.meta.supportURL">
         <icon name="question"></icon>
       </a>


gera2ld:
Great work!
Actually `<tooltip>` component was added for this purpose. I have already changed some of the button labels into icons, e.g. `New`, `Check updates for All`.

### Add SVG icons
It is OK to add more icons, if only they are pure colored SVG and do not use `transform` attributes.
1. add svg files to `./icons`.
2. run `npm run svgo` to minify it.

We can paste it here for discussion before adding to the project.

### About the icons
1. I think the *pair of brackets* looks good for edit button.
1. How about a [toggle icon](http://fontawesome.io/icon/toggle-on/) for the enable / disable button?
1. Trash can is OK for remove button.

bershan2:
> <tooltip> component was added for this purpose.

That's what I figured and used.

>  I have already changed some of the button labels into icons, e.g. New, Check updates for All.

I made sure to pull in all changes from your repo before editing anything (I always do: that simplifies merging later). Your changes actually inspired me to do this (in the name of consistency).

Thanks for info on `npm run svgo`!

Also, Font Awesome is under SIL OFL 1.1 license[1], right? And that license requires ["Attribution"](http://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL#2855b85f). As far as I understand, it's OK to use under MIT as long as there is a clear indication that they are derived from Font Awesome. Do you think it is a good idea to include such notice somewhere?

[1] Footer of <http://fontawesome.io/> says:
> Font Awesome licensed under [SIL OFL 1.1](http://scripts.sil.org/OFL)

bershan2:
Current status:
untitled

Icons look too small, I think. I'd like to remove the border around them (in CSS) and make them actually occupy the entire `viewBox` (in SVG)? That would reduce clutter and give them more space.

There are two trash cans: [the lighter one I mentioned earlier](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/trash-o.svg) and [the darker one](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/trash.svg). Which one do you like? (I used the lighter one because I like it more, but it's your call.)

OK, I guess here are the new sources then:
 - [Togge ON](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/toggle-on.svg)
 - [Togge OFF](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/toggle-off.svg)
 - [A pair of brackets](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/code.svg)
 - Which trashcan?

My earlier statement about Font Awesome being under SIL OFL 1.1 was false: the project I linked to above (where I got the new icons) is under MIT.

For reference, here the links to other icons:
 - [Cog](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/cog.svg)
 - [Check mark](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/check.svg) (is it relevant anymore?)
 - [Plus](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/plus.svg)
 - [Refresh](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/refresh.svg)
 - "remove.svg" (is it relevant anymore?)
 - [Search](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/search.svg)
 - "Question" and "More" -- did not find, see comment below 

I think current question mark and "more" arrow are too sharp for all other icons, how about replacing them with [this question mark](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/question-circle.svg) and [this arrow](https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/black/svg/chevron-right.svg)?

gera2ld:
I'm not sure whether it's a good idea to replace the check and cross icons in popup menu with toggles. Is it too complexed to display a toggle for each script?
Maybe we could just use font-awesome icons if there is no licence concern.

bershan2:
### The popup
So would you like toggles or check and cross for the individual scripts on popup? Currently they are toggles because they are the same as global "scripts enabled/disabled" icons, but I can make them different in no time. For reference, here is the current look:
untitled
The only thing I do not like is the small Cog icon because "cog.svg" has a large `viewbox` comparing to the actual image.
### The Dashboard
I used some "Custom CSS" rules (see below) and got this:
untitled

```
/* To fix resizing of "Blacklist" and "Custom Style" text boxes */
textarea {
    min-height: 10em; /* min-height instead of height, is for Firefox; Chrome treated height abs min-height */
    resize: vertical; /* to restrict horizontal resize */
}

.button-icon{
    width:  2rem;
    height: 2rem;
    border: none;
    border-radius: 50%;
    background: none;
}

.button-icon:hover{
    background: #c0c0c0
}

.icon {
    width: 1rem;
    height: 1rem;
    vertical-align: middle
}

button, script, textarea, .export-list, .ellipsis{
border-radius: .3rem;
}

/* To prevent cursor from changing to text look on non-interactive writings */
p,h1,h2,h3 {
    cursor: default
}

/* Clearly indicate clickable elements: 
   - the checkbox labels, and
   - tab switchers on Edit */
label, .edit-nav{
    cursor:pointer;
}

.edit-header>* {
    cursor:auto;
}

.edit-header>a {
   cursor: alias;
}
```

bershan2:
### Off topic, but you know that Firefox addon store has version 2.6.3 and Opera 2.5.7, right?

### I post stuff below mostly for myself, but your suggestions are very welcome.
This is a laundry list of stuff related to UI in general:
 - For every script in the list, "Update" button should appear if and only if `@downloadURL` is set
 - The font is not set
   - This is OK in Chrome because Chrome default font is pretty
     - Changing Chrome font settings does not appear to have any effect, even [advanced settings](https://chrome.google.com/webstore/detail/advanced-font-settings/caclkomlalccbpcdllchkeecicepbmbm)
	   - Settings -> Apperance -> Fonts
	   - `chrome://settings/appearance`
   - This is bad in Firefox because its default Times New Roman is ugly (out of style)
     - Firefox font settings are under 
	   - "Options" -> "Content" tab -> "Fonts & Colors"
	   - `about:preferences#content`
 - The tooltips should be descriptive, the concise button titles are not proper tooltips
   - Create another message field and copy the button names translation for now?
     - Is anyone actively translating stuff and would take advantage of this?
     - Is Transifex useful? It is definitely out of date...

gera2ld:
Review on Firefox takes quite long time, I've submitted v2.7.1 for some time.
Opera reviewers have messaged me that they have no time to review the new version since it is quite different from the last one.

- > For every script in the list, "Update" button should appear if and only if @downloadURL is set

  This is almost the current behavior. The only difference is that `@downloadURL` will fallback to the last installed URL if not set. FYI, GreasyFork will remove the `@downloadURL` and force script managers to update from the last installed URL.

- > The font is not set

  Should be fixed.

- > The tooltips should be descriptive

  Agreed. Currently some people help translation with Transifex and some with GitHub PRs. I think we can update the English version and leave other locales to their users.
  In case you'd like to add some messages, you might use `i18n('someKey')` in `.js` or `.vue` files and run `npm run i18n`. But for now I think we should deal with the icons first.

PS. I like the borderless style, it looks much better than before.

bershan2:
> we should deal with the icons first.

I agree, I posted the rest for later.

Since this is a style thing, I can not choose which is best, I can only propose some options (after all, it is your project). Thus, should I make a patch moving my "Custom CSS" to the default style? Or would you like me to change button size (they are a bit taller than they used to be) or `:hover` colors a bit?

gera2ld:
Hi, I am going to change the layout to make use of full screen.
Also I've updated the icons and use them almost as you suggested. Thanks a lot for your great suggestions. Now it looks like this:
vm

Would you like to add your custom style as a theme?

gera2ld:
Closing since we have updated the UI this way already. Thanks a lot.

PR #190, opened 2017-08-25
blob url should be allowed in GM_openInTab

tiansh:
After I confused by `data:` url not supported by Firefox. I found out that `blob:` url is the best choice for me. But it seems broken in VM on Chrome. It seems that this commit should fix it. Hope it can be published asap, since one of my script is relaying on this feature. many thx.


tiansh:
so sad. just noticed that `blob:` url is also not supported by firefox. never minde.

PR #197, opened 2017-09-03
Update for German locale

spodermenpls:
A couple of small tweaks, after seeing my translation in action the first time.

### DIFF
--- a/src/_locales/de/messages.yml
+++ b/src/_locales/de/messages.yml
@@ -68,7 +68,7 @@ msgLoadedData:
   description: >-
     Message shown in the confirm page when a javascript file to be installed is
     loaded.
-  message: Skriptdaten geladen
+  message: Skriptdaten geladen.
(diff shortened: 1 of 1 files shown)


src/_locales/de/messages.yml
```diff
@@ -148,7 +148,7 @@ labelAuthor:
   message: 'Autor: '
 anchorAuthor:
   description: Author shown on about tab.
-  message: '<a href=mailto:i@gerald.top>Gerald</a>'
+  message: ''
```
gera2ld: Why is this removed?

spodermenpls: Your adress and name needs no translation, and by delegating this field back to the English locale, a change there gets displayed in the German version without having to edit its locale.

gera2ld: OK, I am just curious since it makes no change to the result. 😄

PR #204, opened 2017-09-08
Traditional Chinese (Taiwan) translation

ngoo5tsong5sik:
Added Traditional Chinese (Taiwan).
轉為繁體中文並修改為臺灣用語。


gera2ld:
According to [the document](https://developer.chrome.com/webstore/i18n?csw=1#localeTable), `zh-TW` should be renamed to `zh_TW`.

PR #246, opened 2017-10-22
Change injection technique to use src with Blob

### LINKED ISSUE #173: Firefox CSP issue
Firefox does not support script injection from extension when page has CSP restrictions. As a result, no script can be injected to such pages in Firefox. But it works well with Chrome.

Related issues:
#107, #172 

Some examples suffered from this issue:
- scripts cannot work correctly on Github
- styles installed via userscript cannot be applied on Github

evilpie:
This fixes #173. And makes Violentmonkey work on recent builds of Firefox 58 on pages with CSP.

### DIFF
--- a/src/injected/utils.js
+++ b/src/injected/utils.js
@@ -12,9 +12,12 @@ export function postData(destId, data) {
 export function inject(code) {
   const script = document.createElement('script');
   const doc = document.body || document.documentElement;
-  script.textContent = code;
+  const blob = new Blob([code], { type: 'text/javascript' });
+  const url = URL.createObjectURL(blob);
+  script.src = url;
   doc.appendChild(script);
   try {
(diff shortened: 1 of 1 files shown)


eight04:
Would this work on Firefox 57?

gera2ld:
That's great! Could you fix the lint errors? Maybe we should add `URL.revokeObjectURL` after using it.

evilpie:
@gera2ld Fixed. We should definitely use revokeObjectURL to prevent leaks.

gera2ld:
This has brought up a new problem, which breaks VM web page listeners on Chrome stable Win 10. But works on Chrome any Mac OS X, and Chrome canary Win 10. I think it's a bug of Chrome on Windows but should be fixed. See #249 .

PR #251, opened 2017-10-27
fix: avoid page jumps when setClipboard in Firefox

yfdyh000:
User reports from https://bbs.kafan.cn/thread-2106491-5-1.html.
Testcase:
```
// ==UserScript==
// @name New Script
// @namespace Violentmonkey Scripts
// @match *://*/*
// @grant					GM_setClipboard
// ==/UserScript==
GM_setClipboard("test", "text");
```

Ref: https://github.com/mdn/webextensions-examples/blob/master/context-menu-copy-link-with-types/clipboard-helper.js. I hope this code does not have a license issue.

I don't understand the usage of `src/background/utils/clipboard.js`, so there is no test.

### DIFF
--- a/src/background/utils/clipboard.js
+++ b/src/background/utils/clipboard.js
@@ -1,17 +1,14 @@
-const textarea = document.createElement('textarea');
-document.body.appendChild(textarea);
-
 let clipboardData;
 function onCopy(e) {
+  document.removeEventListener("copy", onCopy, true);
+  e.stopImmediatePropagation();
   e.preventDefault();
   const { type, data } = clipboardData;
   e.clipboardData.setData(type || 'text/plain', data);
 }
-document.addEventListener('copy', onCopy, false);
-
 export default function setClipboard(data) {
   clipboardData = data;
-  textarea.focus();
+  document.addEventListener("copy", onCopy, true);
   const ret = document.execCommand('copy', false, null);
   if (!ret && process.env.DEBUG) {
     console.warn('Copy failed!');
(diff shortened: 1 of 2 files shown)


src/background/utils/clipboard.js
```diff
@@ -1,17 +1,14 @@
-const textarea = document.createElement('textarea');
-document.body.appendChild(textarea);
-
 let clipboardData;
 function onCopy(e) {
+  document.removeEventListener("copy", onCopy, true);
+  e.stopImmediatePropagation();
```
gera2ld: This module is used by browsers other than Firefox, so please don't change it.

src/injected/content/clipboard.js
```diff
@@ -1,29 +1,17 @@
...
 }
 
 export default function setClipboard({ type, data }) {
   clipboardData = { type, data };
-  if (!textarea) init();
-  document.addEventListener('copy', onCopy, false);
-  document.documentElement.appendChild(textarea);
-  textarea.focus();
+  document.addEventListener("copy", onCopy, true);
   const ret = document.execCommand('copy', false, null);
```
gera2ld: If `copy` event fails to emit, `onCopy` will never be called, and the listener will never be removed from `document`.

src/injected/content/clipboard.js
```diff
@@ -1,29 +1,17 @@
...
-  textarea = document.createElement('textarea');
-  textarea.style.position = 'absolute';
-  textarea.style.width = 0;
-  textarea.style.height = 0;
-  textarea.style.left = '-20px';
-}
-
 function onCopy(e) {
+  document.removeEventListener("copy", onCopy, true);
+  e.stopImmediatePropagation();
```
gera2ld: This is the key, I think. Just adding this line and removing the `<textarea>` should make it work as expected.

PR #256, opened 2017-10-30
Skip injection of injected-web if there are no scripts to run

andersk:
If we’re not going to run any scripts in the webpage, we may as well not inject the initializer either, right? This saves a bit of work, and avoids unnecessary Firefox CSP errors, on pages with no matched scripts.

### DIFF
--- a/src/injected/content/index.js
+++ b/src/injected/content/index.js
@@ -54,20 +54,27 @@ export default function initialize(contentId, webId) {
     if (handle) handle(req.data, src);
   });
 
-  sendMessage({ cmd: 'GetInjected', data: window.location.href })
+  return sendMessage({ cmd: 'GetInjected', data: window.location.href })
   .then(data => {
+    let needInject = false;
     if (data.scripts) {
       data.scripts.forEach(script => {
         ids.push(script.props.id);
-        if (script.config.enabled) badge.number += 1;
+        if (script.config.enabled) {
+          badge.number += 1;
+          needInject = true;
+        }
       });
     }
-    bridge.ready.then(() => {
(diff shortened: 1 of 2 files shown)


gera2ld:
Thanks @andersk . Since v2.8.15 there should be no wrong CSP errors caused by Violentmonkey.

This patch breaks `exposeVM` [here](https://github.com/violentmonkey/violentmonkey/blob/73f8299388ae1791c2326a1d2955fa486ee2ff81/src/injected/web/index.js#L483). It's used to expose an API to sites like <https://greasyfork.org> to check  whether a script has been installed, with the help of [this script](https://github.com/violentmonkey/vm-detector). So it should be run even if there is no script to be injected. However, such sites can be whitelisted so it's not a big problem. I will add a whitelist for it in the content script before merging this.

andersk:
Oh, interesting. Couldn’t the `exposeVM` functionality be handled through the normal `@grant` mechanism rather than being based on a hostname whitelist?

There are still CSP errors in Firefox 57 (beta). I assumed this is known because the fixes refer to bugs fixed in Firefox 58 (nightly).

andersk:
Wait, would this patch actually break vm-detector? vm-detector is a userscript with `@match https://greasyfork.org/*`, so if it is installed, the initializer will still be injected on greasyfork.org and `exposeVM` will work as before.

I think the only thing that would break is if greasyfork.org itself tried to use the `Violentmonkey` object without a userscript involved, which AIUI it doesn’t.

gera2ld:
That's true... 👍  I'll leave it this way. Thanks a lot.

PR #357, opened 2018-02-23
Create icon.svg

perdolka:


perdolka:
1.56 KB

perdolka:
I thought it was better to have one tiny file for all sizes.
Are SVG converted to PNG for compatibility?

gera2ld:
The PNG icons are required by the extension, to show in different places. The browsers do not support SVG icons in such cases.

Martii:
@gera2ld 
> The PNG icons are required by browsers, to show in different places. Browsers do not support SVG icons in such cases.

This might be the case in Opera Presto, if I recall correctly, however most current browsers **do** support SVGs. Even IE11 does and that will fall off the map at some point.

Anyhow... it would be helpful for SEO and marketing of your product to have this available. Right now we use a rastered version on OUJS that doesn't look all that great when zoomed in or out, i.e. browser sizing of raster images leaves something to be desired which can affect negatively or positively user interest.

If you would share it on the repo we would be more than happy to utilize the actual SVG for free marketing on OUJS within the appropriate safeguards of GH sanitizing at the moment.

Thank you for the official reconsideration please,
OUJS Admin

gera2ld:
@Martii I mean the web extension requires PNG icons to be packed with them, and show in places like toolbar.
It's OK to add it to the repo, maybe a few days later because I don't have it here at the moment.

PR #363, opened 2018-03-02
Switch from Codemirror to Monaco

Ambyjkl:
Also, made some tweaks to performance by avoiding string copying on v-model and string comparisons and using monaco's getAlternateVersionId to track changes


gera2ld:
Monaco looks amazing but it's extremely large (10MB after being compressed) to be packed into an extension. I tried building with Webpack but it failed out of memory.

ruslankonev:
@tophf you can use Monaco from CDN, you no need to fetch it in the extension bundle
https://www.google.com/search?q=use+monaco+editor+from+cdn

tophf:
No, running external code is forbidden by extension galleries.

ArjixWasTaken:
> No, running external code is forbidden by extension galleries.

Sorry to revive this dead thread, but there is no need to load it from the CDN at run-time, it could be done when packaging the extension.

PR #386, opened 2018-04-04
Update for Portuguese Brazilian locale

KingThunder:
Based on PT-EU locale, with some changes.


src/_locales/pt_BR/messages.yml
```diff
@@ -139,7 +139,7 @@ filterExecutionOrder:
   message: ordem de execução
 hintInputURL:
   description: Hint for a prompt box to input URL of a user script.
-  message: 'Coloque o URL:'
+  message: Coloque o URL:
```
gera2ld: The quotes here cannot be removed, otherwise syntax error occurs.

PR #390, opened 2018-04-09
ru locale update

tenno-seremel:
I'm assuming here that "Value (serialized as JSON)" means you must supply a JSON text in the value field.


gera2ld:
> I'm assuming here that "Value (serialized as JSON)" means you must supply a JSON text in the value field.

Exactly

PR #438, opened 2018-08-30
Update French translation

ariasuni:
I fixed a lot of mistakes or missing strings.

Some (two) strings were not translatable so I fixed that.

I couldn’t find a way to make the popup size adaptable, so I increased its size a bit, see before/after:
screenshot_20180830_131131screenshot_20180830_131210


gera2ld:
Thanks a lot for your work. I have just updated the style a bit so that we don't have to increase the size of panel. Could you please rebase on the latest commit?

ariasuni:
I couldn’t figure out how to do it properly, nice!

PR #450, opened 2018-09-12
Update Japanese translation

ScratchBuild:


iop369247:
么么么哒

gera2ld:
There are a few new entries now, would you like to update them by the way? Thx.

PR #475, opened 2018-11-09
Update Slovak translation

kubalav:


.gitignore
```diff
@@ -2,3 +2,4 @@
 node_modules/
 *.log
 /stats.json
+*.bak
```
gera2ld: Why do we need this?

kubalav: I'm sorry. It not necessary.

PR #504, opened 2019-01-24
Add support for modifying Sec- headers

### LINKED ISSUE #503: Support modifying Sec-* headers
For example, modifying Sec-Metadata, which provides information to prevent CSRF. I need to do a call that forges a "same-site" Sec-Metadata header, but unfortunately, [this is blocked by design](https://mikewest.github.io/sec-metadata/#sec-prefix), similar to how Referer is blocked.

qsniyg:
Fixes #503

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -13,6 +13,9 @@ const specialHeaders = [
   'host',
   'cookie',
 ];
+const specialHeaderPrefixes = [
+  'sec-',
+];
 // const tasks = {};
 
 export function getRequestId() {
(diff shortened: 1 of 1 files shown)


src/background/utils/requests.js
```diff
@@ -68,6 +71,11 @@ function xhrCallbackWrapper(req) {
   };
 }
 
+function isSpecialHeader(lowerHeader) {
+  return specialHeaders.includes(lowerHeader)
+    || specialHeaderPrefixes.includes(lowerHeader.replace(/-.*/, '-'));
```
gera2ld: How about using `startsWith` instead of regular expression since comparison between plain text is much lighter?

qsniyg: Fixed. The reason I did it that way was to enable multiple prefixes in theory via `specialHeaderPrefixes`, but you're right, it could just as easily be added as an extra line to the function.

Issue #511, opened 2019-02-10
no update found if `@version` does not exist


divinity76:
### What is the problem?
the "check for updates" button always say "no update found" for scrips installed from file:// URI's.

### How to reproduce it?
1. make a script in C:\foo.js
2. install script from the uri file:///C:/foo.js
3. change the script in C:\foo.js
4. press the "check for updates" button

### What is the expected result?
the script is updated.

### What is the actual result?
"No update found."

### Environment
- Browser: Google Chrome
- Browser version:  Version 72.0.3626.96 (Official Build) (64-bit)
Google Chrome is up to date (2019-02-10)
- Violentmonkey version: 2.10.1
- OS: Windows 7 Ultimate SP1 X64

divinity76:
err, after some investigation, turns out this only happens if the script does not have a @version tag..
not sure if this is a bug or intended behavior then, but IMO, if @version metadata is absent, the version detection *should* fall back to a full byte comparison of the source, instead of assuming that the code is unchanged (as seem to be the case right now)

gera2ld:
This is by design. I will update the documentation later.

Only if `@version` exists, the script will update automatically. Otherwise scripts without `@version` will always be reinstalled, causing side effects like unexpected increase of installation times.

In your case, either reinstall the script manually or develop it like [this](https://violentmonkey.github.io/2017/03/14/How-to-edit-scripts-with-your-favorite-editor/).

divinity76:
oh i see, thanks for explaining.

PR #512, opened 2019-02-10
consider undefined @Version to be different

divinity76:
was causing scripts lacking @Version metadata to never get updated, see https://github.com/violentmonkey/violentmonkey/issues/511 for more info.


divinity76:
https://github.com/violentmonkey/violentmonkey/issues/511#issuecomment-462210134

PR #536, opened 2019-03-24
Update Romanian translation

cristisilaghi:


src/_locales/ro/messages.yml
```diff
@@ -3,20 +3,22 @@ anchorAuthor:
   message: '<a href=mailto:i@gerald.top>Gerald</a>'
 anchorTranslator:
   description: Translator shown on about tab.
-  message: '<a href="http://www.forumit.ro/">Nicolae Crefelean</a>'
+  message: >-
+    <a href="http://www.forumit.ro/">Nicolae Crefelean</a>, <a 
+    href=mailto:cristian.silaghi@mozilla.ro>Cristian Silaghi</a>'
```
gera2ld: There seems to be an extra single quote.

cristisilaghi:
It should be OK now.

PR #546, opened 2019-04-15
Define default text colors in CSS

jakseb:
So that we do not rely on unknown browser defaults and avoid a situation in which text has a similar color as the background (which has its color already defined in a stylesheet).


gera2ld:
Thanks 👍

PR #564, opened 2019-05-11
Fix grammar of labelNoScripts.message string

roryokane:
When there are zero of something, you refer to it with the plural.

Also, I replaced the straight single quote with the actual apostrophe character (which is the same character as a closing curly single quote).

### Before

1 before

### After

This is just a mockup made with the Inspector. I didn’t actually rebuild the extension.

2 after (mockup)


gera2ld:
Thanks for the explanation. 👍

PR #612, opened 2019-09-25
preload CustomCSS to avoid FOUC and white flash during page load

tophf:
Fixes the white flash and FOUC (flash of unstyled content) during page load to enable the use of `customCSS` option in ViolentMonkey with a dark theme without being blinded by the flash.

Here's the [theme itself](https://gist.github.com/tophf/cc109b0ba5cb98353c9ea51579c85ae2) (very early WIP I mocked in devtools).

Fixed by putting scripts into `<head>` in the webpack's html template, see https://github.com/violentmonkey/violentmonkey/pull/612#issuecomment-536225835 for more info.

<details><summary>outdated initial take</summary>

This is the first time I deal with WebPack/Vue so don't freak out too much if I messed up something. At least `yarn build` completes successfully!

This solution uses a small separate CSS preloader that directly accesses the background page (if possible) that exposes `getOption` function. It may seem hackish but I see no other way around it after trying various things. Looks like the main code takes too long to initialize so Chrome's renderer loses patience and draws an intermediate white frame.

Added a fixup - do nothing if customCSS is not set.~~
</details>

image

With this PR the intermediate frame is styled; no FOUC occurs:

image

### DIFF
--- a/src/popup/index.js
+++ b/src/popup/index.js
@@ -10,12 +10,13 @@ tld.initTLD();
 
 Vue.prototype.i18n = i18n;
 
-const el = document.createElement('div');
-document.body.appendChild(el);
-new Vue({
-  render: h => h(App),
-})
-.$mount(el);
+waitForBody()
+.then((body) => {
+  new Vue({
+    render: h => h(App),
+  })
+  .$mount(body.appendChild(document.createElement('div')));
+});
 
 Object.assign(handlers, {
   SetPopup(data, src) {
@@ -50,3 +51,14 @@ browser.tabs.query({ currentWindow: true, active: true })
     store.domain = tld.getDomain(domain) || domain;
   }
 });
+
+function waitForBody() {
+  return Promise.resolve(document.body || new Promise((resolve) => {
+    new MutationObserver(((_mutations, observer) => {
+      if (document.body) {
+        observer.disconnect();
+        resolve(document.body);
+      }
+    })).observe(document.documentElement, { childList: true });
+  }));
+}
--- a/scripts/plaid.conf.js
+++ b/scripts/plaid.conf.js
@@ -1,4 +1,5 @@
 const { isProd } = require('@gera2ld/plaid/util');
+const { defaultOptions } = require('@gera2ld/plaid-webpack/util');
 
 /**
  * For each entry, `key` is the chunk name, `value` has following properties:
@@ -59,4 +60,8 @@ exports.optimization = {
       },
     },
   },
+};
+exports.htmlOptions = {
+  ...defaultOptions.htmlOptions,
+  template: require.resolve('./template.html'),
 };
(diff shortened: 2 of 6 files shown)


src/common/preload-custom-css.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-var, func-names, vars-on-top */
```
gera2ld: ESLint is not supposed to be disabled. I think we should achieve this by a more modern way.

tophf: Yeah I agree but since I don't know anything about bundlers and how to make them compile the normal modern syntax into ES5, I had to write in ES5 manually, otherwise the minifier couldn't process the file.

tophf: BTW if you think this is really crucial you may want to add an option to disable disabling eslint via the inline comments.

gera2ld:
I think this implementation is kind of hacky.
I'll look into this one further to see if there is an easier way to avoid FOUC.

tophf:
FWIW I've tried a few proper methods and none worked because Vue/WebPack take too long to initialize from the viewpoint of Chrome's renderer.

tophf:
One of the attempts I've made was to get the customCSS value from getBackgroundPage() in `#/common/ui/style` but even that is way too late because Vue takes all that time to initialize itself during which the white frame is shown:

chrome_2019-09-28_21-22-10

As you can see the only solution is to apply CSS before Vue starts to initialize itself and AFAIK the only way to achieve that is to use webpack hooks the way it's done in this PR.

tophf:
Some news:
* Now it's compiled properly as an entry point in webpack so I don't need to modify gulpfile and disable eslint. It bloated the file by adding the standard module wrapper but there's seemingly no way to configure webpack to omit it because webpack is all about being modular (c).
* Even moving css-preloader.js as it is now from `<head>` into `<body>` produces a random very short FOUC so it's pretty final that you won't find a way to avoid FOUC while staying within Vue.

tophf:
More good news. Turns out if I move all app scripts into `<head>` they run before DOMContentLoaded and there's no FOUC even if I access the background page inside `ui/style`.

Now no separate files or entries needed. I've copied template.html from gera2ld/plaid-webpack and adjusted it accordingly.

I guess you wouldn't want to modify the upstream template as it'll require all dependent projects to make sure to wait for `<body>` either via MutationObserver (for example the way I've done it in `popup/index.js`) or an synchronous callback that runs in a new event loop cycle like any extensions API callback.

src/popup/index.js
```diff
@@ -50,3 +51,14 @@ browser.tabs.query({ currentWindow: true, active: true })
     store.domain = tld.getDomain(domain) || domain;
   }
 });
+
+function waitForBody() {
```
tophf: This function can be removed if Vue is created and mounted inside browser.tabs.query() callback.

tophf:
Closing in favor of #626.

PR #613, opened 2019-09-25
guard against an implicit global variable for <html id="browser">

tophf:
If a web page has `<html id="browser">` it'll create a global variable `browser` as per the [specification](https://html.spec.whatwg.org/multipage/window-object.html#named-access-on-the-window-object).

Violentmonkey needs to guard against it since browser.js also runs in a content script. This PR adds a check for the prototype just like [Mozilla's WebExtension polyfill](https://github.com/mozilla/webextension-polyfill/pull/153) does.

### DIFF
--- a/src/common/browser.js
+++ b/src/common/browser.js
@@ -123,7 +123,13 @@ const meta = {
   },
   webRequest: true,
 };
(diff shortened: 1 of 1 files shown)


gera2ld:
Good job, thanks!

PR #614, opened 2019-09-25
allow customCSS to reorder/hide disabled scripts in popup

tophf:
With this PR customCSS can reorder/hide disabled scripts in popup by setting `flex` on `.submenu` and `order` or `display` on `.submenu .disabled`.

Ideally I would like to implement this as a user option in the dashboard but you seem to prefer the minimalistic built-in set of options, right?

### DIFF
--- a/src/popup/style.css
+++ b/src/popup/style.css
@@ -37,7 +37,7 @@ footer {
     background: cornflowerblue;
     color: white;
   }
-  &.disabled {
+  .disabled & {
     color: gray;
     &:hover {
       color: silver;
--- a/src/popup/style.css
+++ b/src/popup/style.css
@@ -37,7 +37,7 @@ footer {
     background: cornflowerblue;
     color: white;
   }
-  .disabled & {
+  .disabled > & {
     color: gray;
     &:hover {
       color: silver;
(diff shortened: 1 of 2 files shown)


gera2ld:
The execution order of scripts can be resorted by dragging, which requires each script item to have the same height, and the display order must be the same as that in the JS list. I'm afraid this change will break the resorting thing.

BTW, currently we support sorting in alphabetical order and also support showing enabled scripts first.
image

It's OK to add a new option if it does not require the user to make a choice.

tophf:
This PR is for the popup, not the dashboard. 
The dashboard already puts the `.disabled` attribute on the entry itself so it's alright.

tophf:
Also, I don't see why moving the class name to its proper place on the entry worsens anything. AFAICT it only improves things and allows for easier customizing of the UI.

gera2ld:
My fault, I was looking into the wrong place.

> Also, I don't see why moving the class name to its proper place on the entry worsens anything. AFAICT it only improves things and allows for easier customizing of the UI.

This is reasonable.

src/popup/style.css
```diff
@@ -37,7 +37,7 @@ footer {
     background: cornflowerblue;
     color: white;
   }
-  &.disabled {
+  .disabled & {
```
gera2ld: Maybe stricter?
```css
.disabled > & {
```

tophf: Yep.

PR #615, opened 2019-09-25
improve clarity of icons

tophf:
Improves clarity of the small icon in the browser toolbar which was previously scaled by the browser from 19px (or 38px) to 16px so it was very blurry on a single-density display. 

The missing `w` version for the 32px icon is also added.

All three modified icons were compressed with `pngquant --quality 50-80`. BTW if applied to all other icons it would reduce them from ~25k to ~9k with no difference in quality if you're interested.

* Before:
  image

  It was especially vague with darker themes:
  image

* After - the icon was re-scaled from 128px source and the top 1px of the hair, which wasn't legible anyway, was sacrificed to improve the overall clarity, especially the smile and ears:
  image

  image


gera2ld:
I am going to put the original SVG in this repo and generate different size of icons with [Jimp](https://github.com/oliver-moran/jimp) in the future so that we don't need to deal with images much even if the preferred sizes are change.

I think we'd better not refine the images manually. Maybe we can modify the image to have a better look in different situations.

tophf:
This main application icon at 16px is a special case (close to Pixel Art images where every pixel is thoughtfully crafted) because its contents was not designed to be legible at that size unlike a typical icon. For example there are empty margins that I had to overcrop manually to increase the meaningful image area from 12x12px to 14x15px which produces a very big visual improvement. Can jimp be configured to do that?

gera2ld:
Jimp is able to crop image if we can give the rules.

PR #616, opened 2019-09-25
allow extra/missing spaces in meta: around // and start/end/keys

### LINKED ISSUE #610: Metadata parser cannot recognize a userscript
Press the new script button and paste this:
```js
// ==UserScript==//
// @name     a
// ==/UserScript==
```

Expected: The script is recognized and saved
Observed: `Invalid script` is displayed; it's not saved

---

As you can see there's a `//` after `==UserScript==` but VM incorrectly expects a whitespace there although other managers (TM, GM) don't. There's also no such requirement in the [GM wiki](https://wiki.greasespot.net/Metadata_Block).

P.S. On the other hand, [this regexp used by VM](https://github.com/violentmonkey/violentmonkey/blob/master/src/injected/web/index.js#L188) has `\/\/\s+==` which is too lax: it will allow linebreaks which is definitely wrong so `[\t\x20]+` should be used instead, AFAICT.

---

### Environment
- Browser: Chrome
- Browser version: 74, 79
- Violentmonkey version: 2.11.2
- OS: Windows

tophf:
Fixes #610.

I wonder if I should write some tests for the parser?  
If so, I'll probably need some help as I don't see an existing analog.

### DIFF
--- a/src/background/utils/script.js
+++ b/src/background/utils/script.js
@@ -48,15 +48,18 @@ export function parseMeta(code) {
     [key]: metaTypes[key].default(),
   }), {});
   let flag = -1;
-  code.replace(/(?:^|\n)\s*\/\/\s*([@=]\S+)(.*)/g, (_match, group1, group2) => {
-    if (flag < 0 && group1 === metaStart) {
+  // Allow metadata lines to start with SPACE? '//' SPACE?
+  // Allow anything to follow the predefined text of the metaStart/End
+  // The spaces must be on the same line so [\t\x20] is used as \s also matches \r\n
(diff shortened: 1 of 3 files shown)


gera2ld:
There are some tests in `test/background/script.test.js` for the meta block parser.

tophf:
Added a test.
I've used `{...objectSpread}` there - is it okay?
If tests aren't compiled into ES5 I should have probably kept Object.assign.

gera2ld:
> I've used `{...objectSpread}` there - is it okay?

It's OK to use ES6 in this project.

PR #618, opened 2019-09-26
Expose reason for empty script list via icon, popup, and tooltip

### LINKED ISSUE #542: [Feature request] Different icon for blacklisted sites
### What is the problem?
Icon on the bar is same when we have enabled blacklist or without it. Tapermonkey makes icon grey(pale). I just want to get some feedback to avoid dumb issues with filter and make sure that it doesn't work in online bank for example.

tophf:
New extension icon `b` is a grayed out version of `w` icon and is shown for:
* blacklisted URLs
* unsupported pages like the browser UI or that of another extension

New localizable strings shown in the extension icon tooltips for the affected states:
* >Blacklisted in Violentmonkey's settings
* >Violentmonkey cannot run userscripts in this page<br>(common examples: browser UI or another extension)

The above text is also shown in the popup in place of the script list as well as the old "Scripts disabled" message when applicable.

Fixes #542

image

image

image

/me braces myself as there will be probably a lot of changes during the review...


src/background/utils/tester.js
```diff
@@ -11,6 +11,10 @@ hookOptions((changes) => {
 });
 const RE_HTTP_OR_HTTPS = /^https?$/i;
 
+let bltCache;
```
tophf: somehow I thought it was for BlackListTest but it's actually TestBlackList so now I'm not sure how to rename it because tbl is commonly used for tables...

src/background/utils/message.js
```diff
@@ -11,15 +11,26 @@ export function notify(options) {
...
+export function forEachTab(callback) {
+  // we'll run the callback for each tab in a separate event loop cycle
+  // because hundreds of tabs would make our extension process unresponsive,
+  // the same process used by our own pages like the background page, dashboard, or popups
+  browser.tabs.query({})
+  .then(function throttle(tabs) {
+    const tab = tabs.shift();
+    if (tab) {
+      callback(tab);
+      setTimeout(throttle, 0, tabs);
```
gera2ld: Is this a bit too slow? AFAIK `setTimeout(..., 0)` is actually a delay of 1/60s, so it takes 1s for a broadcast to 60 tabs?

src/popup/views/app.vue
```diff
@@ -127,6 +134,22 @@ export default {
         data: script,
       }));
     },
+    failureReason() {
```
gera2ld: What's this for? I didn't find a place using it.

src/background/index.js
```diff
@@ -248,66 +237,5 @@ initialize()
   sync.initialize();
   resetBlacklist();
   autoCheckRemove();
+  global.dispatchEvent(new Event('backgroundInitialized'));
```
gera2ld: How about calling the function directly?

src/background/utils/message.js
tophf: Yeah, it's best to do in batches of say 10?

src/popup/views/app.vue
tophf: At the start of the template - `:data-failure-reason="failureReason">` which enables custom styling of the entire app depending on the state.

src/background/index.js
tophf: I didn't want to make the parent context do the child's work explicitly. Event-driven approach would allow multiple children to subscribe. WDYT?

src/background/utils/message.js
gera2ld: Make sense.
BTW, I think a loop with `await delay(0);` in it might be more readable.

src/popup/views/app.vue
gera2ld: Do you mean that users can apply different styles depending on the state? I thought you were writing CSS for it and didn't find any. 😅

tophf: Yep, it's for customCSS. We did it for all extension pages in https://github.com/openstyles/stylus

src/background/utils/message.js
tophf: Ah indeed!

src/background/index.js
gera2ld: Sounds convincing.

src/background/utils/message.js
```diff
@@ -1,6 +1,4 @@
-import { i18n, defaultImage, noop } from '#/common';
-
-const THROTTLED_BATCH_SIZE = 20;
+import { defaultImage, i18n, noop } from '#/common';
```
tophf: Turns out my IDE was sorting the imports automatically.
I've disabled it now.

PR #619, opened 2019-09-27
reuse meta parser regexp and require a single space after //

tophf:
* `//` + single ASCII space is apparently hardcoded in almost every userscript-related app/site as noted in the discussion that ensued in #610 after the initial PR relaxed the parser
* we'll still allow whitespace **before** `//` like TM and greasyfork.org do in the hope that other entities reconsider as indentation sometimes is added by bundlers/postprocessors wrapping the code in IIFE and anyway indentation of a comment should not be enforced, especially in a language that's not indentation-based.

### DIFF
--- a/src/background/utils/script.js
+++ b/src/background/utils/script.js
@@ -1,12 +1,10 @@
 import { encodeFilename } from '#/common';
+import { METABLOCK_RE } from '#/common/consts';
 import { getOption } from './options';
-
-const metaStart = '==UserScript==';
-const metaEnd = '==/UserScript==';
 
 export function isUserScript(text) {
   if (/^\s*</.test(text)) return false; // HTML
-  if (text.indexOf(metaStart) < 0) return false; // Lack of meta block
+  if (text.indexOf('// ==UserScript==') < 0) return false; // Lack of meta block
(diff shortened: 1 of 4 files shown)


src/background/utils/script.js
```diff
@@ -47,28 +45,16 @@ export function parseMeta(code) {
...
-      const [keyName, locale] = group1.slice(1).split(':');
-      const camelKey = keyName.replace(/[-_](\w)/g, (m, g) => g.toUpperCase());
-      const key = locale ? `${camelKey}:${locale.toLowerCase()}` : camelKey;
-      const val = group2.trim();
-      const metaType = metaTypes[key] || defaultType;
-      let oldValue = meta[key];
-      if (typeof oldValue === 'undefined') oldValue = metaType.default();
-      meta[key] = metaType.transform(oldValue, val);
-    }
+  const metaBody = code.match(METABLOCK_RE)[1] || '';
```
gera2ld: `code.match(METABLOCK_RE)` might be `null` if code does not have a valid metadata block.

gera2ld: I see, it's alright then.

PR #623, opened 2019-09-28
throttle rendering of script list

### LINKED ISSUE #611: Render elements lazily in dashboard to open faster when many scripts are installed
Currently the dashboard renders all the scripts at once which takes ~5ms per script on my 4GHz i7 machine. I have 100 scripts so overall there's a 650ms pause until the page is shown when I switch to or open the script list. This is too long for a local UI where everything is expected to run instantly. 

Using DevTools network panel to capture the timeline with minimum overhead:

* VM, 100 scripts.

  image

* Stylus extension, 160 styles, same time per element, but they are rendered in groups so the first page is presented in ~200ms:

  image

I don't know Vue.js but apparently there's an even better solution based on DOM IntersectionObserver API: [Lazy Load Vue.js Components When They Become Visible](https://markus.oberlehner.net/blog/lazy-load-vue-components-when-they-become-visible/).

tophf:
Fixes #611.

Uses performance.now() to group rendering of script elements in batches, not IntersectionObserver because we need to render them all anyway, otherwise other code would have to be rewritten to accommodate for lazy render, moreover performance.now() is much cheaper.

The script element's height is pinned explicitly at its actual value so that the page height and DOM layout are correctly calculated just once, meaning the browser doesn't have to perform a very expensive DOM layout calc after the contents appears due to v-if in each throttled batch.

* First Meaningful Paint before: ~800ms
  image

* First Meaningful Paint after: ~300ms
  image


src/options/views/script-item.vue
```diff
@@ -366,6 +375,8 @@ export default {
   border-radius: .3rem;
   transition: transform .5s;
   background: white;
+  /* image size + padding + margin + border + 1 */
+  height: calc(3rem + 28px);
```
tophf: I'd like to use postcss compile-time constants/variables for these calculations.
Need help since I'm a noob at this.

src/options/views/script-item.vue
```diff
@@ -4,86 +4,88 @@
...
-          <span class="btn-ghost" @click="onEdit">
-            <icon name="code"></icon>
-          </span>
-        </tooltip>
-        <tooltip :content="labelEnable" align="start">
-          <span class="btn-ghost" @click="onEnable">
-            <icon :name="`toggle-${script.config.enabled ? 'on' : 'off'}`"></icon>
-          </span>
-        </tooltip>
+    <template v-if="canRender">
```
tophf: This huge block of changes is just an indent. You may want to "hide whitespace changes" in the config menu at the top of the page if you're viewing this on github.

src/options/utils/render-throttled.js
```diff
@@ -0,0 +1,31 @@
+const MAX_BATCH_DURATION = 50;
+const MAX_BATCH_COUNT = 50;
+const queue = [];
+let startTime = performance.now();
+let rAF = 0;
+
+export default function throttle(component) {
```
gera2ld: I think lazy loading is much simpler to achieve this, we don't have to render the invisible scripts in such a complex way.
Also we may have to find a new way to set execution orders, drag'n'drop is not easy when there are so much scripts.

tophf: Are you sure there's something much simpler than this already very simple code? Until this moment I thought there couldn't be anything simpler. Compared to all the rest of the extension, especially. I don't know Vue so it sounds to me like you're exaggerating. Also, I wouldn't say 100 scripts is "so much", half of them are disabled and I enable them occasionally, in order not to crowd the popup menu where I hide the disabled script as I always did in Tampermonkey.

tophf: As for drag'n'drop it's quite easy on a big screen and a custom theme I have that displays the scripts in up to 3 columns depending on the display width.

tophf: FWIW I never needed to adjust execution order in 10+ years of using and 5 years writing userscripts.

tophf: You probably just dislike any code that is too "close-to-metal" judging by the other "hackish" PR. While I can sympathize with the sentiment, unfortunately such things are still occasionally needed for efficient rendering when there's a bit more data than in a demo app.

gera2ld:
@tophf 
When I say "too complex" or "hacky", I don't mean the code snippet is too much or hard to understand. I mean it does not make use of what we already have and may be using something we are trying to avoid.

For example here, we are using Vue as a data driven framework, we can easily control what to render in a list, by limiting the array to a proper size, and extending it when scrolling the window.
Honestly I am a little surprised to see `requestAnimationFrame` in such case without need to show animations. Rendering the whole list in multiple event loops can cause problems. If something happens during the throttled rendering and we need to render something new, it's hard to deal with the current rendering and make sure the final view is up to date. We might need to implement a way to cancel the outdated render and to restart a new render. That's why we are always trying to avoid uncontrollable implementation in data driven development.
Similarly, we also want to avoid operating the DOM directly, or doing stuff outside the life cycle of the framework. It might work if you make it very carefully, but it's not a good pattern, and may be not friendly to others.

In this case, if we want to throttle anyway, we can save a timer in the component and clear it when outdated. But it's more complex and error prone because we are doing one thing in multiple places, right?

Anyway, thanks a lot for your great work, you've brought some great new sights to me.
Please be forgiven if you feel uncomfortable for something I said, I am not a native speaker and might be using improper words sometimes. I really appreciate your contributions.

tophf:
Yeah I thought about canceling the timer but then decided it's just so improbable to ever happen that I didn't pursue that angle. I'll see if adding the check can be performed in a way that's easy to maintain.

tophf:
As for `we also want to avoid operating the DOM directly, or doing stuff outside the life cycle of the framework`, the thing is the framework should be designed specifically to support rendering of the currently visible viewport into the data. Polymer can do it. CodeMirror can do it. Vue cannot AFAIK.

tophf:
Apparently there's no need to clear the timer. To ensure lifecycle integrity we only have to remove the destroyed components from the queue. I've updated the PR.

P.S. I've replaced requestAnimationFrame with setTimeout as there is no difference in behavior for this case. The former was used initially only because it provided a timestamp in its callback parameter but then I ended up not using it so there's no need for rAF specifically.

tophf:
OTOH, it might be much faster/better to store HTML of the last fully rendered page in localStorage, load it right after browser.js and before webpack/vuejs and use [rehydration](https://ssr.vuejs.org/guide/hydration.html). There would be a few minor quirks like for half a second some script's pre-autoupdated version number and all of the last update times will be displayed from the last page load, but it's probably negligible or we can set a CSS fade-in animation on the last update time labels (well, technically we could use DOMParser and modify the time elements but I already can imagine your reaction...)

gera2ld:
Actually it cost a lot of time to render SVG icons, not the list with just hundreds of elements. So I don't think SSR would help in this case. It would be a lot faster if we replace all SVG icons with PNG ones, but we would lose a lot of flexibility that SVG brings.
ps. I tested this couple of months ago, might be wrong.

tophf:
SVG vs PNG:

* same speed when replacing with ``
* 10% (100ms) faster when removing all `<icon>` in script-item.vue

Restoring body.innerHTML is twice as fast: the frame is painted at ~400ms but I couldn't get Vue to hydrate it at the first try...

tophf:
Hmm, there's an amazing [rendersubtree](https://github.com/WICG/display-locking/blob/master/README.md) feature, which should help here, no throttling needed.
I'll try hooking it up and test in Canary 79.

tophf:
Well, rendersubtree is a long shot, but meanwhile I'd like to reiterate that there's no simpler solution than the one suggested here (and ignoring the problem isn't a solution of course). All "proper" solutions would use 10-100 times more code/CPU.

>we can easily control what to render in a list, by limiting the array to a proper size, and extending it when scrolling the window

That won't help if customCSS displays the scripts in several columns on a big display - even if we "know" that all 100 scripts fit in the view, how would it help reduce the huge 500ms pause? The only real and proper solution is throttling and render the list in chunks. This solution has been used for decades in all kinds of apps that care about keeping their UI responsive.

With the latest code in this PR, there should be no data inconsistency regardless of anything because only the contents of a script-item is postponed, not the component itself, so Vue will automatically use the updated data thanks to its observers.

I've been using my local copy of Violentmonkey with this PR for two weeks. Every time I switch to a master-based branch for testing with its half-a-second pauses on rendering of the script list I feel like I'm accessing a slow remotely served web page...

tophf:
Now script names and icons are always rendered initially so the page appears complete.
Tomorrow I'll adjust CSS a bit so the names don't shift by 1px afterwards.

gera2ld:
Merging now. But I think the implementation is not Vue-style, so maybe I will change the throttle part a little later.

tophf:
Yeah, I admit it's ugly, but the internals of Vue arguably have uglies too.

PR #624, opened 2019-09-28
fix: update routes on 'popstate'

tophf:
This is a follow-up for #621: now that the state is not replaced we also need to observe `popstate` event on history back/forward.

Bug repro in the current master branch:
1. close all VM tabs
2. click '+' in the popup
3. click "Close" button in the editor
4. press the history back button/key in the browser
5. click "Close" button in the editor

Expected: the editor is closed and the script list is shown
Observed: nothing happens

### DIFF
--- a/src/common/router.js
+++ b/src/common/router.js
@@ -12,17 +12,21 @@ function parse(pathInfo) {
 }
 
 export const route = {};
-export const lastRoute = {};
+
+Object.defineProperties(route, {
+  stack: { value: [] },
+  last: { get: () => route.stack[route.stack.length - 1] || {} },
+});
 
 updateRoute();
 
 function updateRoute() {
-  Object.assign(lastRoute, route);
   Object.assign(route, parse(window.location.hash.slice(1)));
 }
 
+// popstate should be the first to ensure hashchange listeners see the correct route.last
+window.addEventListener('popstate', () => route.stack.pop());
 window.addEventListener('hashchange', updateRoute, false);
-window.addEventListener('popstate', updateRoute, false);
 
 export function setRoute(hash, replace) {
   let hashString = `${hash}`;
@@ -30,6 +34,7 @@ export function setRoute(hash, replace) {
   if (replace) {
     window.history.replaceState('', null, hashString);
   } else {
+    route.stack.push(Object.assign({}, route));
     window.history.pushState('', null, hashString);
   }
   updateRoute();
--- a/src/common/router.js
+++ b/src/common/router.js
@@ -21,8 +21,8 @@ function updateRoute() {
   Object.assign(route, parse(window.location.hash.slice(1)));
 }
 
-// popstate should be the first to ensure hashchange listeners see the correct route.last
-window.addEventListener('popstate', () => route.stack.pop());
+// popstate should be the first to ensure hashchange listeners see the correct lastRoute
+window.addEventListener('popstate', () => stack.pop());
 window.addEventListener('hashchange', updateRoute, false);
 
 export function setRoute(hash, replace) {
--- a/src/common/router.js
+++ b/src/common/router.js
@@ -22,6 +22,7 @@ function updateRoute() {
 }
 
 window.addEventListener('hashchange', updateRoute, false);
+window.addEventListener('popstate', updateRoute, false);
 
 export function setRoute(hash, replace) {
   let hashString = `${hash}`;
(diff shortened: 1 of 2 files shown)


gera2ld:
As I tested here, this fix works for the case in #621 but doesn't work for the case above. I think the problem is with the `window.history.back()`.

Maybe we should use `pushState` instead of `replaceState` anywhere to make it easier? Actually when I made this simple router I did not expect users to use history buttons that much.

tophf:
The fix certainly works for me, and without the fix I certainly observe the bug reported in the repro case above. Either I described the repro poorly or there is another bug that you encounter but I don't.

Master without this PR:


Master with this PR:
with-PR-ok

gera2ld:
I was having the wrong code locally. It worked but the fix seems to be accidental...
History back / forward already triggers `hashchange` so `popstate` is not necessary.

I added some logs and here is the result:

3. click "Close" button in the editor
```
replace pathname: scripts

updateRoute undefined 
route.pathname: scripts 
lastRoute.pathname: scripts/_new
```

4. press the history back button/key in the browser
```
updateRoute popstate 
route.pathname: scripts/_new 
lastRoute.pathname: scripts

updateRoute hashchange 
route.pathname: scripts/_new 
lastRoute.pathname: scripts/_new
```

The `updateRoute` has run twice, once for `popstate`, and once for `hashchange`. As a result, `lastRoute` becomes `scripts/_new` which should actually be `scripts` in this case. The accident makes this condition work:

```js
      if (!id && pathname === lastRoute.pathname) {
        // without this PR, it goes here, doesn't work since there is nothing back in the history
        window.history.back();
      } else {
        // with this PR, it goes here because the pathname of lastRoute is accidentally changed
        setRoute(pathname);
      }
```

Something more, in this case the editor becomes the last state if I click history back until the end. I think it's a little weird.

The expected result (as designed but maybe not implemented correctly) for me is:
- editor window should never be reopened in history operations, so step 4 is invalid

Anyway, this design is error prone when used with history APIs, so I think it's easier to just allow editors stored in history.

tophf:
Indeed! I think the only solution is to maintain a stack of navigations and pop it explicitly on popstate. See the updated PR.

tophf:
Jeez, webpack generates a lot of code for `{ ...foo }` so I've reverted to the good old Object.assign.

src/common/router.js
```diff
@@ -12,15 +12,20 @@ function parse(pathInfo) {
 }
 
 export const route = {};
-export const lastRoute = {};
+
+Object.defineProperties(route, {
+  stack: { value: [] },
+  last: { get: () => route.stack[route.stack.length - 1] || {} },
```
gera2ld: `route.last` can be calculated as needed or a computed property powered by Vue, so I don't think we need `defineProperties` here, just `const route = { stack: [] }` is enough.

tophf: A normal property will be copied by Object.assign so I would have to declare a separate variable for `stack`. Should I do it anyway?

gera2ld: I see, it makes sense then.
Normally I would explicitly pick the fields that I care so that no other data will be stored unexpectedly.

tophf: I also tried a separate variable for `stack` (not even exported) and a separate exported function like `lastRoute`, which is also fine by me, but I just wanted to expose the single `route` entity. Anyway I don't know how to organize this better so I'm all ears.

src/common/router.js
```diff
@@ -29,6 +34,7 @@ export function setRoute(hash, replace) {
   if (replace) {
     window.history.replaceState('', null, hashString);
   } else {
+    route.stack.push(Object.assign({}, route));
```
gera2ld: How about just picking `pathname` and `query` here?
```js
route.stack.push({
  pathname: route.pathname,
  query: route.query,
});
```
So we don't have to define `stack` in a different way.

tophf: You didn't include `paths` and what if in the future you add another property and it needs to be explicitly copied here? Nah. Let's see if you like the new code better.

gera2ld: `paths` is for convenience and thus not necessary to be cached, only `pathname` and `query` are real data. It can even be removed I think.

tophf: So what about the new code?

gera2ld: How about exporting last route as a function?

tophf: Okay.

PR #625, opened 2019-09-29
Drop support for older browsers

### LINKED ISSUE #622: Skip running polyfills in modern browsers
Currently Violentmonkey spends 6-12ms on each page load of every web page on my i7 4GHz machine to [install polyfills](https://github.com/violentmonkey/violentmonkey/blob/master/src/common/polyfills.js) for features that are implemented natively since 2016 in all modern browsers. 

polyfill           | Chrome | Firefox 
-------------------|--------|--------
Object.assign      | 45     | 34
Object.values      | 54     | 47
Array.includes     | 47     | 43
Array.find         | 45     | 25
Array.findIndex   | 45     | 37
String.includes    | 41     | 40
String.startsWith | 41     | 17
String.endsWith   | 41     | 17
String.repeat      | 41     | 24

Chrome 41-45 is 2015, 54 is 2016.
Firefox 17 is 2012, 47 is 2016.

profiler

It takes more time than the actually required code of all other content scripts combined! This is a ridiculous amount of time (also half a 60fps frame) for such a trivial task, especially since it's not needed in 99.999% of cases. On a slower machine the time wasted will be bigger of course.

How we can solve it:

1. Abandon support for those presumably 0.001% of users who may still use a pre-2016 browser, but it probably requires analytics, which is frowned upon (see #602).

2. Find another, faster set of polyfills

3. Write the faster polyfills ourselves

4. Add a runtime check of `navigator.userAgent` and write it into the output script using a webpack hook. While rewriting the output is hackish, there's probably no other way to add a runtime condition to code built from a static `import`.

gera2ld:
browserslist:
```
Chrome >= 55
Firefox >= 52
```

close #622

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,1 +1,2 @@
-Chrome >= 45
+Chrome >= 55
+Firefox >= 52
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,5 @@ module.exports = {
   },
   globals: {
     browser: true,
-    zip: true,
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
     },
   },
   rules: {
+    'prefer-object-spread': 'off',
   },
   globals: {
     browser: true,
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -73,10 +73,6 @@ function manifest() {
     const data = yaml.safeLoad(input);
     // Strip alphabetic suffix
     data.version = pkg.version.replace(/-[^.]*/, '');
-    if (process.env.TARGET === 'firefox') {
-      data.version += 'f';
-      data.applications.gecko.update_url = 'https://violentmonkey.top/static/updates.json';
-    }
     file.path = file.path.replace(/\.yml$/, '.json');
     return JSON.stringify(data);
   }))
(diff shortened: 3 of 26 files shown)


tophf:
Tested it on a few sites, seems fine.
Now browser.js takes only ~1ms to run.
The code changes seem straightforward so this is a LGTM from me!

Chrome 55 and Firefox both add support for `async/await` so I guess we can use that freely now?

tophf:
How about you merge this and I'll update my PRs to use Map/Set/async/await where appropriate?

Also, what do you think about using a `Proxy` so instead of `sendMessage({ cmd: 'GetMetas', data: data.ids })` we would use `vmAPI.GetMetas(data.ids)` ?

Here's a tentative implementation:
```js
const vmAPI = new Proxy({}, {
  get(_, cmd) {
    return data => sendMessage({ cmd, data });
  },
});
```

gera2ld:
It's OK to use `async / await` everywhere.
It's OK to use `Map / Set` if necessary but I think in most cases plain objects are enough.

I suggest not to use `Proxy` because it's not able to be polyfilled and it's hard for us to port to other platforms or older browsers. FYI there is a Violentmonkey for Maxthon, of which this code is synced from this repo periodically.

tophf:
What about a babel transform or a babel macro plugin?
The improved clarity and abstraction is worth it, I think.

tophf:
Also, `"strict_min_version": "52"` for FF and `"minimum_chrome_version":"55"` for Chrome in manifest.json are needed to make sure those few users of the old browsers don't get the new incompatible extension.

PR #626, opened 2019-09-29
fix: avoid FOUC

gera2ld:
- Move scripts except `index.js` to `<head>` but keep `index.js` in `<body>` so that we don't have to check if `body` is ready everywhere. When `index.js` runs synchronously, first paint won't trigger because `<body>` is empty (except for `<script>`s).

```html
<head>
<script src="browser.js"></script>
<script src="common.js"></script>
</head>
<body>
<script src="index.js"></script>
</body>
```

- Load custom CSS from cache (`localStorage`) for the first time to avoid first paint with empty style.

When `DOMContentLoaded` is triggered after `index.js` runs, the custom style should have been attached to `<head>` and the main UI should have been rendered by Vue synchronously. So FOUC is avoided.

### DIFF
--- a/scripts/plaid.conf.js
+++ b/scripts/plaid.conf.js
@@ -6,11 +6,20 @@ const { isProd } = require('@gera2ld/plaid/util');
  * - value.html: options object passed to HtmlWebpackPlugin.
  * - value.html.inlineSource: if true, JS and CSS files will be inlined in HTML.
  */
+const injectTo = item => {
+  if ([
+    '../browser.js',
+    '../common.js',
+  ].includes(item.attributes.src)) {
+    return 'head';
+  }
+};
 const htmlFactory = extra => options => ({
   ...options,
   title: 'Violentmonkey',
   ...extra,
   chunks: ['browser', ...options.chunks],
+  injectTo,
 });
 exports.pages = {
   'browser': {
--- a/src/common/ui/style/index.js
+++ b/src/common/ui/style/index.js
@@ -2,15 +2,25 @@ import options from '../../options';
 import './style.css';
 
 let style;
+const CACHE_KEY = 'cacheCustomCSS';
+
+const setStyle = (css) => {
+  if (css && !style) {
+    style = document.createElement('style');
+    document.head.appendChild(style);
+  }
+  if (css || style) {
+    css = css || '';
+    style.textContent = css;
+    localStorage.setItem(CACHE_KEY, css);
+  }
+};
+
+setStyle(localStorage.getItem(CACHE_KEY));
+
 options.hook((changes) => {
   if ('customCSS' in changes) {
     const { customCSS } = changes;
-    if (customCSS && !style) {
-      style = document.createElement('style');
-      document.head.appendChild(style);
-    }
-    if (customCSS || style) {
-      style.textContent = customCSS;
-    }
+    setStyle(customCSS);
   }
 });
--- a/src/common/ui/style/index.js
+++ b/src/common/ui/style/index.js
@@ -7,16 +7,26 @@ const CACHE_KEY = 'cacheCustomCSS';
 const setStyle = (css) => {
   if (css && !style) {
     style = document.createElement('style');
-    document.head.appendChild(style);
+    document.documentElement.appendChild(style);
   }
   if (css || style) {
     css = css || '';
     style.textContent = css;
-    localStorage.setItem(CACHE_KEY, css);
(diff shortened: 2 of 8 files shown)


src/common/ui/style/index.js
```diff
@@ -2,15 +2,25 @@ import options from '../../options';
...
+    document.head.appendChild(style);
+  }
+  if (css || style) {
+    css = css || '';
+    style.textContent = css;
+    localStorage.setItem(CACHE_KEY, css);
+  }
+};
+
+setStyle(localStorage.getItem(CACHE_KEY));
```
tophf: Why are you relying on localStorage not being [accidentally] edited by the user in devtools? Using the browser.extension.getBackgroundPage() is more reliable, especially in Firefox where just accessing `localStorage` object will throw with a security error for users who disable DOM storage in about:config.

scripts/plaid.conf.js
```diff
@@ -6,11 +6,20 @@ const { isProd } = require('@gera2ld/plaid/util');
  * - value.html: options object passed to HtmlWebpackPlugin.
  * - value.html.inlineSource: if true, JS and CSS files will be inlined in HTML.
  */
+const injectTo = item => {
+  if ([
+    '../browser.js',
+    '../common.js',
+  ].includes(item.attributes.src)) {
+    return 'head';
```
tophf: After I **manually** reposition the tags in html file it indeed solves the FOUC and the white frame flashing, but `yarn build` doesn't place the script tags correctly, they're still inside `<body>`. Same for `yarn dev`.

src/common/ui/style/index.js
```diff
@@ -2,15 +2,25 @@ import options from '../../options';
 import './style.css';
 
 let style;
+const CACHE_KEY = 'cacheCustomCSS';
+
+const setStyle = (css) => {
```
tophf: What about `(css = '')` so you won't need `css = css || ''` below?

src/common/ui/style/index.js
```diff
@@ -2,15 +2,25 @@ import options from '../../options';
...
+  if (css && !style) {
+    style = document.createElement('style');
+    document.head.appendChild(style);
```
tophf: To make this usable in **dev build** you need to put the style element after `<head>`:
```js
document.documentElement.appendChild(style);
```
because we're running early now and other components will add their styles after customCSS thus unpredictably messing up selectors with the same CSS specificity.

src/common/ui/style/index.js
gera2ld: As [the document of getBackgroundPage](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getBackgroundPage) says:

> Note: This method cannot be used in Private Browsing mode — it always returns an empty array. For more info see bug 1329304.

> especially in Firefox where just accessing localStorage object will throw with a security error for users who disable DOM storage in about:config.

Just tested, `browser.extension.getBackgroundPage()` does not work in private browsing mode, but `localStorage` does. I'd better use `try / catch` here.

> Why are you relying on localStorage not being [accidentally] edited by the user in devtools?

If the user modifies things in devtools, everything becomes unreliable, so does `browser.extension.getBackgroundPage()`. I think this is not the case we want to cover.

> There are/were exceptions in Firefox such as private mode or containers (maybe not, I've tested it a year ago)

I may have experienced this also but I can't find the reference. Anyway, it worked just now.

src/common/ui/style/index.js
gera2ld: I will change it to `document.documentElement.appendChild(style);`.

src/common/ui/style/index.js
gera2ld: > because getItem returns null, not undefined.

That's why I use `css = css || ''`.

I prefer `localStorage.getItem` because it's more explicit for reading a value.

src/common/ui/style/index.js
gera2ld: > Why are you relying on localStorage not being [accidentally] edited by the user in devtools?

Also, the cache will be updated once options are loaded from background, and every time user changes `customCSS`, so this won't be a problem.

tophf:
LGTM.

As for those FF users with DOM storage disabled (what a weird thing to do but oh well), hopefully they won't see FOUC/flash anyway because it's very rare in Firefox as is, apparently because FF waits a bit longer for the meaningful frame to appear.

PR #627, opened 2019-09-30
chore: compile to native ES2015

tophf:
Removes transpiling of async/await into the extremely clunky [regeneratorRuntime](https://github.com/facebook/regenerator/blob/master/packages/regenerator-runtime/runtime.js) as well as some of the other ES2015 stuff implemented Chrome 55 and Firefox 52 which are the minimum supported versions for ViolentMonkey.


tophf:
Don't forget to run `npm install`, I still often do.

gera2ld:
We don't need to do that much, `.browserslistrc` should be responsible for it.

As for why async function is compiled for Firefox 52, see https://github.com/babel/babel/issues/8204.

I think we can just exclude `transform-generator` as a workaround for now and upgrade the minimum Firefox version to 53 after users move on.

gulpfile.js
```diff
@@ -2,7 +2,7 @@ const gulp = require('gulp');
 const del = require('del');
 const log = require('fancy-log');
 const gulpFilter = require('gulp-filter');
-const uglify = require('gulp-uglify');
+const uglify = require('gulp-uglify-es').default;
```
gera2ld: This is unnecessary because uglify is only used for third-party libraries, which are already compiled to ES5.

tophf:
It should be possible to have a different setting in babelrc when building for Firefox. Quick googling shows babel-loader can be configured in webpack. I'd rather not use `async/await` at all if it's replaced with the obscene regeneratorRuntime even in Chrome which is my primary browser.

tophf:
>We don't need to do that much, .browserslistrc should be responsible for it.

Evidently it doesn't work.

gera2ld:
> Evidently it doesn't work.

Either changing `Firefox >= 52` to `Firefox >= 53` or excluding `transform-generator` will work. I prefer the latter since we have quite a few users on Firefox 52 now.

> I'd rather not use async/await at all if it's replaced with the obscene regeneratorRuntime even in Chrome which is my primary browser.

It's not replaced. This is just a workflow for bundlers to work for different browsers. They check if browsers support the features and only transform them if the browsers do not supported well. Somehow Firefox 52 is marked as not supporting generators well, leading to impact on its upstream feature `async` functions. But we know exactly we don't depend on special cases of generators so we can just disable generator check as a workaround, and then everything goes as expected.

tophf:
Just rebuilt the mater branch using `yarn build` and like I said I see all `async` functions are replaced with regeneratorRuntime.

Did you actually check your output js?

tophf:
I mean don't you have a separate build command for Firefox? Why should all Chrome users be encumbered by the terrible async polyfill?

tophf:
Hm, I see you've disabled transform-generator and the rest of the exceptions in this PR are redundant so I'm closing it.

gera2ld:
> Just rebuilt the master branch using yarn build and like I said I see all async functions are replaced with regeneratorRuntime.

You need to change the config as I said above: Either changing `Firefox >= 52` to `Firefox >= 53` or excluding `transform-generator`. I have updated master branch just now. You can try it again.

> Did you actually check your output js?

Yes, I did.

> I mean don't you have a separate build command for Firefox? Why should all Chrome users be encumbered by the terrible async polyfill?

Because there is no need to. I chose `Chrome >= 55` and `Firefox >= 52` because they both have good support for relatively new features.

tophf:
Ah, I see.
Let's hope no one gets hit by those vague failed generator tests.

tophf:
Oops, for-of is still expanded to something terrible in master:
```js
try {
  for (var u, $ = n.entries()[Symbol.iterator](); !(a = (u = $.next()).done); a = !0) {
    /////// code
  }
} catch (e) {
  l = !0,
  c = e
} finally {
  try {
    a || null == $.return || $.return()
  } finally {
    if (l)
      throw c
  }
}
```

PR #628, opened 2019-09-30
speed up setting icon data in Chrome

tophf:
Chrome extracts the icon data via Canvas API from each icon size specified and it wastes up to 40ms in our background page context (hence, blocking it for our code) each time a tab is opened that has a blacklisted or noninjectable state. The most frequent abuser is the default new tab page in Chrome.

This PR adds an explicit cache for imageData obtained via Canvas API.

### DIFF
--- a/src/background/utils/icon.js
+++ b/src/background/utils/icon.js
@@ -121,15 +121,19 @@ async function setIcon(tab = {}, data = {}) {
   // modern Chrome and Firefox use 16/32, other browsers may still use 19/38 (e.g. Vivaldi)
   const mod = data.blocked && 'b' || !isApplied && 'w' || '';
   const iconData = {};
-  const iconOptions = {
-    tabId: tab.id,
-    [iconCache ? 'imageData' : 'path']: iconData,
-  };
   for (const n of [16, 19, 32, 38]) {
     const path = `/public/images/icon${n}${mod}.png`;
-    iconData[n] = !iconCache ? path : iconCache[path] || await loadImageData(path);
+    let icon = iconCache ? iconCache[path] : path;
+    if (!icon) {
+      icon = await loadImageData(path);
+      iconCache[path] = icon;
+    }
+    iconData[n] = icon;
   }
-  browserAction.setIcon(iconOptions);
+  browserAction.setIcon({
+    tabId: tab.id,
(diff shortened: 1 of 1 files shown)


src/background/utils/icon.js
```diff
@@ -111,13 +117,36 @@ function updateState(tab, url = tab.url) {
...
+    [iconCache ? 'imageData' : 'path']: iconData,
+  };
+  for (const n of [16, 19, 32, 38]) {
+    const path = `/public/images/icon${n}${mod}.png`;
+    iconData[n] = !iconCache ? path : iconCache[path] || await loadImageData(path);
+  }
+  browserAction.setIcon(iconOptions);
+}
+
+function loadImageData(path) {
```
gera2ld: We can use `memoize` from `#/common` instead of implementing cache repeatedly.

tophf: So far memoize seems to be used only once so maybe it's better to remove it there :-)
I'm not convinced we need an abstraction for something so simple.

gera2ld:
This is amazing. Should we also support Firefox? or maybe in the future?

tophf:
I did some quick measurements in FF and it seems they cache the data efficiently. Or I did the tests incorrectly. There isn't a profiler in their devtools that shows the internals of extensions API.

src/background/utils/icon.js
gera2ld: I think it's not about simple or not. By implementing locally, you need to define a variable outside the function and read / write cache in the original function, which is kind of dirty.

For me cache should be something that can easily opt in or out, without affecting the original function. So `loadImageData` should work exactly as `cachedLoadImageData`, and the user who uses this function need not to know about it.

gera2ld: > So far memoize seems to be used only once

Maybe my old code is badly designed, or there was little chance to cache a function permanently like this.

tophf: Okay, I'll use memoize then.

tophf: Uhm, I don't like this.
`memoize` stores a resolved Promise in its cache so each time `await` will run internally.
And I'll still have to declare a separate variable like isChrome or isIconCacheEnabled.

tophf: I'll just move updating of the cache inside setIcon.

Issue #420, opened 2018-06-17
Implement real @run-at document-start


KOLANICH:
I'm interested in `document-start`.

gera2ld:
`document-start` is not guaranteed.

KOLANICH:
1 It seems that runAt is not implemented at all. Do I understand right?
2 Can `contentScripts.register` solve this issue?

gera2ld:
1. `@run-at document-start` is not guaranteed, the script is just injected ASAP.
2. `contentScripts.register` should help, but Chrome does not support it.

KOLANICH:
1. Searched one more time by downloading and grepping, now I see it is used.
2. Should we implement the separate logic for the browsers with this API?

gera2ld:
> Should we implement the separate logic for the browsers with this API?

Personally, I don't use Firefox as my primary browser, and I don't want to take much time on Firefox-only features. Anyway it's good to have a way to implement it and PRs are welcome.

bfred-it:
> `contentScripts.register` should help, but Chrome does not support it.

You can use the [Chrome polyfill for `browser.contentScript.register`](https://github.com/bfred-it/content-scripts-register-polyfill#readme) for that API

gera2ld:
@bfred-it Nice idea.

But I wonder whether `document-start` is reliable in this polyfill. In the listener there are several asynchronous functions, but the tab is loading and it would not wait for the listener to finish its asynchronous tasks.
Consequently, at the time when scripts are injected, the tab may have been loaded already.
I think it makes no difference with the current Violentmonkey way, to inject ASAP.

bfred-it:
> But I wonder if `document-start` is reliable in this polyfill

I did verify that the CSS works correctly (without FUOC) with `document_start` but I didn't verify the speed of the scripts.

I expect the native Firefox `register()` API to behave as well as the regular `content_script` injection via manifest.json, while the polyfill uses `tabs.executeScript` to inject the script on `tabs.onUpdated`, which is the fastest way I'm aware of.

How are scripts currently injected in `violentmonkey`?

gera2ld:
Just made an experiment and the conclusion is, **even injecting script on tab update synchronously, `document_start` cannot be ensured.**

Here is my code:

```js
chrome.tabs.onUpdated.addListener((id, changedInfo, tab) => {
  if (changedInfo.status === 'loading') {
    browser.tabs.executeScript(id, {
[...]

PR #629, opened 2019-09-30
speedup: script injection

tophf:
* `getScriptsByURL` before: ~90ms every time

  Clip643

* `getScriptsByURL` after: ~40ms first run, 2-20ms subsequent runs

# Changes:

1. only one `setTimeout` call per event loop cycle is used now in most cases, could be more if cacing items with progressively diminishing `lifetime` parameter, but we don't do that in bulk.

    DOM timers are relatively expensive: ~50ms out of ~90ms was wasted on set/clearTimeout when there are many matches/globs (e.g. Handy Image userscript), and lots of cache items are generated by tester.js

2. `getScriptsByURL` is pre-calculated in webRequest.onHeadersReceived for main_frame/sub_frame and http/https URLs so when the content script sends GetInjected we already have the data.

3. All background cache lifetime was extended to 5 minutes.

Now userscripts run after ~20ms since the real document_start.
We can make it even faster, see https://github.com/violentmonkey/violentmonkey/issues/420#issuecomment-536465527

### DIFF
--- a/src/common/cache.js
+++ b/src/common/cache.js
@@ -1,10 +1,14 @@
-const defaults = {
-  lifetime: 3000,
-};
+const DEFAULT_LIFETIME = 3000;
 
-export default function initCache(options) {
-  const cache = {};
-  const { lifetime: defaultLifetime } = options || defaults;
+export default function initCache({
+  lifetime: defaultLifetime = DEFAULT_LIFETIME,
+} = {}) {
+  let cache = {};
+  // setTimeout call is very expensive when done frequently,
+  // 1000 calls performed for 50 scripts consume 50ms on each tab load,
+  // so we'll schedule trim() just once per event loop cycle,
+  // and then trim() will trim the cache and reschedule itself to the earliest expiry time.
+  let timer;
   return {
     get, put, del, has, hit, destroy,
   };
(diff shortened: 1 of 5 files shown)


src/common/cache.js
```diff
@@ -1,40 +1,61 @@
...
-    Object.keys(cache).forEach(del);
+    clearTimeout(timer);
+    timer = 0;
+    cache = {};
+  }
+  function trim() {
+    const now = performance.now();
+    let closest = Number.MAX_SAFE_INTEGER;
+    for (const key in cache) {
+      if (Object.hasOwnProperty.call(cache, key)) {
```
gera2ld: How about this?
```js
for (const key of Object.keys(cache)) {
```

tophf: Possible, if you insist, but usually I dislike the idea of temporary creating a big array just because there's a common prejudice in JS world against for-in...

gera2ld: OK, we can just leave it.

tophf: What if we switch to Map()?

tophf: Like this:
```js
export default function initCache({
  lifetime: defaultLifetime = 3000,
} = {}) {
  let cache = new Map();
  // setTimeout call is very expensive when done frequently,
  // 1000 calls performed for 50 scripts consume 50ms on each tab load,
  // so we'll schedule trim() just once per event loop cycle,
  // and then trim() will trim the cache and reschedule itself to the earliest expiry time.
  let timer;
  let minLifetime = -1;
  return {
    get, put, del, has, hit, destroy,
  };
  function get(key, def) {
    const item = cache[key];
    return item ? item.value : def;
  }
  function put(key, value, lifetime = defaultLifetime) {
    if (value) {
      const expiry = lifetime + performance.now();
      cache.set(key, { value, expiry });
      reschedule(lifetime);
    } else {
      delete cache[key];
    }
  }
  function del(key) {
    cache.delete(key);
  }
  function has(key) {
    return Object.hasOwnProperty.call(cache, key);
  }
  function hit(key, lifetime = defaultLifetime) {
    const entry = cache.get(key);
    if (entry) {
      entry.expiry = performance.now() + lifetime;
      reschedule(lifetime);
    }
  }
  function destroy() {
    clearTimeout(timer);
    timer = 0;
    cache.clear();
  }
  function reschedule(lifetime) {
    if (timer) {
      if (lifetime >= minLifetime) return;
      clearTimeout(timer);
    }
    minLifetime = lifetime;
    timer = setTimeout(trim, lifetime);
  }
  function trim() {
    // next timer won't be able to run earlier than 10ms
    // so we'll sweep the upcoming expired entries in this run
    const now = performance.now() + 10;
    let closestExpiry = Number.MAX_SAFE_INTEGER;
    for (const [key, { expiry }] of cache.entries()) {
      if (expiry < now) {
        cache.delete(key);
      } else if (expiry < closestExpiry) {
        closestExpiry = expiry;
      }
    }
    minLifetime = closestExpiry - now;
    timer = closestExpiry < Number.MAX_SAFE_INTEGER
      ? setTimeout(trim, minLifetime)
      : 0;
  }
}
```

gera2ld: Looks good to me. The only concern is about compatibility, but it should not be a problem for Chrome and Firefox.

tophf: So do we switch or - given your preference for objects - don't?

tophf: Hey, what about https://github.com/nicolo-ribaudo/for-own.macro
```js
import own from "for-own.macro";

for (const key in own(obj)) {
  const value = obj[key];
}
```

gera2ld: I prefer `Map`. The `macro` thing is interesting though, maybe we can use it somewhere else.

PR #630, opened 2019-10-01
speed up sandbox initialization up to ~33%

tophf:
### Performance improvements

Averaged numbers on www.google.com and nine scripts of various complexity/size.  
Six were running on `start`, two on `end`, and one on `idle`.

run-at | start | end | idle | total
---------|------|-----|------|---
Before, ms | 34 | 17 | 18 | 69
After, ms | 23 | 12 | 16 | 51
speedup %  | 33% | 30% | 10% | 25%

The most important improvement is `document-start` because it's the one that delays the page UI.

Measured using the simpler `JavaScript Profiler` in Chrome's devtools as it applies less overhead. Note that now document-start block in the flamechart is single (contiguous), but previously it consisted of two blocks due to `bridge.ready` which is now removed (see notes below).

image

### Changes in behavior:

* each `document-idle` script runs in its own setTimeout to prevent long pauses due to a lot of scripts running at once. Unfortunately we can't do it for other types of `run-at` as they kinda sorta guarantee the execution moment.

### Trivial but noteworthy changes:

* `transform-parameters` is disabled so that `(...args) => {}` is preserved. All features of this transform are natively supported since Chrome 49, Firefox 52. Maybe it's a Babel bug as its transpiled code needlessly creates an intermediate array copy of the `arguments` object.
* `bridge.ready` is removed because scripts created using textContent run synchronously
* `CustomEvent.detail` won't be serialized in Chrome
* `onLoadScripts` will reuse gmApi descriptors and registers in bulk using Object.defineProperties
* `<html id="__Violentmonkey">` won't fool us anymore

P.S. I like how Violentmonkey splits script creation by `run-at` stages. Whereas TM creates all sandboxes at once, and it produces a very long delay (up to 100ms) at every page start with my userscripts.

### DIFF
--- a/src/injected/content/index.js
+++ b/src/injected/content/index.js
@@ -140,14 +140,11 @@ function checkInjectable() {
 }
 
 function injectScripts(contentId, webId, data, scriptLists) {
-  let props = [];
+  const props = [];
   // combining directly to avoid GC due to a big intermediate object
   const addUniqProp = key => !props.includes(key) && props.push(key);
-  const windowProps = Object.getOwnPropertyNames(window);
-  // buggy Firefox may list duplicate props
-  if (isFirefox) windowProps.forEach(addUniqProp);
-  else props = windowProps;
-  Object.getOwnPropertyNames(global).forEach(addUniqProp);
+  // some browsers may list duplicate props within one window object!
+  [window, global].forEach(wnd => Object.getOwnPropertyNames(wnd).forEach(addUniqProp));
   const args = [
     webId,
     contentId,
--- a/src/injected/content/index.js
+++ b/src/injected/content/index.js
@@ -2,7 +2,7 @@ import { isFirefox } from '#/common/ua';
 import { getUniqId } from '#/common';
 import { INJECT_PAGE, INJECT_CONTENT, INJECT_AUTO } from '#/common/consts';
 import {
-  bindEvents, sendMessage, attachFunction,
+  bindEvents, sendMessage, attachFunction, setJsonDump, jsonDump,
 } from '../utils';
 import bridge from './bridge';
 import { tabOpen, tabClose, tabClosed } from './tabs';
@@ -17,10 +17,13 @@ const IS_TOP = window.top === window;
 
 // Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1408996
 const VMInitInjection = window[process.env.INIT_FUNC_NAME];
+delete window[process.env.INIT_FUNC_NAME];
 
 const ids = [];
 const enabledIds = [];
 const menus = {};
+// userscripts running in the content script context are messaged via invokeGuest
+const invokableIds = [];
 
 function setBadge() {
   // delay setBadge in frames so that they can be added to the initial count
@@ -36,19 +39,35 @@ function setBadge() {
 
 const bgHandlers = {
   Command(data) {
-    bridge.post({ cmd: 'Command', data });
+    const id = +data.split(':', 1)[0];
+    const realm = invokableIds.includes(id) && INJECT_CONTENT;
+    bridge.post({ cmd: 'Command', data, realm });
   },
   GetPopup: getPopup,
   HttpRequested: httpRequested,
   TabClosed: tabClosed,
   UpdatedValues(data) {
-    bridge.post({ cmd: 'UpdatedValues', data });
+    const realms = [
+      { data: {}, present: false },
+      { data: {}, present: false, realm: INJECT_CONTENT },
+    ];
+    Object.keys(data).forEach((id) => {
+      const r = realms[invokableIds.includes(id) ? 1 : 0];
+      r.data[id] = data[id];
+      r.present = true;
+    });
+    realms
+    .filter(r => r.present)
+    .forEach(({ data: d, realm }) => {
+      bridge.post({ cmd: 'UpdatedValues', data: d, realm });
+    });
   },
   NotificationClick: onNotificationClick,
   NotificationClose: onNotificationClose,
 };
 
 export default function initialize(contentId, webId) {
+  // may be overriden in injectScripts
   bridge.post = bindEvents(contentId, webId, onHandle);
   bridge.destId = webId;
 
@@ -56,6 +75,8 @@ export default function initialize(contentId, webId) {
     const handle = bgHandlers[req.cmd];
     if (handle) handle(req.data, src);
   });
+
+  setJsonDump({ native: true });
 
   return sendMessage({
     cmd: 'GetInjected',
@@ -106,7 +127,7 @@ function checkInjectable() {
     a.id = domId;
     document.documentElement.appendChild(a);
   };
-  inject(`(${detect.toString()})(${JSON.stringify(id)})`);
+  inject(`(${detect})(${jsonDump(id)})`);
   const a = document.querySelector(`#${id}`);
   const injectable = !!a;
   if (a) a.parentNode.removeChild(a);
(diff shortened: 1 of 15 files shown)


tophf:
I've also used [more safe copies of functions](https://github.com/tophf/violentmonkey/commit/e3cde1795df435fab8cb23492f02aab78d6830a1) but I'm not sure it should be included in this PR. I didn't replace all of the methods though. For example, Promise or FormData still link to the original `prototype`.

gera2ld:
> I've also used [more safe copies of functions](https://github.com/tophf/violentmonkey/commit/e3cde1795df435fab8cb23492f02aab78d6830a1) but I'm not sure it should be included in this PR. I didn't replace all of the methods though. For example, Promise or FormData still link to the original `prototype`.

OK to me. It's a lot of work 😄

src/injected/utils/helpers.js
```diff
@@ -48,6 +54,20 @@ export const isArray = obj => (
   obj && typeof obj.length === 'number' && typeof obj.splice === 'function'
 );
 
+export const VM_KEY = '__Violentmonkey';
+export const isContentScript = !!window.browser
```
gera2ld: I think it's more reliable to check with `bridge.mode`. Just that some more work is needed to access to `bridge`.
```js
const isContentScript = bridge.mode === INJECT_CONTENT;
```

gera2ld: If we are going to support userscripts that override `Array#toJSON`, this check won't be needed.

src/injected/utils/index.js
```diff
@@ -6,16 +6,21 @@ export {
   sendMessage, request, throttle, cache2blobUrl,
 } from '#/common';
 
+const isChrome = !!window.chrome
+  && /^function\s?\w*\s*\(\)\s*{\s*\[native code]\s*}\s*$/.test(
+    (window.chrome.app || {}).getDetails,
+  );
```
gera2ld: I suggest to check this in the content script, and pass it to `VMInitInjection`.

gera2ld: Can we just use a simple string compare here?

src/injected/utils/helpers.js
tophf: Uhm, isn't `helpers.js` also used in the main content script? That is in addition to the sandbox. I don't have a firm grasp, apparently on how bridge is used exactly, anyway.

src/injected/utils/index.js
tophf: We can but I'm afraid Chrome can just change it in the future by adding a space or a newline. I'm not sure all versions of Chrome have the same string.

src/injected/utils/helpers.js
tophf: I probably should use `let` here and reassign to jsonStringifySafe if a script is received that wants to inject into the content script context.

src/injected/utils/index.js
tophf: Oh, there's no need for this complicated check if we do it in the content script as you suggested.

tophf:
I'm still working on it, trying to ensure the fast path (no jsonDump, object in CustomEvent) is chosen when possible.

tophf:
Just finished with the faster messaging (via direct invocation of onHandle) for userscripts running inside the content script context, which is arguably more important for Firefox where their messaging was always stringified. At first glance it seems to be working but let's test it for a few more days.

tophf:
Guess it's ready. Hopefully you wouldn't mind I've extracted gm-api and scripts handling out of web/index.js, which was quite big. The gm-api module seems fine both content-wise and name-wise, but I'm not so sure about these things in scripts.js.

tophf:
I've temporarily moved the extracted modules back in to simplify reviewing.
I'll extract them again later once you check the PR, if you agree.

tophf:
I've simplified the review diff even more.

I still want to extract these things:
* gm-api.js: createGmApiProps and everything it uses
* gm-wrapper.js: wrapGM, getWrapper, createWrapperMethods
* scripts.js (or load-scripts.js?): onLoadScripts

src/injected/content/index.js
```diff
@@ -106,68 +132,88 @@ function checkInjectable() {
...
-  ].forEach((keys) => {
-    keys.forEach((key) => { props[key] = 1; });
-  });
+  let props = [];
+  // combining directly to avoid GC due to a big intermediate object
+  const addUniqProp = key => !props.includes(key) && props.push(key);
+  const windowProps = Object.getOwnPropertyNames(window);
+  // buggy Firefox may list duplicate props
+  if (isFirefox) windowProps.forEach(addUniqProp);
+  else props = windowProps;
```
gera2ld: What if other browsers give duplicate props too? I think we'd better just use `addUniqProp`.

src/injected/index.js
```diff
@@ -15,10 +17,13 @@ import initialize from './content';
 
   initBridge();
 
+  const { go } = History.prototype;
+  const { querySelector } = Document.prototype;
+  const { get: readyState } = Object.getOwnPropertyDescriptor(Document.prototype, 'readyState');
```
gera2ld: The name `readyState` is confusing, how about `getReadyState`?

src/injected/utils/helpers.js
```diff
@@ -2,110 +2,37 @@
...
-
-export const toString = bindThis(objectProto.toString);
-const numberToString = bindThis(Number.prototype.toString);
-const stringSlice = bindThis(String.prototype.slice);
-const stringCharCodeAt = bindThis(String.prototype.charCodeAt);
-const { fromCharCode } = String;
+export const {
+  filter, forEach, indexOf, join, map, push,
+  // arraySlice, // to differentiate from String::slice which we use much more often
+  includes = function _(item) { return this::indexOf(item) >= 0; },
```
gera2ld: We don't need this polyfill for modern browsers.

src/injected/utils/helpers.js
```diff
@@ -2,110 +2,37 @@
...
+export const { fromCharCode } = String;
+export const { warn } = console;
+export const { addEventListener } = EventTarget.prototype;
+export const { append, setAttribute } = Element.prototype;
+export const { createElement } = Document.prototype;
 
 export const isArray = obj => (
   // ES3 way, not reliable if prototype is modified
-  // toString(obj) === '[object Array]'
+  // Object's toString(obj) === '[object Array]'
```
gera2ld: This line is the old code which was deprecated:

```js
toString(obj) === '[object Array]'
```

Maybe it's better to change to
```js
objectToString(obj) === '[object Array]'
// or
Object.prototype.toString.call(obj) === '[object Array]'
```

src/injected/web/requests.js
```diff
@@ -112,7 +119,59 @@ function start(req, id) {
...
+async function encodeBody(body) {
+  const cls = getType(body);
+  switch (cls) {
+  case 'formdata': {
+    const data = {};
+    const resolveKeyValues = async (key) => {
+      const values = body::getAll(key)::map(encodeBody);
+      data[key] = await Promise.all(values);
+    };
+    await Promise.all([...body::keys()]::map(resolveKeyValues));
```
gera2ld: `map` is incorrect here.
I think the binding syntax `::` is too early to use because both tsserver and ESLint do not support it and thus cannot prevent such errors for us.

src/injected/web/requests.js
```diff
@@ -112,7 +119,59 @@ function start(req, id) {
...
+const { FileReader } = global;
+const { readAsArrayBuffer } = FileReader.prototype;
+
+async function encodeBody(body) {
+  const cls = getType(body);
+  switch (cls) {
+  case 'formdata': {
+    const data = {};
+    const resolveKeyValues = async (key) => {
+      const values = body::getAll(key)::map(encodeBody);
```
gera2ld: Also `map` here is incorrect.

src/injected/web/requests.js
tophf: Why is map incorrect here? I've checked the original code once again and I don't see the issue.

As for TypeScript, it's also managed by a big bureaucratic entity so they won't implement support for this syntax until it's at least Stage-3 or 4. Do you plan to rewrite everything in TypeScript though?

src/injected/web/requests.js
tophf: Why? it's what the original code did.

src/injected/web/requests.js
gera2ld:
```js
// somewhere above
const map = {};

// Oops, map is not a function
await Promise.all([...body::keys()]::map(resolveKeyValues));
```

tophf: Geez, I'm blind. I've spotted `keys` a few times, but missed this one.

src/injected/web/requests.js
gera2ld: The original code is:
```js
const values = array.map(xxx);
```
Now:
```js
const values = array::map(xxx);
```
Which is the same as:
```js
const values = map.call(array, xxx);
```
`map` is different here! Quite confusing I think. I didn't figure it out until failed running my test script.

tophf: Renamed the variable to `idMap`.

tophf:
I think I found a potential problem and fixed it after you've already merged it. Will you cherry-pick this commit https://github.com/tophf/violentmonkey/commit/9c3122d416d7481e3a721f65d4335813fa7993ea or do I open a new PR?

gera2ld:
@tophf I'll do cherry-pick.

PR #633, opened 2019-10-06
improvement: auto-updatable default script template

tophf:
* Adds a new automatically calculated boolean flag in `options` that's truthy if the default script template was edited. When it's falsy an update of Violentmonkey can also update the default template automatically without overwriting user's customized value.
* Aligns the `@grant` values for better readability
* Adds a few common keys: `@version`, `@author`, `@description`

### DIFF
--- a/src/background/utils/options.js
+++ b/src/background/utils/options.js
@@ -31,12 +31,18 @@ const defaults = {
   },
   scriptTemplate: `\
 // ==UserScript==
-// @name New Script
-// @namespace Violentmonkey Scripts
-// @match {{url}}
-// @grant none
+// @name        New script {{name}}
+// @namespace   Violentmonkey Scripts
+// @match       {{url}}
+// @grant       none
+// @version     1.0
+// @author      -
+// @description {{date}}
 // ==/UserScript==
 `,
+  // Enables automatic updates to the default template with new versions of VM
+  /** @type {?Boolean} this must be |null| for template-hook.js upgrade routine */
+  scriptTemplateEdited: null,
 };
 let changes = {};
 const hooks = initHooks();
--- a/src/background/utils/script.js
+++ b/src/background/utils/script.js
@@ -75,6 +75,8 @@ export function getDefaultCustom() {
 export function newScript(data) {
   const state = {
     url: '*://*/*',
+    name: '',
+    date: new Date().toLocaleString(),
     ...data,
   };
   const code = getOption('scriptTemplate')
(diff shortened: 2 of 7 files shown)


src/background/utils/script.js
```diff
@@ -75,6 +75,8 @@ export function getDefaultCustom() {
 export function newScript(data) {
   const state = {
     url: '*://*/*',
+    name: '',
+    date: `${new Date()}`.split(' GMT', 1)[0],
```
gera2ld: According to MDN:

> Until ECMAScript 2018 (edition 9), the format of the string returned by `Date.prototype.toString` was implementation dependent. Therefore it should not be relied upon to be in the specified format.

maybe `new Date().toLocaleString()` is a better idea?

src/background/utils/options.js
```diff
@@ -98,6 +120,9 @@ export function setOption(key, value) {
       optionValue = objectSet(getOption(mainKey), keys.slice(1), value);
     }
     options[mainKey] = optionValue;
+    if (key === 'scriptTemplate') {
+      options[`${key}Edited`] = !!value && value !== defaults[key];
+    }
```
gera2ld: What about calculate this value in options page and pass it back?
I don't think such `if`-condition in a common function is a good idea. 😅

tophf: There are two places where `setOption` is called for this value (vm-template.vue and background/index.js). We would have to check in both places and remember to add a check to each such new place in the future so I decided it's too fragile.

tophf: Also, if we do it elsewhere, we would have to run an additional `setOption`, which is kinda wasteful in this case as the option is just an internal indicator that doesn't need to be broadcast.

tophf: We can use a declarative approach:
```js
const onSet = {
  scriptTemplate: (value, options) => { .... },
}
```
Then in setOption just before browser.storage.local.set:
```js
const hook = onSet[key];
if (hook) hook(optionValue, options);
```

gera2ld: There is something similar:

https://github.com/violentmonkey/violentmonkey/blob/c0a7ec8256f452fb83f2bbf947b534eb2e86e5e0/src/background/index.js#L28-L33

tophf: Ah, I didn't realize I can hijack this. So you're not concerned with the unnecessary processing and broadcasting of two options due to two setOption calls? Or maybe we can add a noBroadcast parameter?

gera2ld: The hook is debounced so it won't be called frequently. The `UpdateOptions` message is only posted to options page and popup page, so I won't worry about broadcasting.
I don't think `noBroadcast` is a good idea because it may lead to difference between data in different contexts.

tophf: Done.

PR #634, opened 2019-10-07
fix: don't continue line comments in some edge cases

### LINKED ISSUE #631: Is there a Setting that will prevent new lines from having // when I press Enter?
Some of my lines are commented out with //.  If my cursor is at the end of those lines and I press Enter, each new line will start with a //.  Yes, you don't need to tell me that I can put the cursor at the end of non-commented lines before pressing Enter.  But even if I try extremely hard to train myself to remember and always do that, there is still a chance that I will make the mistake or forget and then end up with a bunch of new lines with // at the beginning of them.  The thing is, I NEVER EVER need new lines to start off commented.  So I was wondering if there is a Setting so that I can turn that off, or if the ViolentMonkey developer could add a NEW Setting that will let me toggle that off.  Thanks.

tophf:
Fixes #631.

This is a local copy of the fix submitted to CodeMirror in case you'll want to release Violentmonkey before CodeMirror ships the fix, which isn't guaranteed to happen in its next release.

### DIFF
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -50,7 +50,11 @@
 import 'codemirror/lib/codemirror.css';
 import 'codemirror/theme/eclipse.css';
 import 'codemirror/mode/javascript/javascript';
-import 'codemirror/addon/comment/continuecomment';
+// Fixing a bug in CodeMirror's continuecomment-fix.js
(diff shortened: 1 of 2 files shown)


tophf:
The upstream fix was just merged so I guess it would make sense to postpone merging this and wait for a CM release.

PR #636, opened 2019-10-09
GM_addValueChangeListener, GM_removeValueChangeListener, and follow-up modularization

### LINKED ISSUE #483: Support Tampermonkey's GM_addValueChangeListener api
### What is the problem?

Scripts that are currently running in Tampermonkey have access to an api [GM_addValueChangeListener](https://tampermonkey.net/documentation.php#GM_addValueChangeListener) that enables a script executing in one tab to be notified when a script in another tab changes a value.  Adding support fir this API and [GM_removeValueChangeListener](https://tampermonkey.net/documentation.php#GM_removeValueChangeListener) will enable some scripts that can only run in TamperMonkey to also run in ViolentMonkey.

tophf:
* As a follow-up to the previous PR dealing with the sandbox, I've extracted a lot from the huge web/index.js and a little from content/index.js so now there are more cozy isolated modules.
* `bridge.addHandlers()` is now called inside the isolated modules. It's DRY now. And easier to switch between event handler and event emitter using the find usages/jump to definition IDE command.
* `bridge.addHandlers({CommandName: sendMessage})` will relay the request via sendMessage as is
* A questionable change might be that I put `ids` and friends inside `bridge` but it can be argued that the purpose of bridge is to be a helper for the scripts.

---

Fixes #483.

### GM_addValueChangeListener and GM_removeValueChangeListener

Run the test in https://example.org/, then open another tab with https://example.org/ and look at the first one's console.
```js
// ==UserScript==
// @name        test GM_addValueChangeListener
// @match       https://example.org/
// @grant       GM_addValueChangeListener
// @grant       GM_removeValueChangeListener
// @grant       GM_setValue
// @grant       GM_deleteValue
// @noframes
// ==/UserScript==

const cid = GM_addValueChangeListener('foo', console.log);
GM_setValue('foo', {foo: 123});
GM_setValue('foo', {foo: 456});
GM_deleteValue('foo');
GM_removeValueChangeListener(cid);
GM_setValue('foo', 'bar');
GM_deleteValue('foo');
```

---

### Trivial enhancement:
2019-10-10 17-55-44-fs8
The URL is properly redirected to the actual editor when opened in a tab.

P.S. I'd like to rebase this PR to hide/group all the trivial fixups and cosmetics when (and if) you okay it.


src/injected/content/inject.js
```diff
@@ -0,0 +1,158 @@
...
+    if (callback) callback();
+  }
+})(${attachFunction},`;
+
+function injectScript(data) {
+  const [vId, codeSlices, vCallbackId, mode, scriptId, scriptName] = data;
+  // trying to avoid string concatenation of potentially huge code slices as long as possible
+  const injectedCode = [
+    injectedScriptIntro,
+    `"${vId}"`, ',',
```
gera2ld: How about merging `,` with the adjacent string?

src/injected/content/inject.js
```diff
@@ -0,0 +1,158 @@
...
+  const id = getUniqId('VM-');
+  const detect = (domId) => {
+    const a = document.createElement('a');
+    a.id = domId;
+    document.documentElement.appendChild(a);
+  };
+  inject(`(${detect})(${jsonDump(id)})`);
+  const a = document.querySelector(`#${id}`);
+  const injectable = !!a;
+  if (a) a.parentNode.removeChild(a);
```
gera2ld: We can just use `a.remove()` now.

src/injected/web/gm-api.js
```diff
@@ -0,0 +1,220 @@
...
+
+const { getElementById } = Document.prototype;
+const { lastIndexOf } = String.prototype;
+const { hasOwnProperty } = Object.prototype;
+
+export function createGmApiProps() {
+  // these are bound to script data that we pass via |this|
+  const boundProps = {
+    GM_deleteValue(key) {
+      const { id } = this;
```
gera2ld: I prefer passing an context object explicitly, like what *koa v2* did, so that we don't have to concern about function bindings any more.
Is there any advantage to use `this` rather than pass an object explicitly?

tophf: I made an attempt to implement it explicitly as I knew you wouldn't like `this`, however it looked clumsy, un-JavaScript-like, actually Python-like.

src/injected/web/gm-values.js
```diff
@@ -0,0 +1,146 @@
...
+  });
+  if (localChanges.length && !cleanupTimer) {
+    cleanupTimer = setTimeout(cleanupChanges, CLEANUP_TIMEOUT);
+  }
+}
+
+// { hooks, key, val, raw, oldRaw, remote }
+function notifyChange(change) {
+  const {
+    hooks, key, val, raw, oldRaw, remote = false,
```
gera2ld: Where does this `val` come from?

tophf: `val` property is set in GM_setValue, then both dumpValue and notifyChange use the change object as is. I should probably rename `args` to `change` in other places for uniformity?

src/injected/web/gm-api.js
tophf: It's not like I have a strong opinion on this, so do you insist? :-)

src/injected/web/gm-values.js
```diff
@@ -0,0 +1,146 @@
...
+   * (remote=false), we need to avoid re-reporting with the wrong remote=true parameter,
+   * so we coalesce the more detailed local changes to match the remote changes.
+   * For a contrived example let's look at this local history:
+   *     undefined 'a' - created
+   *     'a' 'b'       - changed
+   *     'b' undefined - deleted
+   * The remote history is debounced so it might look like:
+   *     undefined 'a'
+   *     'a' undefined
+   */
```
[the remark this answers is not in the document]
tophf: 1. Current implementation matches TM.
2. Sounds like a standard data synchronization problem that should be solved by the userscript's author, not us.

src/injected/web/gm-api.js
gera2ld: Alright, it does not matter since all the functions are called with `.apply(selfData, args)`.

tophf: FWIW we can call all of them as gmFunction(context, ...args).
Like I said I'm not exactly *against* the more verbose approach.

src/injected/web/gm-values.js
gera2ld: Oh I see. Never mind. Poor JavaScript.

src/injected/content/inject.js
```diff
@@ -0,0 +1,160 @@
...
+  const id = getUniqId('VM-');
+  const detect = (domId) => {
+    const a = document.createElement('a');
+    a.id = domId;
+    document.documentElement.appendChild(a);
+  };
+  inject(`(${detect})(${jsonDump(id)})`);
+  const a = document.querySelector(`#${id}`);
+  const injectable = !!a;
+  if (a) a.remove(a);
```
gera2ld: should be `a.remove()` instead of `a.remove(a)`.

tophf:
While moving the logic to resolve simple URLs, I've noticed onBeforeRequest is invoked for all URLs but it can be scoped to `.user.js`. So I did it here in this PR since you merge the commits individually anyway.

tophf:
Following the discussion in #639:
* `GM_console` is a copy of the original console object with all the methods
* `GM_log` runs in the userscript scope so it prints the name of the script automatically on the right.

PR #638, opened 2019-10-10
ui-cosmetics: use variables to align things in popup reliably

tophf:
Visual changes:
* script icon is increased from 17px to 18px
* find icon is increased from 14px to 16px
* script names got shorter by ~1px
* `Violentmonkey` title is centered vertically to match the buttons and uses the global font size
* failure reason left margin is fixed (it wasn't properly updated in the icons PR)

Before | After
-|-
before-fs8 | after-fs8

Implementation notes:
* `postcss-custom-properties` plugin was hacked into postcss options directly as I couldn't find any other way due to my ignorance
* `px` are used for all the calculations so postcss-calc can resolve them at compile-time

### DIFF
--- a/src/popup/style.css
+++ b/src/popup/style.css
@@ -1,3 +1,24 @@
+$fontSize: 12px;
+$padding: 8px;
+$itemPaddingX: 14px;
+
+$scriptIconSize: 18px;
+$scriptIconMLR: 2px;
+
+$iconSize: 14px;
+$iconML: calc($scriptIconSize + 2 * $scriptIconMLR);
+$iconMR: 10px;
+
+$leftPaneWidth: calc($iconML + $iconSize + $iconMR);
+
+$logoSize: 32px;
+$logoMLR: calc(($leftPaneWidth - $logoSize) / 2);
+
+$topIconSize: 20px;
+
+$findIconSize: 16px;
+$findIconML: calc($leftPaneWidth - $findIconSize - $iconMR);
+
 body {
   width: 320px;
   max-width: 100%;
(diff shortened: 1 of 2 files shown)


gera2ld:
> `postcss-custom-properties` plugin was hacked into postcss options directly as I couldn't find any other way due to my ignorance

Let me handle this, there is an easier way. :)

> `px` are used for all the calculations so postcss-calc can resolve them at compile-time

We should use `px` everywhere and transform them with plugins if needed. My bad to write `rem` manually.

scripts/webpack.conf.js
```diff
@@ -13,6 +13,20 @@ module.exports = Promise.all([
...
+        if (loader === 'postcss-loader') {
+          options.plugins.unshift(
+            require('postcss-custom-properties')({
+              preserve: false
+            }),
+          );
+          break findPostCssRule;
+        }
+      }
+    }
```
gera2ld: Actually we don't have to do anything here, `postcss-custom-properties` is already supported in `precss`.
FYI postcss will respect `.browserslistrc`. If CSS variables are supported by all the browsers defined in `.browserslistrc`, they will no longer be transformed.

tophf: I don't want to preserve the variables in the final CSS as they add overhead and don't provide any runtime benefits.

gera2ld: In that case you may want sass-like syntax, see [precss](https://github.com/jonathantneal/precss). CSS variables are supposed to be preserved so that we can change theme with custom CSS easily by modifying the variables, without overriding CSS properties everywhere. I think here we should use sass variables for `px`s and CSS variables for colors. They are both supported in this project.

tophf: Done.

PR #639, opened 2019-10-11
GM_logging to different log-levels

MadeOfMagicAndWires:
This pull will make users able to specify the log-level `GM_log` logs to. If the user specifies a specific log-level as the first parameter it'll log any further parameters to that log-level rather than `console.log` (which is still the default).

Now this *does* (slightly) break backwards compatibility with other userscript managers, and I realise people might be a bit skittish about that, but thinking about it the only use-case in which that would feasibly happen is if script calls GM_log specifically with only a log-level, say `error`, as the first parameter, in which case they probably want to log to the appropriate log-level anyway.

Alternatively, we could write specific functions `GM_debug`, `GM_error`, `GM_info`, and `GM_warn` to avoid messing with `GM_log` but in my view this would break more with other managers, as they would not support these separate calls. 

There is an argument to be made about that being additional functionality, rather than messing with backwards compatibility, and so we wouldn't have to worry about other managers, so I'm more than willing to change it to work that way instead if that is preferred.

### DIFF
--- a/src/injected/web/index.js
+++ b/src/injected/web/index.js
@@ -338,7 +338,15 @@ function createGmApiProps() {
       }
     },
     GM_log(...args) {
-      log('log', [this.script.meta.name || 'No name'], ...args);
+      let loglevel = 'log';
+      if (args[0] === 'error'
+        || args[0] === 'debug'
+        || args[0] === 'info'
+        || args[0] === 'log'
+        || args[0] === 'warn') {
+        loglevel = args.shift();
+      }
+      log(loglevel, [this.script.meta.name || 'No name'], ...args);
     },
     GM_registerMenuCommand(cap, func) {
       const { id } = this;


gera2ld:
Besides compatibility issue as you said, this change relies on an implicit first argument to decide which level to use, likely leading to unexpected loss of the first argument. This is unacceptable for me.

I think there is another option: attach additional methods to `GM_log`.

```js
GM_log('log level');
GM_log.error('error level');
GM_log.warn('warn level');
```

And in order to keep consistency we can also add `GM_log.log` which is the same as `GM_log`.

tophf:
Technically it's possible to abuse `this` like this: `GM_log.call('error', 'Message', foo, bar)`
In VM's wrapGM there'll be something like:
```js
prop.value = name == 'GM_log'
  ? function fn(...args) { gmFunction.call(selfData, this, ...args); }
  : ((...args) => gmFunction.apply(selfData, args));
```

MadeOfMagicAndWires:
@gera2ld 
I did think about that, and what eventually led me to implement it this way is that if say you're calling GM_log like `GM_log('warn', 'Some warning message')` doesn't this achieve the same goal (arguably better), even if it unintentionally eats the "warn" parameter? I do see why your point though.

As for the sub-method implementation, I hadn't thought of that. 
I have tried to fiddle around with but I' woefully inexperienced with OOP in Javascript and I can't seem to get it to work with defining the function in a way which is both directly callable *and* has working submethods. 

I have a very hacky solution that works on its own but breaks by gmFunction.apply(), so if a very nice someone could spare some time to point me in the right direction that'd be great.

gera2ld:
@MadeOfMagicAndWires This method is already used by many userscripts in different script managers so we can't just change its parameters with such a breaking change.

tophf:
I can add whichever solution is best but remind me why do we have GM_log at all? Is it because some sites disable/break the console functions?

tophf:
And there's a bigger problem: VM is already breaking the ability to specify [console formatter](https://console.spec.whatwg.org/#formatter) as the first parameter like `GM_log('foo %o bar', {foo: 123});` because it adds its own "tags" to the output. I think this tagging should be removed because it can be trivially implemented by the userscript author if needed. Or we must add the tags to the first parameter if it's a formatter string.

gera2ld:
`GM_log` is added because the original Greasemonkey has such API and people use it.
AFAIK `GM_log` was never meant to be the same as `console` in Chrome, it accepted only one string in the original version, and VM extended it to support multiple strings for convenience.

VM adds a prefix to the output so that we can easily figure out this is a log from Violentmonkey, which is useful when debugging in a page with many console messages.

Anyway, `console` is always available for userscript authors if they want powerful features like formatters. So we don't have to care much about differences between `GM_log` and `console`, just that between different script managers.

@tophf `GM_console` looks good for me. But I am wondering whether we should add such a feature, or maybe it's better to tell users to use `console` directly. GM v4 seems to have removed `GM.log`.

tophf:
I don't have any preference on this because I've never encountered a site that mangles/removes console.log and such. That is, I always use the standard console.* methods.

MadeOfMagicAndWires:
That's fair enough,

I'll close this PR for now then. if I ever manage to get a different implementation working I might try again.

PR #643, opened 2019-10-12
Trivial editor improvements

### LINKED ISSUE #242: Feature request: More editor options?
### What is the problem?
CodeMirror provides many options such as indentUnit, tabSize, theme, highlight, etc. Plan to support some of the essentials?

### How to reproduce it?
1. Open any script
2. Try to find option to, say change indentation spaces

Some options to control the editor. For example, the one [Stylus](https://github.com/openstyles/stylus) provides (Stylus uses CodeMirror as well).
The "Custom Style" editor should share the options as well.

Currently Violentmonkey does not provide any option to control the editor.

### Environment
- Browser: Firefox
- Browser version: 58 nightly
- Violentmonkey version: 2.8.13
- OS: Windows 10

### LINKED ISSUE #596: Indent with tabs
Quite like #364 which introduced some configuration item for indentation with spaces.
Could we get a new setup to allow indentation with single tabulation characters?

> Indent with 🔘 2️⃣ Space characters or with 🔘 Tabulation character

tophf:
The built-in editor can be much more convenient for quick edits if we enable a few trivial options in CodeMirror.

Fixes #242.
Fixes #596.

* the CodeMirror's `sublime` key map is a very small addon that makes the editor not a total PITA to use.
* user-customizable editor options JSON to accommodate for some weird setups like indentWithTabs etc.
* `?` sub-tab with the active key map cheat sheet:
  

Trivial UX changes:
* <kbd>Esc</kbd> closes the sub-tab first (Settings, Values, ?)
* <kbd>F1</kbd> shows the hotkeys tab
* CodeMirror is focused when switching back from other tabs so it's possible to press F1 (view the hotkeys), then Esc, then just keep editing without using the mouse.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -115,6 +115,12 @@ descCustomCSS:
   message: >-
     Custom CSS for options page and script installation page. If you are not
     sure what this is for, please do not edit it.
+descEditorOptions:
+  description: Description of editor options JSON section.
+  message: >-
+    Custom CodeMirror options in JSON object like {"indentUnit":2, "smartIndent":true}
+    however note that Violentmonkey doesn't support all of them.
+    Full list: https://codemirror.net/doc/manual.html#config
 editLabelMeta:
   description: Metadata section in settings tab of script editor.
   message: Custom metadata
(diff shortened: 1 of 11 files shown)


gera2ld:
1. open editor and switch to the **?** tab.
2. click **Close**

Expected: close editor completely
Actual: switched to code editor

tophf:
Yep. Solved by introducing `fromCM` parameter. Not too keen about the name, though. Any better ideas?

gera2ld:
> Also what do you think about removing the two old editor options in settings UI? I've replaced them with JSON textarea where the users can customize whatever they want, which simplifies the code. It's not like many people change these anyway.

Totally agree. I think it's much better than before.

PR #652, opened 2019-10-14
Editor tweaks

tophf:
* toggle comments via <kbd>Ctrl</kbd><kbd>/</kbd>
* make /* block comment */ via <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>/</kbd>
* fix: <kbd>Tab</kbd> was incorrectly inserting a tab character when nothing is selected after #643
* fix: don't show missing CM commands in hotkeys tab
* a button to toggle display of current CodeMirror options in editor settings
  image

### DIFF
--- a/src/options/views/tab-settings/vm-editor.vue
+++ b/src/options/views/tab-settings/vm-editor.vue
@@ -5,6 +5,8 @@
     <setting-text name="editor" ref="editor" :json="true" />
     <button v-text="i18n('buttonSave')" @click="onSave"></button>
     <button v-text="i18n('buttonReset')" @click="onReset"></button>
+    <button v-text="i18n('buttonShowEditorState')" @click="toggleStateHint" />
+    <pre v-text="hint" class="monospace-font dim-hint" />
   </section>
 </template>
 
@@ -16,6 +18,9 @@ import SettingText from '#/common/ui/setting-text';
 import defaults from '#/common/options-defaults';
 
 export default {
+  data() {
+    return { hint: null };
+  },
   components: {
     SettingText,
   },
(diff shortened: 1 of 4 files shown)


src/options/views/tab-settings/vm-editor.vue
```diff
@@ -37,6 +42,47 @@ export default {
...
+        ...(await import('codemirror')).defaults,
+        ...(await import('#/common/ui/code')).cmOptions,
+        ...options.get('editor'),
+      })
+      .sort(([a], [b]) => (a < b ? -1 : a > b))
+      .forEach(([k, v]) => {
+        if (!HIDE_OPTS.includes(k) && typeof v !== 'function') {
+          opts[k] = v;
+        }
+      });
```
tophf: you seem to prefer a more verbose name like key/value, is it intentional and should I use it here too?

src/options/views/tab-settings/vm-editor.vue
```diff
@@ -37,6 +42,47 @@ export default {
...
+        'lineNumberFormatter',
+        'specialCharPlaceholder',
+      ];
+      const opts = {};
+      Object.entries({
+        ...(await import('codemirror')).defaults,
+        ...(await import('#/common/ui/code')).cmOptions,
+        ...options.get('editor'),
+      })
+      .sort(([a], [b]) => (a < b ? -1 : a > b))
```
gera2ld: What's the sorting for?

src/options/views/tab-settings/vm-editor.vue
gera2ld: It does not matter for such a simple case. I prefer `filter |> reduce` though.

src/options/views/tab-settings/vm-editor.vue
tophf: This code displays the current options object which has like 30 keys and they come in a random order. Sorting makes it easier to read.

jesus2099:
Hi @tophf.
Your improvements to the editor are awesome!
Thank you so much.

I have the following setting (light theme and tabs):

```javascript
{
  "indentUnit": 4,
  "indentWithTabs": true,
  "lineWrapping": false,
  "tabSize": 4,
  "theme": "default",
  "undoDepth": 200
}
```

But my settings are only followed when I open the Violentmonkey dashboard then one of the `</>` Edit buttons from there.
My settings are not followed if I press an `</>` Edit button, directly from Violentmonkey extension popup menu.

tophf:
@jesus2099 if you use Firefox, it's #873, fixed in the [release candidate](https://github.com/violentmonkey/violentmonkey/releases) which you can probably use until the new version comes out if you don't restart Firefox often (temporary extensions are reset to the original ones on restart).

jesus2099:
I noticed this bug in Vivaldi.
It's a minor issue for me so I can wait final 2.12.8 to see if that fix will also solve Vivaldi issue. :)

tophf:
That bug doesn't affect Chromium-based browsers and I couldn't reproduce the problem in Chrome but Vivaldi is known to have some weird extension-related bugs. Check the RC I linked and if you see the bug there as well, open a new issue.

jesus2099:
Wow it works! Thank you again.

For what it's worth, this 2.12.8 RC 2 has version 2.12.17 in its About panel.

tophf:
Yes the version is the same because I don't want to modify manifest.json for each RC.

PR #662, opened 2019-10-18
refactor VM- headers handling, use webRequest only when needed

tophf:
Attaches onBeforeSendHeaders listener only when a GM_xhr is active, automatically detaches when all done.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "@gera2ld/plaid": "~1.4.12",
     "@gera2ld/plaid-vue": "~1.4.9",
     "@gera2ld/plaid-webpack": "~1.5.0",
+    "@types/chrome": "latest",
     "cross-env": "^6.0.0",
     "del": "^5.1.0",
     "fancy-log": "^1.3.2",
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -1,11 +1,12 @@
 import {
-  getUniqId, request, i18n, buffer2string,
+  getUniqId, request, i18n, buffer2string, isEmpty,
 } from '#/common';
 import cache from './cache';
 import { isUserScript, parseMeta } from './script';
 import { getScriptByIdSync } from './db';
 import { openerTabIdSupported } from './tabs';
 
+const VM_VERIFY = 'VM-Verify';
 const requests = {};
 const verify = {};
 const specialHeaders = [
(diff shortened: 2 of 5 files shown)


src/background/utils/requests.js
```diff
@@ -188,58 +191,36 @@ function decodeBody(obj) {
...
-    types: ['xmlhttprequest'],
+const transformedHeaders = Object.create(null);
+function onBeforeSendHeaders({ requestHeaders: headers, requestId }) {
+  const reqId = (headers.find(h => h.name === VM_VERIFY) || {}).value;
+  const req = reqId && requests[reqId];
+  if (!reqId || !req) return { requestHeaders: headers };
+  verify[requestId] = reqId;
+  req.coreId = requestId;
+  // if the listener is to be unregistered, it must occur in the next event loop cycle
+  setTimeout(transformHeaders, 0, reqId);
```
tophf: Hmm, I wonder if onBeforeSendHeaders may be called multiple times during redirects? If so I need to unregister later, probably by using onResponseStarted and/or a bigger timeout.

src/background/utils/requests.js
```diff
@@ -188,58 +193,37 @@ function decodeBody(obj) {
...
+function onBeforeSendHeaders({ requestHeaders: headers, requestId }) {
+  // only the first call during a redirect/auth chain will have VM-Verify header
+  const reqId = (headers.find(h => h.name === VM_VERIFY) || {}).value || verify[requestId];
+  const req = reqId && requests[reqId];
+  if (!reqId || !req) return { requestHeaders: headers };
+  verify[requestId] = reqId;
+  req.coreId = requestId;
+  return {
+    requestHeaders: headers
+    .filter(({ name }) => name !== VM_VERIFY && (!req.anonymous || !/^cookie$/i.test(name)))
```
tophf: I wonder if `filter` should come after `concat` so that a `cookie` is removed from custom headers as well? OTOH it's the userscript's responsibility to provide sensible noncontradictory params...

tophf:
Added Stage-3 optional?.chaining?.operator :wink: hopefully you don't mind.

src/background/utils/requests.js
tophf: Nevermind, previously they were filtered so I'll just do it here as well.

.babelrc.js
```diff
@@ -16,5 +16,8 @@ module.exports = {
   ],
   plugins: [
     '@babel/plugin-proposal-function-bind',
+    // `loose:true` produces `foo==null` and it's fine because we'll never use `document.all`
+    // which is the only reason for the default `false`
+    ['@babel/plugin-proposal-optional-chaining', { loose: true }],
```
gera2ld: This is already supported in `@gera2ld/plaid/config/babelrc`.

package.json
```diff
@@ -18,10 +18,12 @@
   "description": "Violentmonkey",
   "devDependencies": {
     "@babel/plugin-proposal-function-bind": "^7.2.0",
+    "@babel/plugin-proposal-optional-chaining": "^7.6.0",
```
gera2ld: We don't need this.

.babelrc.js
tophf: Great!

package.json
```diff
@@ -22,6 +22,7 @@
     "@gera2ld/plaid": "~1.4.12",
     "@gera2ld/plaid-vue": "~1.4.9",
     "@gera2ld/plaid-webpack": "~1.5.0",
+    "@types/chrome": "latest",
```
gera2ld: `yarn.lock` should be updated.

tophf: Done.
P.S. turns out my IDE was using npm by default so I've switched it to yarn now.

tophf:
Had to force-push a few times because I accidentally added a bug yesterday while trying to harden the conditions for the listener to be registered. Fixed that and added a few comments that clarify the logic, hopefully.

PR #665, opened 2019-10-21
chore: update CodeMirror

tophf:


gera2ld:
`yarn.lock` should be updated.

tophf:
Updated.

PR #670, opened 2019-10-24
Show unique iframe scripts in popup

### LINKED ISSUE #664: List scripts from iframes
Currently VM's popup doesn't show scripts that are unique to iframes, that is those that don't run in the main page. This is inconvenient, especially if the main page URL is blacklisted.

I guess VM can show an additional panel in the popup per each iframe with scripts. The panel's caption will be the URL of the iframe.

Or maybe just one additional panel for all iframes with a list of all unique scripts that don't run in the main page.

tophf:
Fixes #664.

Adds `sendCmd(cmd, data)` method, which I can split off into a separate PR (and also convert to async/await) if you wish. 

Test links (also try blacklisting the main page URL):
* https://jsfiddle.net/chrisvfritz/50wL7mdz/
* https://chrome.google.com/webstore/detail/ocfjjjjhkpapocigimmppepjgfdecjkb

BTW what do you think about enabling Stage-3 [nullish-coalescing-operator](https://babeljs.io/docs/en/babel-plugin-proposal-nullish-coalescing-operator) and then replacing most of `objectGet(obj, 'foo.bar', def)` with `obj.foo?.bar ?? def` ? The only caveat I see is that technically it's not the same behavior as objectGet, which checks `key in res`, so if VM stores `null` or `undefined` somewhere in its data objects, we'd better not use `??` at all in order to avoid possible confusion.

### DIFF
--- a/src/popup/index.js
+++ b/src/popup/index.js
@@ -31,9 +31,8 @@ Object.assign(handlers, {
   async SetPopup(data, src) {
     if (store.currentTab.id !== src.tab.id) return;
     const isTop = src.frameId === 0;
-    const ids = isTop
-      ? data.ids
-      : (await mutex.ready, data.ids.filter(id => !allScriptIds.includes(id)));
+    if (!isTop) await mutex.ready;
+    const ids = data.ids.filter(id => !allScriptIds.includes(id));
     allScriptIds.push(...ids);
     if (isTop) {
       mutex.resolve();
--- a/src/popup/index.js
+++ b/src/popup/index.js
@@ -1,5 +1,5 @@
 import Vue from 'vue';
-import { i18n, sendMessage } from '#/common';
+import { i18n, sendCmd } from '#/common';
 import { INJECTABLE_TAB_URL_RE } from '#/common/consts';
 import handlers from '#/common/handlers';
 import * as tld from '#/common/tld';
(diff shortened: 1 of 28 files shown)


src/popup/style.css
```diff
@@ -30,6 +30,12 @@ body {
   }
 }
 
+.page-popup {
+  /* hardcoded popup height in Chrome */
+  max-height: 600px;
```
tophf: This is a kludge to hide the overflow when `.submenu` is tall. I couldn't find a proper CSS solution to limit `.submenu` height. And I didn't want to write JS-based height calc.

src/popup/index.js
```diff
@@ -17,26 +17,41 @@ const vm = new Vue({
...
+      mutex.resolve();
+      store.commands = Object.entries(data.menus)
+      .reduce((map, [id, values]) => {
+        map[id] = Object.keys(values).sort();
+        return map;
+      }, {});
+    }
+    if (ids.length) {
+      // frameScripts may be appended multiple times if iframes have unique scripts
+      store[isTop ? 'scripts' : 'frameScripts'].push(...await sendCmd('GetMetas', ids));
```
gera2ld: `SetPopup` may be called several times so we need to filter the duplicates.

A simple reproduction:
- create a script that registers a menu command:

  ```js
  // ==UserScript==
  // @name        do stuff
  // @namespace   Violentmonkey scripts
  // @match       *://*/*
  // @grant       GM_registerMenuCommand
  // ==/UserScript==

  GM_registerMenuCommand('do stuff', () => {});
  ```
- reload a page and open VM popup while the page is still loading.

tophf: Done.

PR #675, opened 2019-10-31
Group rendering the Vue way

gera2ld:
## Changes
- Remove complicated stages since we just care about whether the item should be rendered.
- Data driven. States are calculated from plain values.
- Control the group rendering flow in a simple loop, which is more intuitive.

## Performance
- v2.11.2
image

- Before this PR
image

- After this PR
image

Time cost of each frame is acceptable, and no significant difference from before.

ps. I didn't know that virtual DOM could be so slow in such cases, especially when the number of elements is not very large I guess.


tophf:
There is one small but moderately annoying regression - the first displayed frame never includes the first batch so there's always a delay in the attention focus area (which is likely at the top of the list) on opening or switching from settings or about tab. In my implementation most of the times the first frame included the first batch.

* before:
  
* after:

gera2ld:
Fixed by setting the initial `limit` to the cached `step`.

tophf:
Yep.

PR #678, opened 2019-11-02
Sort & filtering in popup

tophf:
Since I've never needed to care about the execution order in all the years I've been using userscripts I have a reason to believe the same applies to most users, at least to a lot of them, so this PR adds several options to display the scripts in the popup in a more human-friendly way, mostly useful when there are many scripts installed.

* alphabetical order (default: execution)
* enabled scripts first (default: false)
* disabled scripts hidden (default: false)

The default state preserves the historical behavior.

image

**Collateral change:**

* Removed the implicit space between the checkbox and its text in settings + aligned them vertically against each other.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -331,6 +331,9 @@ labelNoSearchScripts:
 labelNotifyUpdates:
   description: Option to show notification when script is updated.
   message: Notify script updates
+labelPopupSort:
+  description: Label in the VM settings tab for script list sort order in popup
+  message: Sort scripts in popup by
 labelRelated:
   description: Label of related links.
   message: 'Related links: '
@@ -542,6 +545,12 @@ msgUpdating:
 noValues:
   description: Label shown when there is no value for current script.
   message: No value is stored
+optionPopupEnabledFirst:
+  description: Option to show enabled scripts first in popup.
+  message: Enabled first
+optionPopupHideDisabled:
+  description: Option to hide disabled scripts in popup.
+  message: Hide disabled
 optionShowEnabledFirst:
   description: Option to show enabled scripts first in alphabetical order.
   message: Show enabled scripts first
--- a/src/common/options-defaults.js
+++ b/src/common/options-defaults.js
@@ -20,7 +20,14 @@ export default {
   version: null,
   defaultInjectInto: 'page', // 'page' | 'auto',
   filters: {
+    /** @type 'exec' | 'alpha' | 'update' */
     sort: 'exec',
+  },
+  filtersPopup: {
+    /** @type 'exec' | 'alpha' */
+    sort: 'exec',
+    enabledFirst: false,
+    hideDisabled: false,
   },
   editor: {
     lineWrapping: false,
--- a/src/options/views/tab-settings/index.vue
+++ b/src/options/views/tab-settings/index.vue
@@ -41,6 +41,21 @@
         </label>
         <a class="ml-1" href="https://violentmonkey.github.io/2018/11/23/inject-into-context/" target="_blank" rel="noopener noreferrer" v-text="i18n('learnInjectionMode')"></a>
       </div>
+      <div class="mb-1 multi-opt-row">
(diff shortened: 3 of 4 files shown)


gera2ld:
It does not display well when lines wrap.

image

tophf:
* Added `flex-wrap` on the last line.
* I don't see why vertical centering on the second line is a problem (IMO it looks fine).

gera2ld:
> * I don't see why vertical centering on the second line is a problem (IMO it looks fine).

If there are a few lines of text beside a checkbox, it's hard to tell where to start. (I guess we don't want to add separator lines between them.)

```
[ ] first line
    second line
[ ] third line
    fourth line
```

Intuitively I would say `first line\nsecond line` is for the first checkbox, and `third line\nfourth line` is for the second checkbox. This is also the default behavior in lists (`<ul>`).
But if the checkbox is aligned in the middle, `second line\nthird line\nfourth line` should be the content of the second checkbox.

There could be different margins, but I still think it hard to find the start line at first sight.

tophf:
The problem I was trying to solve is the bad vertical centering of checkboxes against the associated text label. It looks better when the text is not wrapped, which arguably constitutes 99% of use cases. I think the 1% of cases where it's wrapped due to narrow width can be ignored. Mm?

tophf:
I'd rather rephrase the overly long labels e.g. "Reload the tab when toggling a script in the popup"

gera2ld:
> I think the 1% of cases where it's wrapped due to narrow width can be ignored.

We support running on Firefox Android, so this is not the 1% of cases. Display on mobile devices is always taken into account.

tophf:
Ah, indeed.
I've added a media restraint on min-width.

tophf:
BTW extensions are also supported in mobile Chrome forks such as Yandex and Kiwi.

gera2ld:
The style change here has some problems in other languages because of extra margins. I'm gonna fix it by removing the flex layout and keep white spaces as before.

tophf:
What problems exactly? That space is a bug itself so a proper fix should remove it. An alternative method to flexbox is to start the next tag immediately, not on the next line, or use a comment:

```html
    <foo .....></foo><!--
 --><bar ...></bar>
```

tophf:
There's an [option to strip spaces](https://github.com/vuejs/vue/tree/dev/packages/vue-template-compiler#options) in vue template compiler, which looks like it should help here and everywhere.

gera2ld:
Some languages need space between words, e.g. English, and some not, e.g. Chinese. So it looks weird as `select { margin-left: .5em; }` is added. That's why I kept the white spaces in locale files.

After setting `font-size: 1rem` for `<select>`s and adding `vertical-align: middle`, it looks fine.
It's strange that I found many cases from StackOverflow seem to have worked in the past but do not work any more.

> There's an [option to strip spaces](https://github.com/vuejs/vue/tree/dev/packages/vue-template-compiler#options) in vue template compiler, which looks like it should help here and everywhere.

This option is already on. It seems that spaces are still kept in development mode, which might be a bug.

gera2ld:
Bootstrap gave up on this alignment problem. 😃 https://github.com/twbs/bootstrap/issues/18797

tophf:
I see `collapseWhitespace: true` and `preserveWhitespace: false` but the option I've linked is `whitespace: 'condense'` - is there any difference? I've tried adding it in [plaid-vue/webpack/vue.js](https://github.com/gera2ld/plaid/blob/d07a51149257ededd438dbe0d5f450d0d3b1e9d5/packages/plaid-vue/webpack/vue.js#L15) and it removed the space in dev build.

As for vertical-align, it was worsening things in Firefox when I tried it which is why I've used flexbox. Did you test your fix in FF?

gera2ld:
I've tested in Chrome / Firefox / Vivaldi, trying to test in mobile.

gera2ld:
@tophf I was wrong, I was using the deprecated `preserveWhitespace` option. I'll try the new one later.

PR #682, opened 2019-11-06
Fix WebDAV sync to Nextcloud

jpumc:
Nextcloud WebDAV doesn't return the DAV:displayname prop. Using filename from DAV:href.

Tested using Nextcloud 16.0.5.

### DIFF
--- a/src/background/sync/webdav.js
+++ b/src/background/sync/webdav.js
@@ -234,13 +234,24 @@ const WebDAV = BaseService.extend({
       .map((node) => {
         const prop = node.find('DAV:propstat').find('DAV:prop');
         const type = prop.find('DAV:resourcetype').find('DAV:collection') ? 'directory' : 'file';
-        const displayName = prop.find('DAV:displayname').text();
-        if (type === 'file' && isScriptFile(displayName)) {
-          const size = prop.find('DAV:getcontentlength');
-          return normalize({
-            name: displayName,
-            size: size ? +size.text() : 0,
-          });
+        if (type === 'file') {
+          let displayName;
+          const displayNameNode = prop.find('DAV:displayname');
+
+          if (displayNameNode !== undefined) {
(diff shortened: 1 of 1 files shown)


src/background/sync/webdav.js
```diff
@@ -234,13 +234,20 @@ const WebDAV = BaseService.extend({
...
-        if (type === 'file' && isScriptFile(displayName)) {
-          const size = prop.find('DAV:getcontentlength');
-          return normalize({
-            name: displayName,
-            size: size ? +size.text() : 0,
-          });
+        if (type === 'file') {
+          const href = node.find('DAV:href').text();
+          let displayName = href.substring(href.lastIndexOf('/') + 1);
+          const displayNameNode = prop.find('DAV:displayname');
```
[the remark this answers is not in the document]
jpumc: Changed it.

src/background/sync/webdav.js
```diff
@@ -234,13 +234,24 @@ const WebDAV = BaseService.extend({
...
-          });
+        if (type === 'file') {
+          let displayName;
+          const displayNameNode = prop.find('DAV:displayname');
+
+          if (displayNameNode !== undefined) {
+            displayName = displayNameNode.text();
+          } else {
+            const href = node.find('DAV:href').text();
+            displayName = href.substring(href.lastIndexOf('/') + 1);
```
gera2ld: Do we need to decode the name if it contains special chars like whitespace or unicodes? I'm not sure what the URL is like though.

gera2ld: I just did a test and found it encoded, so I think `decodeURIComponent` is needed for `displayName` here.

jpumc: I also tested it and with `decodeURIComponent` it now works with some filenames that get encoded.

PR #687, opened 2019-11-07
fix: GM_setClipboard in FF

### LINKED ISSUE #685: GM_setClipboard() has stopped working
### What is the problem?
I feel like GM_setClipboard() has stopped working.
Is this a bug or is this the intended specification?
Personally, I hope GM_setClipboard() will be kept.

### How to reproduce it?
1.use GM_setClipboard("foo") in any userscript

foo is copied to the clipboard

Nothing happens

### Environment
- Browser: Firefox Dev x64
- Browser version: 71.0b7
- Violentmonkey version: 2.12.0
- OS: Windows7 x64

tophf:
Fixes #685.

In addition to a bug in my implementation (preventDefault and stopImmediatePropagation should be picked from Event.prototype) there's apparently a bug/quirk in Firefox which prevents the entire setClipboard+onCopy hack from working when ::safe calls are used. ~I guess Firefox somehow changes the initial methods or it uses a Proxy internally on `document` specifically. It's a mystery for me because in other parts of the content script the similar safe calls work correctly with `window::addEventListener` and `document::getReadyState`.~ 

Turns out there was another bug in my code + old versions of FF define execCommand in document.__proto__ not in Document.prototype...

### DIFF
--- a/src/injected/content/clipboard.js
+++ b/src/injected/content/clipboard.js
@@ -3,10 +3,12 @@ import { sendCmd } from '../utils';
 import { addEventListener, logging } from '../utils/helpers';
 import bridge from './bridge';
 
-const { execCommand } = Document.prototype;
+// old Firefox defines it on a different prototype so we'll just grab it from document directly
+const { execCommand } = document;
 const { setData } = DataTransfer.prototype;
 const getClipboardData = Object.getOwnPropertyDescriptor(ClipboardEvent.prototype, 'clipboardData').get;
-const { preventDefault, removeEventListener, stopImmediatePropagation } = EventTarget.prototype;
+const { preventDefault, stopImmediatePropagation } = Event.prototype;
+const { removeEventListener } = EventTarget.prototype;
 
 let clipboardData;
 
@@ -27,7 +29,7 @@ function onCopy(e) {
   e::stopImmediatePropagation();
   e::preventDefault();
   const { type, data } = clipboardData;
-  e::getClipboardData::setData(type || 'text/plain', data);
+  e::getClipboardData()::setData(type || 'text/plain', data);
 }
 
 function setClipboard({ type, data }) {


tophf:
Interestingly, different versions of FF seem to fail in different parts of the code even when it's fixed to use the correct source for preventDefault and stopImmediatePropagation, which reinforces my guess it's a bug or an internal quirk of FF.

src/injected/content/clipboard.js
```diff
@@ -1,13 +1,8 @@
 import { isFirefox } from '#/common/ua';
 import { sendCmd } from '../utils';
-import { addEventListener, logging } from '../utils/helpers';
+import { logging } from '../utils/helpers';
 import bridge from './bridge';
 
-const { execCommand } = Document.prototype;
-const { setData } = DataTransfer.prototype;
-const getClipboardData = Object.getOwnPropertyDescriptor(ClipboardEvent.prototype, 'clipboardData').get;
-const { preventDefault, removeEventListener, stopImmediatePropagation } = EventTarget.prototype;
```
gera2ld: The real problem is, `preventDefault` and `stopImmediatePropagation` comes from `Event.prototype`. I corrected it locally and it worked in Firefox!

gera2ld: Sorry, there is one more thing. The getter `clipboardData` should also be kept.

gera2ld:
I am wondering whether it's safe to transform getters into functions and call them on the original objects. I think this may involve some implementation details of browsers. Maybe we should skip guarding the getters?

src/injected/content/clipboard.js
```diff
@@ -23,18 +18,20 @@ bridge.addHandlers({
...
 
+// not using ::safe calls as it mysteriously doesn't work here in FF (probably a bug)
+
 function onCopy(e) {
-  e::stopImmediatePropagation();
-  e::preventDefault();
+  e.stopImmediatePropagation();
+  e.preventDefault();
   const { type, data } = clipboardData;
-  e::getClipboardData::setData(type || 'text/plain', data);
```
gera2ld: It should be `e::getClipboardData()` instead of `e::getClipboardData` here.

tophf: Gee, I **really** can be blind, but this still doesn't help in Firefox ESR: like I explained above FF fails in different places and ESR won't even run the onCopy listener with document::addEventListener, only with document.addEventListener

tophf: Eek, FF ESR defines `execCommand` in `document.__proto__` not in `Document.prototype`...

src/injected/content/clipboard.js
```diff
@@ -3,10 +3,13 @@ import { sendCmd } from '../utils';
 import { addEventListener, logging } from '../utils/helpers';
 import bridge from './bridge';
 
-const { execCommand } = Document.prototype;
+// some old versions of Firefox define it on __proto__
+// eslint-disable-next-line no-proto
+const execCommand = Document.prototype.execCommand || document.__proto__.execCommand;
```
gera2ld:
```
document
  |
HTMLDocument - (Firefox ESR) execCommand
  |
Document - (Chrome and newer Firefox) execCommand
```

So we can just read `execCommand` from `HTMLDocument.prototype` or even `document` since `document` cannot be modified at `document-start`.

tophf: Yeah, good idea.

PR #690, opened 2019-11-08
Editor tweaks

tophf:
* fix: reuse indent type/size for Tab key without selection in the middle of a line
* feat: autofill search query with selected text like normal editors/IDE do
  * main find key - always fills
  * next/prev key - only if query is empty

### DIFF
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -342,6 +342,17 @@ export default {
         }
       }, this.cm);
     },
+    findFillQuery(force) {
+      const { state } = this.search;
+      const { cm } = this;
+      if (!state.query || force) {
+        const sel = cm.listSelections();
+        // use the currently selected text if it's within one line
+        if (sel?.length === 1 && sel[0].anchor.line === sel[0].head.line) {
+          state.query = cm.getSelection();
+        }
+      }
+    },
     doSearch(reversed) {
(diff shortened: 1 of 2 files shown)


src/common/ui/code.vue
```diff
@@ -90,12 +90,16 @@ function getHandler(key) {
...
   'Shift-Ctrl-/': 'commentSelection',
 });
 CodeMirror.commands.commentSelection = cm => {
   cm.blockComment(cm.getCursor('from'), cm.getCursor('to'), { fullLines: false });
 };
+// pressing Tab key inside a line with no selection will reuse indent type (tabs/spaces)
+const { insertTab, insertSoftTab } = CodeMirror.commands;
+CodeMirror.commands.insertTab = () => (
+  this.cm.options.indentWithTabs ? insertTab() : insertSoftTab()
+);
```
gera2ld: `this` does not work here and commands must be called with `cm` as a parameter:

```js
CodeMirror.commands.insertTab = cm => (
  cm.options.indentWithTabs ? insertTab(cm) : insertSoftTab(cm)
)
```

Didn't dig into the code so I'm not sure if there are other potential problems here.

tophf: It was a bad copypaste, sorry.

PR #691, opened 2019-11-08
Spanish translation update

Freddynic159:
- Fixed some messages.
- Added some missing messages.

### DIFF
--- a/src/_locales/es/messages.yml
+++ b/src/_locales/es/messages.yml
@@ -4,8 +4,7 @@ anchorAuthor:
 anchorTranslator:
   description: Translator shown on about tab.
   message: >-
-    <a href="http://greydus.tk" target="_blank" rel="noopener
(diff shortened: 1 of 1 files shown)


src/_locales/es/messages.yml
```diff
@@ -4,8 +4,7 @@ anchorAuthor:
 anchorTranslator:
   description: Translator shown on about tab.
   message: >-
-    <a href="http://greydus.tk" target="_blank" rel="noopener
-    noreferrer">Greydus</a>
+    <a href="http://greydus.tk" target="_blank" rel="noopener noreferrer">Greydus</a>
```
gera2ld: This change is unnecessary, the style is generated automatically by YAML tools. The values they represent have no difference. It will get back if we normalize all locale files the next time.

You can try `yarn i18n` if you are familiar with the Node.js tool chain.

PR #694, opened 2019-11-08
display errors in content script mode

### LINKED ISSUE #692: [Suggestion][Firefox] Wrap code of userscript in try..catch when using injection into the content context
When using `@inject-into content` mode in Firefox, it is difficult to debug scripts, since errors are not displayed in the console. But this can be solved by using try..catch:
```
try {
    // code of userscript
} catch (e) {
    console.error("==UserScript==", e);
}
```

tophf:
* Fixes #692: shows uncaught errors for `@inject-into content` mode
* Restores script name in GM_log (shown as the source at the right side of devtools console) lost in 8f43c1e65a2809f6c8a5a681ed2d91eb247a12f7

### DIFF
--- a/src/injected/web/gm-api.js
+++ b/src/injected/web/gm-api.js
@@ -174,7 +174,6 @@ export function createGmApiProps() {
       };
       return el;
     },
-    GM_log: logging.log,
     GM_openInTab(url, options) {
       const data = options && typeof options === 'object' ? options : {
         active: !options,
@@ -207,6 +206,8 @@ export function createGmApiProps() {
       target[k] = propertyFromValue(target[k]);
     });
   });
+  // not using propertyFromValue to keep native toString on the real console.log
+  props.GM_log = { value: logging.log };
   return {
     props,
     boundProps,
--- a/src/injected/web/gm-api.js
+++ b/src/injected/web/gm-api.js
@@ -227,13 +227,8 @@ export function createGmApiProps() {
 }
 
 export function propertyFromValue(value) {
-  const prop = {
-    writable: false,
-    configurable: false,
-    value,
-  };
   if (typeof value === 'function') value.toString = propertyToString;
-  return prop;
+  return { value };
 }
 
 function propertyToString() {
--- a/src/injected/web/index.js
+++ b/src/injected/web/index.js
@@ -20,6 +20,7 @@ export default function initialize(
 ) {
   let invokeGuest;
   bridge.props = props;
+  bridge.isFirefox = isFirefox;
   if (invokeHost) {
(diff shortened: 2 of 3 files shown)


src/injected/web/gm-api.js
```diff
@@ -226,13 +238,8 @@ export function createGmApiProps() {
 }
 
 export function propertyFromValue(value) {
-  const prop = {
-    writable: false,
-    configurable: false,
```
tophf: These are `false` [by default](https://developer.mozilla.org//docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Description).

src/injected/web/gm-api.js
```diff
@@ -201,12 +201,24 @@ export function createGmApiProps() {
...
+  }
   // convert to object property descriptors
   [props, boundProps].forEach(target => {
     objectKeys(target).forEach(k => {
       target[k] = propertyFromValue(target[k]);
     });
   });
+  // defining manually to avoid setting toString on the real console.log
+  if (bridge.mode !== INJECT_CONTENT) {
+    props.GM_log = { value: logging.log };
```
gera2ld: We need to wrap `console.log` into a new function to avoid `console.log` being modified. (#688)

gera2ld: Also I prefer consistent behaviors of `GM_log` no matter it's in page context or content context. Otherwise users have to figure out the difference and learn about `@inject-into`.

tophf: 1. No, we don't need to wrap it since we don't change its toString, I thought the comment I've added on line 218 is obvious but I guess I should improve it?
2. The old internal behavior of GM_log with the "[tags]" is a hack which was used only because the script name wasn't displayed. Since the `content` mode is also a hack it's only natural that two hacks are used together, but normal scripts in page mode don't suffer from this deficiency and use normal logging with a proper script name displayed as the source.

tophf: On the other hand, GM_log itself is a rarely used thing which I never needed so I don't have a strong opinion. Let's go with your preference.

gera2ld: I think I see why you insist to provide the original `console.log` now, it seems that the line info of where `console.log` is exactly called will be displayed on the right. So if we wrap it into a new function, the line info will always be that of `GM_log` definition.

Actually the source URL works also for `content` mode, I'll make a PR for that before merging this one.

tophf:
* Since the script name is displayed in content mode, try/catch has been moved inside the generated code.
* console.log fix has been reapplied using the new method (by skipping propertyFromValue)

PR #695, opened 2019-11-09
fix: show source URL in content mode

gera2ld:


tophf:
Awesome!

FF ESR still shows an incorrect source line (it links into its own internal API code) but I guess we can ignore it because 1) ESR+VM usage is low, 2) ESR will eventually update to a version that shows the source line correctly.

PR #698, opened 2019-11-09
chore: shrink injected* by 20%

tophf:
* added a missing DefinePlugin for injected-web.js to strip 1.6kB of 'process' dep
* converted functions to arrows in terser in injected*.js
* using array destructuring in params to strip 0.5kB of babel's _iterableToArrayLimit added for `const [foo] = bar` assignments

22->19kB `–17%` injected-web.js
17->11kB `–35%` injected.js

~I don't know how to apply DefinePlugin to the entire project to remove the `process` dep from common.js as well.~

### DIFF
--- a/scripts/plaid.conf.js
+++ b/scripts/plaid.conf.js
@@ -38,10 +38,10 @@ exports.pages = {
   },
 };
 
-const splitVendor = name => ({
-  [name]: {
-    test: new RegExp(`node_modules[/\\\\]${name}`),
-    name: `public/lib/${name}`,
+const splitVendor = prefix => ({
+  [prefix]: {
+    test: new RegExp(`node_modules[/\\\\]${prefix}.*?\\.js`),
+    name: `public/lib/${prefix}`,
     chunks: 'all',
     priority: 100,
   },
--- a/scripts/plaid.conf.js
+++ b/scripts/plaid.conf.js
@@ -36,9 +36,6 @@ exports.pages = {
     entry: './src/popup',
     html: htmlFactory(),
   },
-  injected: {
-    entry: './src/injected',
-  },
 };
 
 const splitVendor = name => ({
@@ -60,12 +57,7 @@ exports.optimization = {
         name: 'common',
         minChunks: 2,
         enforce: true,
-        chunks(chunk) {
-          return ![
-            'browser',
-            'injected',
-          ].includes(chunk.name);
-        },
+        chunks: chunk => chunk.name !== 'browser',
       },
       ...splitVendor('codemirror'),
       ...splitVendor('tldjs'),
--- a/scripts/webpack.conf.js
+++ b/scripts/webpack.conf.js
@@ -1,47 +1,95 @@
 const { modifyWebpackConfig, shallowMerge, defaultOptions } = require('@gera2ld/plaid');
+const { isProd } = require('@gera2ld/plaid/util');
 const webpack = require('webpack');
 const WrapperWebpackPlugin = require('wrapper-webpack-plugin');
+const TerserPlugin = require('terser-webpack-plugin');
 const projectConfig = require('./plaid.conf');
+const mergedConfig = shallowMerge(defaultOptions, projectConfig);
 
 const INIT_FUNC_NAME = 'VMInitInjection';
 
-module.exports = Promise.all([
-  modifyWebpackConfig(async (config) => {
-    config.plugins.push(
-      new webpack.DefinePlugin({
-        'process.env.INIT_FUNC_NAME': JSON.stringify(INIT_FUNC_NAME),
-        'process.env.DEBUG': JSON.stringify(process.env.DEBUG || false),
-      }),
-    );
-    return config;
+const definitions = new webpack.DefinePlugin({
+  'process.env.INIT_FUNC_NAME': JSON.stringify(INIT_FUNC_NAME),
+  'process.env.DEBUG': JSON.stringify(process.env.DEBUG || false),
+});
+const minimizerOptions = {
+  cache: true,
+  parallel: true,
+  sourceMap: true,
+  terserOptions: {
+    output: {
+      ascii_only: true,
+    },
+  },
+};
+const minimizer = isProd && [
+  new TerserPlugin({
+    chunkFilter: ({ name }) => !name.startsWith('public/'),
+    ...minimizerOptions,
+    terserOptions: {
+      ...minimizerOptions.terserOptions,
+      compress: {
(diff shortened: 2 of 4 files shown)


scripts/webpack.conf.js
```diff
@@ -1,47 +1,76 @@
...
   }, {
+    'injected-web': {
+      entry: './src/injected/web',
+    },
+  }),
+  modify(null, {
+    injected: {
+      entry: './src/injected',
+    },
+  }),
```
gera2ld: Why is `injected` splitted into a new webpack config?

gera2ld:
> I don't know how to apply DefinePlugin to the entire project to remove the `process` dep from common.js as well.

It should have been added to all other entries. Only `inject-web` is missed. 😅

scripts/webpack.conf.js
tophf: Since I don't know how to apply custom Terser settings by file mask, I've moved `injected` here.

tophf:
Stripped another 0.5kB of babel's _iterableToArrayLimit added for `const [foo] = bar` assignments by switching to array destructuring in function params. Can be checked by searching for `Symbol.iterator in Object` in dist.

gera2ld:
There are quite a few features marked as supported since Firefox 53. 😢 Maybe we should consider change the minimum support version for Firefox to 53.

See https://github.com/babel/babel/blob/d413a3078c1b7a46be106bebf67816c5f9550fc2/packages/babel-preset-env/data/plugins.json

gera2ld:
Changes in JavaScript from Firefox 52 to 53 are listed here: https://developer.mozilla.org/zh-CN/docs/Mozilla/Firefox/Releases/53#JavaScript

tophf:
Switching to `53` seems a better solution since it's the minimum version in many Babel's tranforms and it would reduce injected* even more without the need to explicitly care about array destructuring.

gera2ld:
The problem is, we still [have users in Firefox 52](https://addons.mozilla.org/en-US/firefox/addon/violentmonkey/statistics/usage/applications/?last=30). Firefox 52 seems to be the last version supporting Windows XP and Vista so we may not expect them to have an upgrade.

However, according to the release note, the features we rely on are already supported in Firefox 52.  So we could build the application for Firefox >= 53, but leave the min_version to 52 and do some tests to see if there is a need to notify the users.

tophf:
Sounds like a good plan. Those transforms weren't enabled in 52 only because of a few edge case bugs, which we probably won't encounter as we don't use those edge cases.

scripts/webpack.conf.js
```diff
@@ -1,47 +1,76 @@
...
+        minimizer: isProd && [
+          new TerserPlugin({
+            cache: true,
+            parallel: true,
+            sourceMap: true,
+            terserOptions: {
+              compress: {
+                ecma: 6,
+                // 'safe' since we don't rely on function prototypes
+                unsafe_arrows: true,
```
gera2ld: I reviewed the source code just now, and found `unsafe_arrows` is safe to be applied for all files in `src` except those in `src/public`. So there is no need to split `inject` from others?

tophf: I'm still a noob with webpack so you decide, you're the expert :-)

tophf: Turns out splitting of `injected` strips the big unused things like [request](https://github.com/violentmonkey/violentmonkey/blob/master/src/injected/utils/index.js#L6) (3.5kB difference, 14.5 vs 11). Removing `request` there doesn't help.

tophf: I also think it's okay to separate `injected` because it's not related to anything in `plaid`.

tophf: I found `chunkFilter` option in TerserPlugin and updated the PR.

scripts/webpack.conf.js
```diff
@@ -1,47 +1,95 @@
...
+      ...minimizerOptions.terserOptions,
+      compress: {
+        ecma: 6,
+        // 'safe' since we don't rely on function prototypes
+        unsafe_arrows: true,
+      },
+    },
+  }),
+  new TerserPlugin({
+    chunkFilter: ({ name }) => name.includes('public/'),
```
gera2ld: `includes` -> `startsWith` to make it stricter

scripts/plaid.conf.js
```diff
@@ -36,9 +36,6 @@ exports.pages = {
     entry: './src/popup',
     html: htmlFactory(),
   },
-  injected: {
-    entry: './src/injected',
-  },
 };
 
 const splitVendor = name => ({
```
gera2ld: `test` in `splitVendor` should be stricter, by adding trailing `/`.

tophf: Why? I intentionally omit it to include all related repos e.g. lots of vue*.

scripts/webpack.conf.js
```diff
@@ -1,47 +1,95 @@
...
+  },
+);
+
+module.exports = Promise.all([
+  modify(),
+  modify({
+    pages: {
+      injected: {
+        entry: './src/injected',
+      },
```
gera2ld: If not splitted, the generated `injected.js` will be a little smaller as I tested. Because it has to bundle a copy of runtime when splitted.

tophf: I don't understand. In my tests it's the other way around: here `injected` is 11kB, when it's in the old place in plaid it's 14.5kB. What "runtime" does it have to bundle if there's an explicit `runtimeChunk: false` in the config?

gera2ld: Sorry, I was wrong, it's bundled with a copy of common utils.
For me it's 10.33KB (non-splitted) vs 11.11KB (splitted).

scripts/plaid.conf.js
gera2ld: OK, then I prefer `prefix` rather than `name` as the parameter. I noticed this because a small CSS file `public/lib/vue.css` is generated and I thought it's unnecessary.

scripts/webpack.conf.js
gera2ld: Sorry, I was still wrong. You are right about it. 😭

tophf: I can't reproduce 10.33kB. If I restore the old plaid.conf.js and remove `modify` call for `entry: './src/injected'`, I get 14.5kB.

scripts/plaid.conf.js
tophf: Yep, renamed and added restriction on `.js`.

PR #701, opened 2019-11-09
feat: search in script code, /regexp/, lowercase on demand

tophf:
With regexp it's possible to search for `/@grant\s+none/` for example.


tophf:
Revealed `enabledFirst` option for lastUpdateTime too as it seems to be useful and refactored the sort-related code to make it a bit more declarative.

src/options/views/tab-installed.vue
```diff
@@ -378,6 +380,39 @@ export default {
...
+        } catch (err) {
+          searchError = err.message;
+        }
+      }
+      this.searchError = searchError || null;
+      const lowerSearch = !searchRE && (this.search || '').toLowerCase();
+      scripts.forEach(({ $cache }) => {
+        $cache.show = $cache.search.some((item, i, arr) => {
+          if (typeof item !== 'string') {
+            item = item[0].toLowerCase();
```
gera2ld: Since the array is always turned into a string on first search, why do we keep the array at first?

src/options/views/tab-installed.vue
```diff
@@ -378,6 +380,39 @@ export default {
...
+          return searchRE ? searchRE.test(item) : item.includes(lowerSearch);
+        });
+      });
+    },
+    async scheduleSearch() {
+      if (!scriptCodeRetrieved) {
+        scriptCodeRetrieved = true;
+        const { scripts } = this.store;
+        const ids = scripts.map(script => `code:${script.props.id}`);
+        const data = await browser.storage.local.get(ids);
```
gera2ld: I think `browser.storage.local` should not be used directly outside background, because this requires the options page to know about the data structure defined in background, maybe it's better to load from background with a command.

tophf: I thought about that too, but I **really** want to avoid the triple processing of a huge chunk of data: local.get() parses into JS object, messaging stringifies it, the receiver parses it again.

What if I put those storage prefixes in a separate file like `storage-consts.js`?

src/options/views/tab-installed.vue
tophf: To avoid the unnecessary lowercasing of all the strings during initialization, which takes time and stresses garbage collector. It's performed now on demand - only when the user actually searches for something. And since it's enough for a first match to succeed, the unused values won't be lowercased unnecessarily.

gera2ld: > And since it's enough for a first match to succeed, the unused values won't be lowercased unnecessarily.

Then we can build `$cache.search` on first search, even without the need to build an unnecessary array.

src/options/views/tab-installed.vue
gera2ld: In my opinion, the options page should not know about the implementation details of storage. This makes the application more portable and easier to port to other platforms.

And I don't think it's a big deal to transport some large strings during a search by user, especially when it's already used in the script editor and installation confirming.

We can even pass the search text to background and return the filtered items asynchronously if you really care about the large messages.

tophf: >more portable
It shouldn't be an issue because a compatibility polyfill for browser.storage would process the key names regardless of where the request is performed.

>And I don't think it's a big deal to transport some large strings during a search by user, especially when it's already used in the script editor and installation confirming.

Why? I have 500kB in my 100 scripts. My scripts are rather small, but some people may have much bigger scripts. The script editor/confirmer only uses one script, not all of them.

>We can even pass the search text to background
That would be asynchronous so it'll add the unnecessary complexity to the options page. And the background page will have to keep those things in a cache => more complexity.

src/options/views/tab-installed.vue
tophf: The wrapper array (or another method of postponing the unnecessary lowercasing) will still be required. Of course I can move the code that creates $cache.search into `performSearch` but I don't think it's a better way.

tophf: To illustrate what happens here's the initial `search`:
```
['Foo']
['Bar']
```
* When the user types `f` the first item will be unwrapped and lowercased to `'foo'` string - and stop because `f` matches `foo`.
* When the user continues to type `fu` the first item will be checked unsuccessfully and the second item will be unwrapped and lowercased.
* The process will continue until the last item is processed, which is the script code.

src/options/views/tab-installed.vue
gera2ld: > That would be asynchronous so it'll add the unnecessary complexity to the options page. And the background page will have to keep those things in a cache => more complexity.

The scripts are already kept in memory because they are possibly needed whenever a page is loaded. So I think the *complexity* is actually just a `filter` in the background.

> It shouldn't be an issue because a compatibility polyfill for browser.storage would process the key names regardless of where the request is performed.

It is an issue because `browser.storage` or its alternative is not always accessible from a place other than background. For example, Maxthon again.

Maybe it's a good idea to have a separate `common/scripts.js` to load and dump scripts. We can check environment in it and fetch scripts in a proper way.

ps. Maybe we should drop support for Maxthon in the future and just follow the WebExtension standard. And there are some more, Edge, Safari, etc.

tophf: > The scripts are already kept in memory

Only the script info object AFAICT, not the `code`.

> browser.storage or its alternative is not always accessible 

No problem: the polyfill should simply imitate browser.storage.

> separate common/scripts.js to load and dump scripts.

Maybe, but the only part that the options page needs is the code, not the rest of the info which is cached and processed inside the background page.

src/options/views/tab-installed.vue
gera2ld: I can understand it now. But is it worth adding such complexity just to avoid joining and lowercasing a few, to be exact, 6 at most, one-line strings? And there may likely be empty strings, wasting time to be wrapped and unwrapped.

src/options/views/tab-installed.vue
gera2ld: I prefer using an API from options page, rather than the level `browser.storage`, even with prefix constants. How about moving some of the methods in `src/background/utils/db` to common?

src/options/views/tab-installed.vue
tophf: Not for these 6 strings specifically because the difference in performance is almost non-existent, but to support the big `code` added afterwards. Also, wrapping into an array is negligible compared to lowercasing especially if VM will switch to `toLocaleLowerCase` to support unicode fully, which is very slow. As for "complexity" there are only a few statements/calls added which doesn't really seem complex to me.

tophf: ...but I'll remove the modifications I've made in filter/map, as there are no noticeable gains.

src/options/views/tab-installed.vue
tophf: I'll try, okay.

src/common/storage.js
```diff
@@ -0,0 +1,118 @@
+import { request, buffer2string, ensureArray } from '#/common';
```
tophf: or maybe just `'.'`?

gera2ld: `'.'` is ok if it does not cause circular dependency. Maybe better to move the methods to a file other than `index.js`. 😅

PR #703, opened 2019-11-11
feat: open tabs from popup using tabOpen()

tophf:
Tabs opened from popup will use `tabOpen()` so they are opened next to the source and have openerTabId (on closing of such a tab the source tab will be focused).

Collateral refactor: deduplicate getActiveTab() and makePause()

### DIFF
--- a/src/common/index.js
+++ b/src/common/index.js
@@ -58,7 +58,6 @@ export function sendMessage(payload, { retry } = {}) {
  * or when configured to restore the session, https://crbug.com/314686
  */
 export async function sendMessageRetry(payload, retries = 10) {
-  const makePause = ms => new Promise(resolve => setTimeout(resolve, ms));
   let pauseDuration = 10;
   for (; retries > 0; retries -= 1) {
     const data = await sendMessage(payload).catch(noop);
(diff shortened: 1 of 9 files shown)


src/common/index.js
```diff
@@ -195,3 +194,17 @@ export function encodeFilename(name) {
 export function decodeFilename(filename) {
   return filename.replace(/-([0-9a-f]{2})/g, (_m, g) => String.fromCharCode(parseInt(g, 16)));
 }
+
+export async function getActiveTab() {
+  // not using destructuring as it adds 0.5kB of babel's _iterableToArrayLimit to injected*
```
gera2ld: I think we should avoid such micro-optimizing and let the tools handle it.

tophf: We should, in theory, but in reality the tools don't handle it, and I find it conceptually wrong that 99.99% of users who run Violentmonkey on modern browsers are forced to run the totally unnecessary polyfill for something that can be easily avoided.

tophf: Ah, you mean we can switch to `Firefox>=53` in the other PR? That would be great! Let's handle that first then.

PR #705, opened 2019-11-12
feat: shift-del cuts entire line in FF

tophf:
Firefox-only.

Adds the convenient and pretty standard <kbd>Shift</kbd><kbd>Delete</kbd> to cut the entire line when nothing is selected.
Weirdly, FF implements this behavior in <kbd>Ctrl</kbd><kbd>X</kbd> and <kbd>Ctrl</kbd><kbd>Insert</kbd>.

### DIFF
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -68,6 +68,7 @@ import Tooltip from 'vueleton/lib/tooltip/bundle';
 import { debounce } from '#/common';
 import ToggleButton from '#/common/ui/toggle-button';
 import options from '#/common/options';
+import { isFirefox } from '#/common/ua';
 
 /* eslint-disable no-control-regex */
 const MAX_LINE_LENGTH = 50 * 1024;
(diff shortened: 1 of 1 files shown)


gera2ld:
Why is this implemented for Firefox? For me <kbd>Shift</kbd><kbd>Delete</kbd> doesn't work either in Chrome or Firefox, and I've never used such a shortcut before.
Isn't it normal to cut a line with <kbd>Ctrl</kbd><kbd>X</kbd> when nothing is selected? This is the way many IDEs or editors works AFAICT.

tophf:
Apparently it's Windows only. It's a well-known standard Windows shortcut that was there for decades since the very first versions as well as Shift-Insert and Ctrl-Insert. I'll change the PR to Windows-only then.

tophf:
It's also used in Linux GNOME/KDE: https://en.wikipedia.org/wiki/Cut,_copy,_and_paste#Common_keyboard_shortcuts

gera2ld:
OK, it's not bad to have a standard shortcut, though I think <kbd>Ctrl</kbd><kbd>X</kbd> is much more convenient to press.

tophf:
Oops, I almost forgot that it should work on multiple lines in multi-caret mode (when lines are selected via Ctrl-click).

tophf:
>though I think CtrlX is much more convenient to press

A matter of habit. Pressing Shift-Del using the right hand which is resting around the arrow keys is super convenient as well.

tophf:
I've [suggested this feature in CodeMirror repo](https://github.com/codemirror/CodeMirror/issues/6061). Let's wait for their response.

tophf:
CM implemented a brilliant workaround so everyone will benefit with its next release.

Issue #706, opened 2019-11-14
[Feature request] GM.xmlHttpRequest argument "binary"


machsix:
### What is the problem?
Compatibility of `GM.xmlHttpRequest`
### How to reproduce it?
No need

### What is the expected result?

Add support of argument `binary`. Both Greasemonkey https://wiki.greasespot.net/GM.xmlHttpRequest and Tampermonkey https://www.tampermonkey.net/documentation.php#GM_xmlhttpRequest only accept `String` as `data` but `blob` as `binary`. 

Although I only use violentmonkey, I couldn't stop users from using tampermonkey, aka 💩 . It seems we only need to add a few lines in `https://github.com/violentmonkey/violentmonkey/blob/master/src/injected/web/requests.js`

### Environment
- Browser: any
- Browser version:
- Violentmonkey version:
- OS:

PR #707, opened 2019-11-14
fea: allow pass {binary: blob}

machsix:
close https://github.com/violentmonkey/violentmonkey/issues/706

I'm new to extension development. Please forgive me if something is wrong


src/injected/web/requests.js
```diff
@@ -113,7 +113,7 @@ function start(req, id) {
       log('warn', null, `Unknown responseType "${responseType}", see https://violentmonkey.github.io/api/gm/#gm_xmlhttprequest for more detail.`);
     }
   }
-  encodeBody(details.data)
+  encodeBody(details.data || details.binary)
```
tophf: This doesn't compute at all. The `data` is an object but `binary` is a boolean. I'll make a proper PR.

PR #709, opened 2019-11-15
fix window.close in Firefox

gera2ld:
Fix `window.close` in `content` mode in Firefox.

Previously `this.window` returns the real `window` instead of the wrapper due to an improper check. As a result, `@grant window.close` does not work.


tophf:
LGTM!

PR #712, opened 2019-11-16
Implement Firefox-only injection-mode `wrap`

### LINKED ISSUE #702: [Suggestion][Firefox] Assign window.wrappedJSObject to unsafeWindow when using injection into the content context
This will allow userscripts to access the global scope of page.

gera2ld:
Implement Firefox-only injection-mode `wrap`.

close #702 .

### Changes
- Add injection mode `wrap` in Firefox. In `wrap` mode `unsafeWindow` will be `window` from the page context even if the script is running in content page.
- Change `@inject-into` into a whitespace separated mode list, which equals to `auto` or consists of `page`, `content`, `wrap`. VM will try to find the first available injection mode from the list. The following definitions are identical:
  ```js
  // @inject-into auto
  // @inject-into page wrap content
  ```
- Change settings of default injection mode from a select box into an input box, so that users can define the desired mode list more easily. Since an input box seems confusing to normal users, I moved it to advanced settings.

### DIFF
--- a/src/injected/content/inject.js
+++ b/src/injected/content/inject.js
@@ -1,14 +1,21 @@
-import { isFirefox } from '#/common/ua';
-import { getUniqId, sendCmd } from '#/common';
 import {
-  INJECT_PAGE, INJECT_CONTENT, INJECT_AUTO,
+  INJECT_AUTO,
+  INJECT_CONTENT,
+  INJECT_INTERNAL_CONTENT,
+  INJECT_INTERNAL_PAGE,
+  INJECT_INTERNAL_WRAP,
+  INJECT_MAPPING,
+  INJECT_PAGE,
   browser,
 } from '#/common/consts';
+import { getUniqId, sendCmd } from '#/common';
+import { isFirefox } from '#/common/ua';
+
 import { attachFunction } from '../utils';
-import bridge from './bridge';
 import {
   forEach, join, jsonDump, setJsonDump, append, createElementNS, NS_HTML,
 } from '../utils/helpers';
+import bridge from './bridge';
 
 // Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1408996
 const VMInitInjection = window[process.env.INIT_FUNC_NAME];
@@ -25,9 +32,12 @@ bridge.addHandlers({
 
 export function triageScripts(data) {
   const scriptLists = {
-    [INJECT_PAGE]: [],
-    [INJECT_CONTENT]: [],
+    [INJECT_INTERNAL_PAGE]: [],
+    [INJECT_INTERNAL_CONTENT]: [],
   };
+  // `INJECT_INTERNAL_WRAP` is a special case of `INJECT_INTERNAL_CONTENT`
+  // so we will just reuse the list to keep the execution orders
+  scriptLists[INJECT_INTERNAL_WRAP] = scriptLists[INJECT_INTERNAL_CONTENT];
   if (data.scripts) {
     data.scripts = data.scripts.filter(({ meta, props, config }) => {
       if (!meta.noframes || window.top === window) {
@@ -40,14 +50,19 @@ export function triageScripts(data) {
       return false;
     });
     let support;
-    data.scripts.forEach((script) => {
-      let injectInto = script.custom.injectInto || script.meta.injectInto || data.injectInto;
-      if (injectInto === INJECT_AUTO) {
+    const injectChecking = {
+      [INJECT_INTERNAL_PAGE]: () => {
         if (!support) support = { injectable: checkInjectable() };
-        injectInto = support.injectable ? INJECT_PAGE : INJECT_CONTENT;
-      }
-      const list = scriptLists[injectInto];
-      if (list) list.push(script);
+        return support.injectable;
+      },
+      [INJECT_INTERNAL_WRAP]: () => isFirefox,
+      [INJECT_INTERNAL_CONTENT]: () => true,
+    };
+    data.scripts.forEach((script) => {
+      const injectInto = script.custom.injectInto || script.meta.injectInto || data.injectInto;
+      const internalInjectInto = INJECT_MAPPING[injectInto] || INJECT_MAPPING[INJECT_AUTO];
+      const availableInjectInto = internalInjectInto.find(key => injectChecking[key]?.());
+      scriptLists[availableInjectInto]?.push({ script, injectInto: availableInjectInto });
     });
   }
   return scriptLists;
@@ -72,7 +87,7 @@ export function injectScripts(contentId, webId, data, scriptLists) {
   if (injectContent.length) {
     const invokeGuest = VMInitInjection()(...args, bridge.onHandle);
     const postViaBridge = bridge.post;
-    bridge.invokableIds.push(...injectContent.map(script => script.props.id));
+    bridge.invokableIds.push(...injectContent.map(({ script }) => script.props.id));
     bridge.post = msg => (
       msg.realm === INJECT_CONTENT
         ? invokeGuest(msg)
@@ -83,7 +98,7 @@ export function injectScripts(contentId, webId, data, scriptLists) {
       data: {
         ...data,
         mode: INJECT_CONTENT,
-        scripts: injectContent,
+        items: injectContent,
       },
       realm: INJECT_CONTENT,
     });
@@ -96,7 +111,7 @@ export function injectScripts(contentId, webId, data, scriptLists) {
       data: {
         ...data,
         mode: INJECT_PAGE,
-        scripts: injectPage,
+        items: injectPage,
       },
     });
   }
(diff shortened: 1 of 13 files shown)


tophf:
I thought it'd be enough to just add a third item `wrap` equal to `content+wrap` internally without the ability to explicitly list modes so the selector in settings would be preserved. And make `auto` do `page wrap content` internally of course. Is there a use case I'm missing?

gera2ld:
Ah, that's true, I was exposing too much implementation details.

tophf:
Oops I've missed the push. Apparently github doesn't notify on forced push. I'll check this soon!

src/common/ui/setting-input.vue
```diff
@@ -0,0 +1,54 @@
+<template>
```
tophf: `setting-input` is unused now so it's probably better to save this file locally until it's needed in the future?

src/injected/content/inject.js
```diff
@@ -40,14 +43,29 @@ export function triageScripts(data) {
...
+      // `auto` tries to provide `window` from the real page as `unsafeWindow`
+      [INJECT_AUTO]: [INJECT_PAGE, INJECT_WRAP, INJECT_CONTENT],
+    };
+    data.scripts.forEach((script) => {
+      const injectInto = script.custom.injectInto || script.meta.injectInto || data.injectInto;
+      const internalInjectInto = internalModes[injectInto] || [injectInto];
+      const availableInjectInto = internalInjectInto.find(key => {
+        const check = injectChecking[key];
+        return check && check();
+      });
```
tophf: `key => injectChecking[key]?.()` optional chaining rocks!

src/injected/content/inject.js
```diff
@@ -40,14 +43,29 @@ export function triageScripts(data) {
...
+      const list = scriptLists[availableInjectInto];
+      if (list) list.push({ script, injectInto: availableInjectInto });
```
tophf: `list?.push({ script, injectInto: availableInjectInto });`

And to fix the [false-positive error from ESLint](https://github.com/eslint/eslint/issues/11045#issuecomment-436685184) the following should be added to `.eslintrc`:

```js
  rules: {
    'prefer-object-spread': 'off',
    // workaround for the false-positive on optional?.chainingCall() statement
    // https://github.com/eslint/eslint/issues/11045#issuecomment-436685184
    'no-unused-expressions': 'off',
    'babel/no-unused-expressions': 'error',
  },
```

src/injected/web/gm-wrapper.js
```diff
@@ -1,4 +1,6 @@
-import { INJECT_PAGE, INJECT_CONTENT, METABLOCK_RE } from '#/common/consts';
+import {
+  INJECT_PAGE, INJECT_CONTENT, INJECT_WRAP, METABLOCK_RE,
```
tophf: I always wanted to ask, why `3` items is the limit for a single line object literal? I would expect the limit to be higher or even based on the line length.

src/options/views/tab-settings/index.vue
```diff
@@ -70,6 +60,21 @@
...
+      <section>
+        <h3 v-text="i18n('labelGeneral')"></h3>
+        <div class="mb-1">
+          <label>
+            <span v-text="i18n('labelInjectionMode')"></span>
+            <select v-model="defaultInjectInto">
+              <option value="auto">auto</option>
+              <option value="page">page</option>
+              <option value="content">content</option>
+              <option value="wrap">wrap</option>
```
tophf: Maybe hide `wrap` in Chrome?

src/options/views/tab-settings/index.vue
```diff
@@ -70,6 +60,21 @@
...
+            </select>
+            <a class="ml-1" href="https://violentmonkey.github.io/posts/inject-into-context/" target="_blank" rel="noopener noreferrer" v-text="i18n('learnInjectionMode')"></a>
```
tophf: Now a short explanation hint near the selector is really needed because `wrap` is cryptic, and there is no substantially better name.

For example:

> `page` uses script elements in the page, `wrap` (Firefox only) same as `content` but unsafeWindow is mapped to the window object of the page,  `content` runs in the content script context to avoid CSP restrictions,  `auto` automatically uses one of these modes that succeeds in the listed order.

src/options/views/tab-settings/index.vue
```diff
@@ -142,6 +152,12 @@ export default {
         if (value !== oldValue) options.set(key || name, normalize(value));
       };
     },
+    validateInjectionMode(value) {
+      return /^\s*(?:auto|(?:(^|\s+)(?:page|content|wrap))+)\s*$/.test(value);
+    },
+    normalizeInjectionMode(value) {
+      return value.split(/\s+/).filter(Boolean).join(' ');
+    },
```
tophf: Now that `<select>` is used again, these methods shouldn't be needed, AFAICT.

src/options/views/tab-settings/index.vue
tophf: On the other hand, it may be confusing to hide, if the user transfers the backup between FF/Chrome...

src/injected/web/gm-wrapper.js
gera2ld: Just default options... I guess one line per item is expected by the lint rule.

src/options/views/tab-settings/index.vue
gera2ld: In most cases `auto` should be enough, I think we could even remove this option so that users do not have to learn a new concept.

src/common/ui/setting-input.vue
gera2ld: Sure

src/injected/web/gm-wrapper.js
tophf: Well, if it doesn't irritate you I'll just try to get used to it.

src/options/views/tab-settings/index.vue
tophf: Hmm, what if `content` will behave internally like `content+wrap` by default and the new mode will be not `wrap` but `content (legacy)` which is internally the old `content`?

gera2ld: > what if `content` will behave internally like `content+wrap` by default and the new mode will be not `wrap` but `content (legacy)` which is internally the old `content`?

Looks good, and simple.

gera2ld: There is a problem, I think the keywords should be consistent either using in the settings or in `@inject-into`. But I don't think we can use `@inject-into content (legacy)`, so maybe something as a word, like `content-legacy`.

tophf: Another problem with renaming is that it will make `content` behave differently between Chrome/FF.

These ideas would be provide a more consistent behavior:

* hiding `wrap` in Chrome
* display it as `wrap (only in Firefox)`
* add `disabled` attribute on the `<option>` element so it's unselectable

And of course a hint below the selector would be helpful.

gera2ld: I'm always trying to hide the differences between browsers and provide same or similar functionality to developers and users. So,
- since `wrap` is a Firefox-only feature, I don't want to expose it to users, and not to show a different UI in Firefox.
- `content` mode is supposed to be the final way to inject scripts, just trying to get scripts executed, without ensuring same environments as in page contexts.

But I'm starting to think that, `wrap` might be better as a fallback for `page`, i.e. `page=page+wrap, content=content, auto=page+wrap+content`. It seems more consistent for both Chrome and Firefox this way.

tophf: Yeah, that makes sense.

src/options/views/tab-settings/index.vue
gera2ld: Since there is no new concept here, maybe we can move the explanation to the linked documentation.

tophf: Ideally I'd like to put all that info inside the extension and make it translatable :-)

tophf: Could be something to do in the future though.

tophf:
Well it works fine in my cursory tests but I'd really wish you could distribute a beta extension in the web store. If not that, at least an auto-generated zip from the master which can be installed as an unpacked extension by the testers who aren't keen on installing the entire node.js shebang.

src/injected/content/inject.js
```diff
@@ -40,14 +50,19 @@ export function triageScripts(data) {
       return false;
     });
     let support;
-    data.scripts.forEach((script) => {
-      let injectInto = script.custom.injectInto || script.meta.injectInto || data.injectInto;
-      if (injectInto === INJECT_AUTO) {
+    const injectChecking = {
+      [INJECT_INTERNAL_PAGE]: () => {
```
tophf: This should not be needed in Chrome because script elements created by extensions are explicitly allowed regardless of page CSP.

gera2ld: It's true but I think it's safer to check it this way than to check browser type with user-agent or some special properties. This will keep working even someday Chrome changes its behavior.

PR #722, opened 2019-11-19
fix: drag'n'drop on touch devices

### LINKED ISSUE #720: Can't reorder on Firefox for Android (and other touch device)
### What is the problem?
Can not reorder on touch device

### How to reproduce it?
1. Install Firefox for Android
2. Install some scripts
3. Long press/swipe/press and drag/whatever

Show button to reorder/Drag and drop to reorder

Nothing happen

### Environment
- Browser: Firefox for Android
- Browser version: 68.2.1
- Violentmonkey version: 2.12.1
- OS: Android 9

tophf:
Fixes #720.

Collateral changes:

+ highlight the dragged element's background, otherwise it's kinda easy to confuse which one is under a finger
+ various speedups noticeable on slow mobile devices

### DIFF
--- a/src/options/utils/dragging.js
+++ b/src/options/utils/dragging.js
@@ -1,0 +1,194 @@
+import { isFirefox } from '#/common/ua';
+
+const SCROLL_GAP = 50;
+// px per one 60fps frame so the max is ~1200px per second (~one page)
+const MAX_SCROLL_SPEED = 20;
+const ONE_FRAME_MS = 16;
+// touch-and-hold duration in ms before recognizing dragstart (needed to allow fling-scrolling)
+const LONGPRESS_DELAY = 500;
+const DROP_EVENT_RELAY = 'VM-drop-event-relay';
+
+const isTouch = 'ontouchstart' in document;
+const eventNames = isTouch
+  ? { start: 'touchstart', move: 'touchmove', end: 'touchend' }
+  : { start: 'dragstart', move: 'mousemove', end: 'mouseup' };
+// inserting a scroll blocker as the first element of the list to avoid changing the list's
+// style/class as it would be slooooooooooow due to style recalc on all of its children
+const noScroll = isTouch && Object.assign(document.createElement('div'), {
+  className: 'dragging-noscroll',
+});
+const eventsToSuppress = ['scroll', 'mouseenter', 'mouseleave'];
+const { addEventListener: on, removeEventListener: off } = EventTarget.prototype;
+
+let dragged;
+let elements;
+let height;
+let index;
+let lastIndex;
+let longPressEvent;
+let longPressTimer;
+let offsetX;
+let offsetY;
+let original;
+let parent;
+let scrollEdgeTop;
+let scrollEdgeBottom;
+let scrollTimer;
+let scrollSpeed;
+let scrollTimestamp;
+let xyCache;
+
+/**
+ * @param {Element} el
+ * @param {function(from,to)} onDrop
+ */
+export default function enableDragging(el, { onDrop }) {
+  if (!parent) {
+    parent = el.parentElement;
+    // pre-FF64 doesn't support `@media (pointer: coarse)`
+    if (isTouch && !matchMedia('(pointer: coarse)').matches) {
+      parent.classList.add('touch');
+    }
+  }
+  el::on(eventNames.start, isTouch ? onTouchStart : onDragStart);
+  el::on(DROP_EVENT_RELAY, () => onDrop(index, lastIndex));
+}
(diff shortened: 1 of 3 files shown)


tophf:
I only tested in mobile Yandex which is Chromium-based.

tophf:
There's a bug I need to fix: dragging should not activate when scrolling the list. I'll add a requirement for the initial touch to be touch-and-hold. As for the hold duration, I'll experiment with different values and will try to find if there's a standard value.

tophf:
Turns out touch-and-hold would require imitating the native fling-scrolling which is both non-trivial and is just conceptually wrong. Instead a two-finger dragging is implemented. I guess I'll also add a dynamic tooltip shown on a longpress on a script element so that the feature is easier to discover.

gera2ld:
I tested in Firefox Android but it does not work well. During dragging operation, the window scrolls in y axis. In other words, dragging works only horizontally, but not vertically. So I can hardly move an item higher or lower unless the window cannot scroll any more.

tophf:
Yeah I'm still fighting this, didn't upload the new code yet.

gera2ld:
I think d'n'd in mobile is kind of difficult both for developer and user.

How about a different way in mobile like this:

- click on a button or something to start moving, fade out the picked script item
- scroll the window to find the new place to put the script
- click on the new place, and finish moving

The *new place* here may be to other scripts. e.g. clicking on a script means to put the picked script after it. We could add a placeholder at the beginning of the list to allow moving a script to the start.

tophf:
I almost finished, and I generally like how DND works, it's easy and fast and uses the standard touch paradigm of the longpress, but I had to remove all animations to make sure the base logic works. I'm currently in the process of trying to restore it. Although personally I'm not a fan of programmatic animations.

gera2ld:
It will be great if d'n'd works well in mobile as in desktop. 👍

tophf:
I've pushed the WIP version that works in mobile Chromium and desktop emulation via devtools.

Still buggy in FF under MEmu emulator, the viewport scrolls on dragging.
Couldn't test on a real mobile device as I'm using an incompatible theming software that crashes FF.

tophf:
I've restored the animation and added a workaround for the bug in FF mobile. The workaround isn't perfect per my tests in an emulator (not on a real device), the container still scrolls at start and is immediately restored to the original position, I don't think I can do better - it looks like FF intentionally runs the first `touchmove` without blocking the main thread (Chrome does this only for global touch listeners on document/window).

tophf:
Now everything seems to be working correctly. It's also faster.

src/options/utils/dragging.js
```diff
@@ -0,0 +1,194 @@
...
+  parent.appendChild(dragged);
+  if (isTouch) parent.insertAdjacentElement('afterBegin', noScroll);
+  document::on(eventNames.move, onDragMouseMove);
+  document::on(eventNames.end, onDragMouseUp);
+}
+
+function onDragMouseMove(e) {
+  const { clientX: x, clientY: y, target } = e.touches?.[0] || e;
+  let moved;
+  const hovered = isTouch ? scriptFromPoint(x, y) : target.closest('.script');
```
gera2ld: In Firefox, `target` could be `document` when dragging to the top of the page, and has no `.closest` method.

src/options/utils/dragging.js
```diff
@@ -0,0 +1,194 @@
...
+  // FF bug: despite placeholder having `pointer-events:none` it's still reported in `target`
+  if (hovered && hovered !== original) {
+    const rect = hovered.getBoundingClientRect();
+    const isDown = y > rect.top + rect.height / 2;
+    moved = original !== hovered[`${isDown ? 'next' : 'previous'}ElementSibling`];
+    if (moved) {
+      hovered.insertAdjacentElement(isDown ? 'afterEnd' : 'beforeBegin', original);
```
gera2ld: The original element does not need to be moved during dragging.

src/options/utils/dragging.js
tophf: Yep.

src/options/utils/dragging.js
tophf: I thought I preserved the old internal behavior. Anyway it seems to work correctly in the sense that I couldn't break it :smile:

gera2ld:
It works well in Firefox Android now, but does not work in Kiwi Browser. Anyway, I think it's good enough with DND.

tophf:
Works barely in Kiwi for me with inexplicable jank and random delays.
Works fine in Chromium-based Yandex browser.

PR #728, opened 2019-11-24
Custom update interval

### LINKED ISSUE #577: Option to set a global script update interval
Checking for updates everyday is a bit excessive. So it would be a good idea to give the user a option to set it to check less frequently, like every week, month etc.

tophf:
Fixes #577

image

### DIFF
--- a/src/_locales/cs/messages.yml
+++ b/src/_locales/cs/messages.yml
@@ -205,8 +205,8 @@ labelAutoReloadCurrentTab:
   description: Option to reload current tab after a script is switched on or off from menu.
   message: Znovu načíst současný list po zapnutí či vypnutí skriptu z nabídky.
 labelAutoUpdate:
-  description: Option to allow automatically checking scripts for updates every 24 hours.
-  message: Automaticky kontrolovat aktualizace skriptu každý den
+  description: Option to allow automatically checking scripts for updates.
+  message: ''
 labelBadge:
   description: Label for option to show number on badge.
   message: ''
--- a/src/_locales/de/messages.yml
+++ b/src/_locales/de/messages.yml
@@ -207,8 +207,8 @@ labelAutoReloadCurrentTab:
   description: Option to reload current tab after a script is switched on or off from menu.
   message: Nach dem Ein-/Ausschalten eines Skripts den geöffneten Tab neu laden
 labelAutoUpdate:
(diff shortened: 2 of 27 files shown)


tophf:
I wonder if VM should allow customizing a per-script update interval in the editor's settings? Theoretically it makes sense e.g. in case of a long global interval (7+ days) and a small interval for an important script or two. On the other hand, this would be a rare feature for like 3½ advanced users, because the majority of the users would be just fine with the default 1 day interval.

gera2ld:
Actually I am always curious why 1 day interval is considered to be excessive by some users. IMO a few requests a day should be negligible to a browser.
Something concerns me in this PR is, what if some other users want more frequent update checking, like 6 hours.

> I wonder if VM should allow customizing a per-script update interval in the editor's settings?

Sure. We can do this in the future, replacing the existed checkbox in script settings.

tophf:
>what if some other users want more frequent update checking, like 6 hours.

greasyfork's owner was complaining once about a surge of update requests that were caused (IIRC) by a bug in Tampermonkey or a common script library that performed the check too frequently. I remember the resolution was to issue the request once per day. In Tampermonkey's options one day is the minimum unit as well.

That said, we can allow more frequent updates on individual scripts in the future - but only if those aren't loaded from the known gallery sites like greasyfork. Obviously, `localhost` or an intranet site can be checked each minute if anyone wants that.

tophf:
Huh, I see TM offers 6 and 12 hours. Well, I was wrong about that, but I think it'd be better to limit the global one to 1 day and allow a per-script interval in the future.

gera2ld:
There is a problem, if we set `autoUpdate` to interval in days, it's hard to offer lower interval like in hours or minutes with the same key.

tophf:
The global interval will always be in days because it seems a safe minimum. The per-script intervals can be set in minutes. These two are different options so there should be no problem.

tophf:
Importing an old backup with `autoUpdate:true` will clear the number input because Vue reads the option directly (not using `normalize`). Reloading the page will correctly convert it into `1`. What is the correct approach to solve this?

tophf:
Actually, it's technically possible to use fractions in the number of days so e.g. `0.25` would be 6 hours. The change is trivial: set `step=0.25` in html and make `normalize()` round the value to the nearest .25 number. That's of course if we ever decide to support it.

PR #734, opened 2019-11-26
fix: remove `wrap` mode in Firefox

### LINKED ISSUE #733: [regression] v2.12.2 breaks activity page on gitlab
### What is the problem?
After updating to [v2.12.2](https://github.com/violentmonkey/violentmonkey/releases/tag/v2.12.2) opening any activity page on gitlab example: https://gitlab.com/Mottie/GitLab-userscripts/activity shows broken page.

This happens when https://greasyfork.org/en/scripts/29352-screen-resolution-spoof userscript is active. It didn't happened before.
### How to reproduce it?
1. Install https://greasyfork.org/en/scripts/29352-screen-resolution-spoof
2. Open https://gitlab.com/Mottie/GitLab-userscripts/activity
Page is loaded correctly
Page loading is broken (activity is blank)

```
Content Security Policy: Ignoring “'unsafe-inline'” within script-src or style-src: nonce-source or hash-source specified
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”). injected.js:1:7684
Content Security Policy: Ignoring “'unsafe-inline'” within script-src or style-src: nonce-source or hash-source specified 2
Error: Permission denied to access object main.0d6be92c.chunk.js:111:376368
Source map error: Error: request failed with status 404
Resource URL: https://assets.gitlab-static.net/assets/webpack/main.0d6be92c.chunk.js
Source Map URL: main.0d6be92c.chunk.js.map
```
### Environment
- Browser: FF
- Browser version: 70.0.1
- Violentmonkey version:2.12.2
- OS: Linux

gera2ld:
In `wrap` mode, `unsafeWindow` is set to `wrappedJSObject`.
Consequently, we get "Permission denied to access object" error
when trying to access objects in `wrappedJSObject` from `content` script.

This is by design in Firefox and seems unable to be fixed by us, so we
remove `wrap` mode here.

close #733

### DIFF
--- a/src/common/consts.js
+++ b/src/common/consts.js
@@ -4,13 +4,12 @@ export const INJECT_CONTENT = 'content';
 
 export const INJECT_INTERNAL_PAGE = 'page';
 export const INJECT_INTERNAL_CONTENT = 'content';
-export const INJECT_INTERNAL_WRAP = 'wrap';
 
 export const INJECT_MAPPING = {
   // `auto` tries to provide `window` from the real page as `unsafeWindow`
-  [INJECT_AUTO]: [INJECT_INTERNAL_PAGE, INJECT_INTERNAL_WRAP, INJECT_INTERNAL_CONTENT],
-  // inject into page context, if failed, try `wrap` mode for Firefox
-  [INJECT_PAGE]: [INJECT_INTERNAL_PAGE, INJECT_INTERNAL_WRAP],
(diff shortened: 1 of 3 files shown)


tophf:
I wonder if there's a proper way to define properties on wrappedJSObject - like maybe use the special Firefox functions cloneInto and exportInto. Maybe we could apply them automatically (for example by setting up a `Proxy` on wrappedJSObject), but that's something for the future.

PR #742, opened 2019-11-30
feat: automatic dark theme in supported browsers/OS

tophf:
A simple built-in dark theme that activates on [prefers-color-scheme: dark](https://caniuse.com/#search=prefers-color-scheme) automatically in new browsers when the OS and the browser are using the dark mode.

Notes:
* almost all colors are mapped to a simple palette of 16 colors ranging from `#000` to `#fff` in the default theme and from `#222222` to `#b8b8b8` in the dark theme
* a few colors were approximated with the closest one from the palette
* a disabled `<button>` was simplified by using opacity instead of three separate colors

image

P.S.

Ideally I'd like to extract all media queries in a separate dark.css and load it conditionally:
```html
<link href="dark.css" rel="stylesheet" media="(prefers-color-scheme: dark)"/>
```
so that I could add a user option to force the dark theme in older browsers. There are some plugins to automate this so I'll probably try them in the future.

### DIFF
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -458,4 +458,126 @@ input[type=text].is-error {
   border: 1px solid #e85600;
   background: #e8560010;
 }
+
+@media (prefers-color-scheme: dark) {
+  // mostly copied from Monokai theme
+  .cm-s-eclipse {
+    &.CodeMirror {
+      color: var(--fg);
+      background: var(--bg);
+    }
+    .CodeMirror {
+      &-scrollbar-filler,
+      &-gutter-filler {
+        background: none;
+      }
+      &-gutters {
+        border-color: var(--fill-2);
+        background-color: var(--fill-0-5);
+      }
+      &-selected {
+        background: #49483e;
+      }
+      &-line::selection,
+      &-line::-moz-selection,
+      &-line > span::selection,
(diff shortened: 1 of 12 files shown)


tophf:
I've tried [media-query-plugin](https://www.npmjs.com/package/media-query-plugin) and it extracted 4 dark.css files (one per each page entry). It's not what I want because the documentation of MQL plugin says the assets should automatically go to `htmlWebpackPlugin.files.extracted` but apparently there's some conflict with mini-css-extract-plugin, which I don't know how to solve as I have no idea what happens in all that webpack-related voodoo.

Here's what I've used in webpack.conf.js instead of `modify()` on line 66:

```js
  modify(null, (config) => {
    const MediaQueryPlugin = require('media-query-plugin');
    config.plugins.push(
      new MediaQueryPlugin({
        include: true,
        queries: {
          '(prefers-color-scheme: dark)': 'dark',
        },
      }),
    );
    config.module.rules.forEach(({ use }) => {
      const i = Array.isArray(use) && !use.includes(MediaQueryPlugin.loader)
        ? use.findIndex(u => (u.loader || u) === 'css-loader')
        : -1;
      if (i >= 0) use.splice(i + 1, 0, MediaQueryPlugin.loader);
    });
  }),
```

PR #745, opened 2019-12-04
UI tweaks

tophf:
* autocomplete using entered values in filter/search/replace inputs, the entered value is remembered when <kbd>Enter</kbd> is pressed in a focused input element
  image
* autofocus values editor input element (`key` for a new value or `value`)
* close value editor on Esc when pressed inside the input/textarea element
* disable spellcheck in the key/value input elements

### DIFF
--- a/src/options/views/tab-installed.vue
+++ b/src/options/views/tab-installed.vue
@@ -62,11 +62,12 @@
             </label>
           </div>
         </dropdown>
-        <div class="filter-search hidden-sm">
+        <!-- form and id are required for the built-in autocomplete using entered values -->
+        <form class="filter-search hidden-sm" @submit.prevent>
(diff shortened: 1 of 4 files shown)


src/options/views/tab-installed.vue
```diff
@@ -63,8 +63,11 @@
           </div>
         </dropdown>
         <div class="filter-search hidden-sm">
-          <input type="search" :placeholder="i18n('labelSearchScript')" :title="searchError"
-                 v-model="search">
+          <!-- form and id is required for the built-in autocomplete using entered values -->
+          <form @submit.prevent>
```
gera2ld: How about changing the outer `<div>` to `<form>` so that the DOM structure does not need to be changed?

tophf: Sounds good.

Issue #726, opened 2019-11-23
Can't get regular expressions to work in extension blacklist; bugs out instead


Vagmer:
### What is the problem?

Violentmonkey's global URL blacklist is supposed to work with @exclude rules, which are supposed to support same regex as @include rules. However this fails when used in the global blacklist.

### How to reproduce it?
1. Try any one rule like the following in Violentmonkey's blacklist:
`@exclude /^https?://example.com/`
`@exclude /http://example.com/`
etc
![image](https://user-images.githubusercontent.com/12949474/69476342-bccafe00-0dd0-11ea-8705-a9ac6a0cb1bd.png)
2. Browse to any site.

### What is the expected result?
Userscripts will not launch only on _example.com_ (accordingly to the specific rule entered).

### What is the actual result?
Userscripts cease to work everywhere, and the Violentmonkey toolbar icon and popup do not reflect this (when this occurs the icon is not grayed out, popup does not specify "Blacklisted in Violentmonkey's settings").

### Environment
Violentmonkey v2.12.1 on Firefox 71.0b11 (64-bit) on Windows 10 x64.
This issue occurred for me on previous versions of Violentmonkey, Firefox and Windows, as well.

tophf:
It fails because `test` is not a function but is a RegExp#test reference.
Looks like I broke it in fcbd53ad or 79ed2ca7.

Vagmer:
Thanks for the fix! :) Using Violentmonkey v2.12.3, seems to work well so far.

I opted to try using some kind of general-purpose blacklist by default to prevent running scripts on bank, payment and financial sites. You might want to look into including (optionally) something like that by default or just mentioning how to do it in the extension's FAQ or something, considering. This actually isn't particularly important for the power or advanced user but is for the average user who would install others' scripts, maybe without even reading or understanding them at all.
`@exclude /^.*?://.*?(bank|pay|card).*?//`

It would be good additions for Violentmonkey's icon to be uniquely different when blacklist is triggered (currently it becomes the same as if Violentmonkey is currently manually disabled) and for the popup UI to mention the specific triggering exclusion rule in its text, so "Blacklisted in Violentmonkey's settings" text becomes more descriptive like "This page is blacklisted due to Violentmonkey's settings by rule: (newline) (_rule_)".

tophf:
I doubt the blacklist feature will get more love because personally I think it's fine as is for like 99.9% of users.

Vagmer:
[...]

PR #749, opened 2019-12-07
feat: show original rule text in blacklist reason in popup

tophf:
The idea was mentioned in #726.

image

### DIFF
--- a/src/popup/views/app.vue
+++ b/src/popup/views/app.vue
@@ -43,10 +43,10 @@
         <div class="flex-1" v-text="i18n('menuFindScripts')"></div>
       </div>
     </div>
-    <div
-      class="failure-reason"
-      v-if="failureReasonText"
-      v-text="failureReasonText" />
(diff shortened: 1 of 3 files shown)


src/popup/views/app.vue
```diff
@@ -185,9 +185,10 @@ export default {
       ].filter(Boolean).join(' ');
     },
     failureReasonText() {
+      const { injectable, blacklisted } = store;
       return (
-        !store.injectable && i18n('failureReasonNoninjectable')
-        || store.blacklisted && i18n('failureReasonBlacklisted')
+        !injectable && i18n('failureReasonNoninjectable')
+        || blacklisted && `${i18n('failureReasonBlacklisted')} <code>${blacklisted}</code>`
```
gera2ld: `blacklisted` needs to be escaped when using `v-html`.

tophf: Got it. Switched to a more declarative method.

PR #753, opened 2019-12-08
Update messages.yml

jojoPL1987:


src/_locales/pl/messages.yml
```diff
@@ -440,7 +442,7 @@ menuFindScripts:
   message: Znajdź skrypty dla tej strony
 menuMatchedFrameScripts:
   description: Label for menu listing matching scripts in sub-frames.
-  message: ''
+  message: Skrypty tylko w ramkach pomocniczych
```
gera2ld: Is the trailing space kept intentionally? If so, quotes are needed.

jojoPL1987: Mistake. I corrected it. Excuse me.

PS. When will I see newly added translations in Violentmonkey?

gera2ld: In the next release, I guess in a few days.

jojoPL1987: OK, thanks for info.

PR #754, opened 2019-12-10
Update Japanese translation

ScratchBuild:


gera2ld:
Please fix syntax error, which can be found in CI details.

PR #760, opened 2019-12-14
Update messages.yml

jojoPL1987:
Polish translation: 91%... so far...


tophf:
FYI, Violentmonkey can be translated now in Transifex directly, which you can do at your own pace in a convenient online editor: https://www.transifex.com/violentmonkey-1/violentmonkey-nex/

PR #764, opened 2019-12-19
feat: expose browser/platform info in GM_info.platform

### LINKED ISSUE #763: get browser info in user scripts
Is there an API to check browser inside user scripts?
for example 
```javascript
if (browser == 'firefox') {
// do stuff
}

### PR DESCRIPTION
**Pull request by tophf**
Implements and fixes #763.

```js
console.log(GM_info);
```

image

```js
const isChrome = GM_info.platform
   ? GM_info.platform.browserName === 'chrome'
   : /Chrome/i.test(navigator.userAgent);
```

### DIFF
--- a/src/common/ua.js
+++ b/src/common/ua.js
@@ -1,12 +1,46 @@
-/* global browser */
+import { browser } from './consts';
 
 // UA can be overriden by about:config in FF or devtools in Chrome
 // so we'll test for window.chrome.app which is only defined in Chrome
 // and for browser.runtime.getBrowserInfo in Firefox 51+
-export const isChrome = !!window.chrome?.app;
 
-// eslint-disable-next-line import/no-mutable-exports
-export let isFirefox = !!browser.runtime.getBrowserInfo;
-browser.runtime.getBrowserInfo?.().then((info) => {
-  isFirefox = info.name === 'Firefox' && parseFloat(info.version);
+/** @typedef UA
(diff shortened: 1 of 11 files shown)


src/common/ua.js
```diff
@@ -1,12 +1,46 @@
...
+  ready: {
+    value: Promise.all([
+      browser.runtime.getPlatformInfo(),
+      browser.runtime.getBrowserInfo?.(),
+    ]).then(([{ os, arch }, { name, version } = {}]) => {
+      Object.assign(ua, {
+        arch,
+        os,
+        browserName: name?.toLowerCase() || 'chrome',
+        browserVersion: version || navigator.userAgent.match(/Chrom\S+?\/(\S+)|$/)[1],
```
gera2ld: I think the regexp here is incorrect. It should be

```
/Chrome\/([\d.]+)|$/
```

tophf: `Chrom` prefix matches a theoretical fork that puts `Chromium` instead of `Chrome`
`\S` matches the future versions which may include letters.
On Chrome it works the same as the regexp you're proposing.

gera2ld: Oh I was wrong, I thought it was `\s`.
Not a big deal but I think there is little chance to be another name than `/Chrom(?:e|ium)/`.

tophf: Yep.

PR #772, opened 2019-12-26
fix: explicitly add window.wrappedJSObject to the wrapper

### LINKED ISSUE #765: [Firefox] window.wrappedJSObject is undefined if the userscript has any grant.
1. Install script:
```
// ==UserScript==
// @name Grant Test
// @grant GM_info
// ==/UserScript==
console.log("window.wrappedJSObject:", typeof window.wrappedJSObject);
```
2. Open https://github.com

In the console there will be a message:
`window.wrappedJSObject: undefined`

If set the `@grant` field to none or delete it, then the userscript works as it should:
`window.wrappedJSObject: object`

tophf:
Fixes #765.

FF doesn't expose wrappedJSObject via Object.getOwnPropertyNames so this PR explicitly adds it.

### DIFF
--- a/src/injected/web/gm-wrapper.js
+++ b/src/injected/web/gm-wrapper.js
@@ -50,6 +50,7 @@ export function wrapGM(script, code, cache, injectInto) {
     gm.window = unsafeWindow;
   } else {
     thisObj = getWrapper();
+    thisObj.wrappedJSObject = window.wrappedJSObject;
     gm.window = thisObj;
   }
   if (grant::includes('window.close')) {


src/injected/web/gm-wrapper.js
```diff
@@ -50,6 +50,7 @@ export function wrapGM(script, code, cache, injectInto) {
     gm.window = unsafeWindow;
   } else {
     thisObj = getWrapper();
+    thisObj.wrappedJSObject = window.wrappedJSObject;
```
gera2ld: How about checking the existence of `window.wrappedJSObject` first so that there won't be an undefined property on `wrapper` when it's not in Firefox?

tophf: I thought about that before and decided it doesn't matter.  Do you think it does?

tophf: ...it's essentially the same as `browser` that we add unnecessarily in Chrome.

gera2ld: Yep

PR #773, opened 2019-12-26
fix: don't send cookies twice in FF

### LINKED ISSUE #771: [Firefox] Duplicate cookies in v2.12.4
### What is the problem?
A recent change (probably in commit 335bf9660799a298b8776e98fca1e04533d8204a) causes GM_xmlhttpRequest to send a duplicate set of cookies, separated from the first set by a comma.

In private browsing mode, GM_xmlhttpRequest will send cookies which originated from both a private, and non-private context if either:
1. A cookie has been received in both private and non-private browsing mode
2. A cookie has been received in private browsing mode, and The GM_xmlhttpRequest itself receives a cookie while in private browsing mode

Usage of non-private cookies in private mode existed prior to v2.12.4. Usage of current private session cookies did not exist.

### How to reproduce it?
1. Serve a response with a cookie from any toy server
2. Make a bare-minimum GM_xmlhttpRequest to that server
3. Observe the raw headers from either client or server side

Example server path using Flask:
```
@app.route('/<cookiename>/')
def custom_cookie(cookiename):
  print(request.headers.get('Cookie'))
  response = Response('')
  response.set_cookie(cookiename, b'0')
  return response
```
Example GM_xmlhttpRequest:
`GM_xmlhttpRequest({method:'GET', url: 'http://localhost:8080/fromaddon/'});`

Cookies visiting `localhost:8080/frombrowser/` twice in normal mode:
```
None //first browser visit
frombrowser=0, frombrowser=0; //first addon visit
frombrowser=0; fromaddon=0; //second browser visit
frombrowser=0; fromaddon=0, frombrowser=0; fromaddon=0; //second addon visit
```
Cookies visiting `localhost:8080/frombrowser/` twice in private mode:
```
None //first browser visit
frombrowser=0; //first addon visit
frombrowser=0; //second browser visit
fromaddon=0, frombrowser=0; //second addon visit
```

Request should be sent with a single set of the appropriate cookies for the page.

`keya=valuea; keyb=valueb;`

Cookies should also be sent from either a private or non-private context in private browsing mode. Not both.

Cookies are duplicated.

`keya=valuea; keyb=valueb, keya=valuea; keyb=valueb;`

Cookies from both private and non-private contexts are sent in private browsing mode.

### Environment
- Browser: Firefox
- Browser version: 71.0
- OS: Windows 10

tophf:
Fixes #771.

Had to rename `requestHeaders` to `headers` to keep the added code on one line.

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -49,24 +49,25 @@ const HeaderInjector = (() => {
   const headersToInject = {};
   /** @param {chrome.webRequest.HttpHeader} header */
   const isVmVerify = header => header.name === VM_VERIFY;
+  const isNotCookie = header => !/^cookie$/i.test(header.name);
   const isSendable = header => header.name !== VM_VERIFY;
-  const isSendableAnon = header => isSendable(header) && !/^cookie$/i.test(header.name);
+  const isSendableAnon = header => isSendable(header) && isNotCookie(header);
   const apiEvents = {
     onBeforeSendHeaders: {
       options: ['requestHeaders', 'blocking', ...EXTRA_HEADERS],
       /** @param {chrome.webRequest.WebRequestHeadersDetails} details */
-      listener({ requestHeaders, requestId }) {
+      listener({ requestHeaders: headers, requestId }) {
         // only the first call during a redirect/auth chain will have VM-Verify header
-        const reqId = requestHeaders.find(isVmVerify)?.value || verify[requestId];
+        const reqId = headers.find(isVmVerify)?.value || verify[requestId];
         const req = reqId && requests[reqId];
         if (reqId && req) {
           verify[requestId] = reqId;
           req.coreId = requestId;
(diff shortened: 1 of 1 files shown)


gera2ld:
I think we should have three groups of cookies:

- `browserCookies`: cookies provided by browser
- `userCookies`: cookies provided by user, from `GM_xmlhttpRequest`
- `extCookies`: cookies loaded from `browser.cookies`

And merge them at `onBeforeSendHeaders` with following steps:

- if `anonymous` is true, discard all.
- otherwise:
  - if `browserCookies` is empty, patch with `extCookies`.
  - append `userCookies`.

Because we can only get `browserCookies` in `onBeforeSendHeaders`, `extCookies` can be stored earlier with a preserved header name.

tophf:
And what's wrong with the simple approach in this PR?

gera2ld:
Just checked again, the result is exactly the same, the simple approach is great.

PR #778, opened 2019-12-30
fix: handle gmxhr blob/arraybuffer > 64MB

### LINKED ISSUE #777: ViolentMonkey's GM_xmlhttpRequest sometimes won't call onload
### What is the problem?
Using `GM_xmlhttpRequest` to download some big(50~100MB) binary blob sometimes won't call `onload`.
The problem with my script: 

### How to reproduce it?
1. Install https://greasyfork.org/zh-TW/scripts/369400-local-youtube-downloader in ViolentMonkey
2. Go to https://www.youtube.com/watch?v=nh5w1DkY4vY
3. Click "Show/Hide Links" then "Download hi res mp4 in one click"

Successfully download the video in ArrayBuffer.

Stuck at 87.74/87.79MB, but it works in TamperMonkey.
The download is triggered at https://github.com/maple3142/browser-extensions/blob/a899753f2188911e25b1b2ecce8884f367de7752/scripts/local-youtube-dl.user.js#L266

<details>
<summary>Logs</summary>
<link rel=preload> must have a valid `as` value
[Deprecation] HTML Imports is deprecated and will be removed in M80, around February 2020. Please use ES modules instead. See https://www.chromestatus.com/features/5144752345317376 and https://developers.google.com/web/updates/2019/07/web-components-time-to-upgrade for more details.
/pagead/viewthroughconversion/962985656/?backend=innertube&cname=1&cver=2_20191218&foc_id=XsZUn6dalbxrnVsn7PezzQ&label=followon_view&ptype=no_rmkt&random=390273645:1 Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
googleads.g.doubleclick.net/pagead/id:1 Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
Local YouTube Downloader.user.js:11979 YTDL: default language: zh-tw
Local YouTube Downloader.user.js:11979 YTDL: shadowHost: <div style=​"display:​ block;​">​…​</div>​
Local YouTube Downloader.user.js:9049 Download the Vue Devtools extension for a better development experience:
https://github.com/vuejs/vue-devtools
Local YouTube Downloader.user.js:9058 You are running Vue in development mode.
Make sure to turn on production mode when deploying for production.
See more tips at https://vuejs.org/guide/deployment.html
www.google.com.tw/pagead/lvz?pg=watch&req_ts=1577679746&evtid=ADszL7NpKpGwyyM_X9VCrwz36IsolyiEYKttMB7Yv36JC9GVww5_VPZr_NmqtysWZQb4GjscatZzSDnCvbcyNlV6slZoAmjEXw&sigh=ADRnf4YkBB2_RT60yu8q4u0y-Zo5SlKdGQ:1 Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
www.google.com/pagead/lvz?pg=watch&req_ts=1577679746&evtid=ADszL7OrFfiC5lExDgiQC4R6um3w-cb4YQwIfuan6x8oPidvM8s55omCMiBEKdtv54mckPQI-c2Ayq7xoNDXJAAhwpDJBHP1Fw&sigh=ADRnf4ZZ556L95nrA6nTvjOq1sjX9jkmcA:1 Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
static.doubleclick.net/instream/ad_status.js:1 Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
network.js:15 [Deprecation] chrome.loadTimes() is deprecated, instead use standardized API: nextHopProtocol in Navigation Timing 2. https://www.chromestatus.com/features/5637885046816768.
g.onreadystatechange @ network.js:15
Local YouTube Downloader.user.js:11979 YTDL: page change: null -> https://www.youtube.com/watch?v=nh5w1DkY4vY
Local YouTube Downloader.user.js:11979 YTDL: start loading new video: nh5w1DkY4vY
Local YouTube Downloader.user.js:11979 YTDL: workerGetVideo start: nh5w1DkY4vY https://www.youtube.com/yts/jsbin/player_ias-vfl22ubNH/zh_TW/base.js
59068a0a-12f6-4657-b390-5b5ae86b79fc:10 YTDL: parsedecsig result: a=>{var Tt={m2:function(a){a.reverse()},
vf:function(a,b){var c=a[0];a[0]=a[b%a.length];a[b%a.length]=c},
uB:function(a,b){a.splice(0,b)}};
a=a.split("");Tt.vf(a,26);Tt.uB(a,1);Tt.m2(a,76);Tt.uB(a,2);Tt.m2(a,21);Tt.uB(a,1);Tt.vf(a,45);Tt.m2(a,24);return a.join("")}
59068a0a-12f6-4657-b390-5b5ae86b79fc:10 YTDL: video nh5w1DkY4vY data: Object
59068a0a-12f6-4657-b390-5b5ae86b79fc:10 YTDL: video nh5w1DkY4vY playerResponse: Object
59068a0a-12f6-4657-b390-5b5ae86b79fc:10 YTDL: video nh5w1DkY4vY stream: Array(1)
59068a0a-12f6-4657-b390-5b5ae86b79fc:10 YTDL: video nh5w1DkY4vY adaptive: Array(16)
59068a0a-12f6-4657-b390-5b5ae86b79fc:10 YTDL: video nh5w1DkY4vY result: Object
Local YouTube Downloader.user.js:11979 YTDL: workerGetVideo end: Object
Local YouTube Downloader.user.js:11979 YTDL: video loaded: nh5w1DkY4vY
Local YouTube Downloader.user.js:11979 YTDL: youtube ui lang: zh-TW
Local YouTube Downloader.user.js:11979 YTDL: ytdl lang: zh-tw
The resource https://r2---sn-u5oxu-u2xe.googlevideo.com/generate_204 was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
The resource https://r2---sn-u5oxu-u2xe.googlevideo.com/generate_204?conn2 was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
The resource https://i.ytimg.com/generate_204 was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
6A cookie associated with a cross-site resource at <URL> was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at <URL> and <URL>.
2googleads.g.doubleclick.net/pagead/id:1 Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
</details>

### Environment
- Browser: Chrome
- Browser version: 79.0.3945.88
- OS: Windows 10

tophf:
Fixes #777.

1. extension messaging has a 128MB limit on the payload size so buffer2string is limited to 64MB as it uses just one byte out of two. **Solution:** URL.createObjectURL is used to transfer a small blob URL, then its contents is retrieved explicitly using XHR in the content script. This workaround is enabled only when `responseType` parameter is `blob` or `arraybuffer`. Ideally we could do that for all response types but I think we shouldn't do it unless someone reports at least one sensible use case.

2. extension API doesn't seem to guarantee the order of messages is preserved so now `xhrCallbackWrapper` will wait for the previous message before sending the next one.

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -1,6 +1,7 @@
 import {
-  getUniqId, request, i18n, buffer2string, isEmpty,
+  getUniqId, request, i18n, isEmpty,
 } from '#/common';
+import { objectPick } from '#/common/object';
 import ua from '#/common/ua';
 import cache from './cache';
 import { isUserScript, parseMeta } from './script';
@@ -107,10 +108,12 @@ const HeaderInjector = (() => {
   };
 })();
 
-export function getRequestId() {
+export function getRequestId(eventsToNotify = []) {
+  eventsToNotify.push('loadend');
   const id = getUniqId();
   requests[id] = {
     id,
+    eventsToNotify,
     xhr: new XMLHttpRequest(),
   };
   return id;
@@ -118,41 +121,48 @@ export function getRequestId() {
 
 function xhrCallbackWrapper(req) {
   let lastPromise = Promise.resolve();
-  const { xhr } = req;
+  let blobUrl;
+  const { id, xhr } = req;
   return (evt) => {
-    const res = {
-      id: req.id,
-      type: evt.type,
-      resType: xhr.responseType,
-      contentType: xhr.getResponseHeader('Content-Type') || 'application/octet-stream',
-    };
+    if (evt.type === 'loadend') clearRequest(req);
+    if (!req.cb) return;
+    const isBlob = xhr.responseType === 'blob';
     const data = {
       finalUrl: xhr.responseURL,
       readyState: xhr.readyState,
       responseHeaders: req.responseHeaders || xhr.getAllResponseHeaders(),
       status: xhr.status,
       statusText: xhr.statusText,
+      response: isBlob && xhr.response ? getBlobUrl(req) : xhr.response,
+      ...getResponseText(),
+      ...('loaded' in evt) && objectPick(evt, ['lengthComputable', 'loaded', 'total']),
     };
-    res.data = data;
+    lastPromise = lastPromise.then(() => req.cb({
+      id,
+      data,
+      isBlob,
+      contentType: xhr.getResponseHeader('Content-Type') || 'application/octet-stream',
+      type: evt.type,
+    }));
+  };
+  function getBlobUrl() {
+    if (!blobUrl) {
+      blobUrl = URL.createObjectURL(xhr.response);
+      setTimeout(expireBlobUrl, 60e3);
+    }
+    return blobUrl;
+  }
+  function expireBlobUrl() {
+    URL.revokeObjectURL(blobUrl);
+    blobUrl = null;
+  }
+  function getResponseText() {
     try {
-      data.responseText = xhr.responseText;
+      return { responseText: xhr.responseText };
     } catch (e) {
       // ignore if responseText is unreachable
     }
-    if (evt.type === 'progress') {
-      ['lengthComputable', 'loaded', 'total'].forEach((key) => {
-        data[key] = evt[key];
-      });
-    }
-    if (evt.type === 'loadend') clearRequest(req);
-    else if (xhr.readyState >= XMLHttpRequest.LOADING) HeaderInjector.del(req.id);
-    lastPromise = lastPromise.then(() => {
-      data.response = xhr.response && xhr.responseType === 'arraybuffer'
-        ? buffer2string(xhr.response)
-        : xhr.response;
-      if (req.cb) req.cb(res);
-    });
-  };
+  }
 }
 
 function isSpecialHeader(lowerHeader) {
@@ -201,7 +211,7 @@ export async function httpRequest(details, src, cb) {
       });
     }
     if (timeout) xhr.timeout = timeout;
-    if (responseType) xhr.responseType = 'arraybuffer';
+    if (responseType) xhr.responseType = 'blob';
     if (overrideMimeType) xhr.overrideMimeType(overrideMimeType);
     if (shouldSendCookies) {
       const cookies = await browser.cookies.getAll({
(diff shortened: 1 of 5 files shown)


tophf:
Er, the real cause of #777 is different: `loadend` isn't necessarily last... Investigating.

tophf:
Should be fixed now.

tophf:
Forced pushes are my progressively more declarative refactoring of xhrCallbackWrapper.

src/background/utils/requests.js
```diff
@@ -107,52 +108,61 @@ const HeaderInjector = (() => {
...
     };
-    res.data = data;
+    await lastPromise;
+    lastPromise = req.cb({
+      id,
+      data,
+      contentType: xhr.getResponseHeader('Content-Type') || 'application/octet-stream',
+      resType: xhr.responseType,
+      type: evt.type,
+    });
```
gera2ld: This is not equivalent to the current behavior.

Take this example:
- While processing *callback1*, there are two new tasks being sceduled, say *callback2* and *callback3*, both waiting for `lastPromise`, a promise for *callback1*.
- After *callback1* is resolved, `lastPromise` will be set twice immediately, once by *callback2* and once by *callback3*.
- As a result, *callback2* and *callback3* are executed concurrently, not as expected.

So `await` is not supposed to be used here because our real purpose is to update `lastPromise` immediately and we don't care about when it is resolved.

src/background/utils/requests.js
```diff
@@ -107,52 +108,61 @@ const HeaderInjector = (() => {
...
+      blobUrl = URL.createObjectURL(xhr.response);
+      setTimeout(expireBlobUrl, 60e3);
+    }
+    return blobUrl;
+  }
+  function expireBlobUrl() {
+    URL.revokeObjectURL(blobUrl);
+    blobUrl = null;
+  }
+  function tryOrIgnore(fn, ...args) {
```
gera2ld: `tryOrIgnore` does not depend on any scoped variables so it's better to move it to the top level.

Personally I prefer a higher order function like this:

```js
function ignoreError(fn) {
  const wrapped = (...args) => {
    try {
      return fn(...args);
    } catch {
      // ignore
    }
  };
  return wrapped;
}
```

so that we don't need to change the way we call `fn`:
```js
fn(a, b);
ignoreError(fn)(a, b);

// as a drop-in replacement
fn = ignoreError(fn);
fn(a, b);
```

src/background/utils/requests.js
tophf: Indeed.

src/background/utils/requests.js
tophf: That's a leftover from my earlier approach when I thought it'd be helpful to use this (or similar) function in more places but I found only one or two where it would improve readability. You're right so now I've reverted to a simple scoped function.

PR #780, opened 2020-01-01
feat: recolor script-item on update error + refactor update.js

### LINKED ISSUE #67: An Update Fail Notification Icon
An icon in the listing to notify if an auto-update fails because the update URL is no longer valid.

tophf:
Fixes #67.

Border, icon, message are reddened:

image

And a standard DOM tooltip with the URL is shown on hover:

image

The feature itself was trivial so to make things more interesting I've refactored update.js to use async/await. Hopefully the new code has a more obvious execution flow.

### DIFF
--- a/src/background/utils/update.js
+++ b/src/background/utils/update.js
@@ -9,100 +9,92 @@ const processes = {};
 const NO_HTTP_CACHE = {
   'Cache-Control': 'no-cache, no-store, must-revalidate',
 };
+const OPTIONS = {
+  meta: {
+    headers: { ...NO_HTTP_CACHE, Accept: 'text/x-userscript-meta' },
+  },
+  script: {
+    headers: NO_HTTP_CACHE,
+  },
+};
 
-function doCheckUpdate(script) {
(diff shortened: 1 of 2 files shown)


tophf:
I've been refactoring other background js, see https://github.com/tophf/violentmonkey/commit/144c167536f01ae3e766f399ef5979a8c7aed350: `async` (using an automatic converter) and some other new stuff like optional chaining and object spread, 2) split `commands` registration per module. If you're not entirely opposed to this, I can either add that commit here or open another PR after this one is handled.

gera2ld:
> I've been refactoring other background js, see [master...tophf:minor-refactor-bg](https://github.com/violentmonkey/violentmonkey/compare/master...tophf:minor-refactor-bg): `async` (using an automatic converter) and some other new stuff like optional chaining and object spread, 2) split `commands` registration per module. If you're not entirely opposed to this, I can either add that commit here or open another PR after this one is handled.

Another PR is better for this, I think it may take quite some time to validate the changes.

PR #790, opened 2020-01-10
refactor bg

tophf:
Despite the big amount of moved stuff the underlying changes are simple: 
* converted functions to `async` using an automatic converter + some manual tweaks
* optional chaining and other new stuff like object spread
* `commands` registration is split per module so the main index.js is now small
* `sendTabCmd` is used for tabs/frames 
* `sendCmd` ignores errors when called in the background script


tophf:
I wonder if I should join all `refactor` commits into one?

gera2ld:
> I wonder if I should join all `refactor` commits into one?

Nope, it looks clear to me this way.

PR #807, opened 2020-01-17
feat: add scope name|code|all to dashboard search filter

tophf:
Default mode is `name`, which restores the old behavior (before code search was enabled by default). The reason is that searching code by default may often produce irrelevant results due to matching a common JS term or a common metablock value.

image

image

I've used a dropdown selector instead of a simple `[x] in code` because we may want to add other modes in the future like maybe `values` or individual meta keys.

Collateral:

* common/options getOption() now uses options-defaults.js to get the fallback value instead of the second parameter which wasn't used anywhere (AFAICT it was added just in case, only because options-defaults.js wasn't exposed).
* a tooltip with regexp search hint is added to the search icon
* refactor: deduplicate/simplify hookSetting() usage so it automatically calls options.get when used in the declarative invocation fashion via `{ prop: 'name' }`

### DIFF
--- a/src/common/hook-setting.js
+++ b/src/common/hook-setting.js
@@ -13,12 +13,23 @@ options.hook((data) => {
   });
 });
 
-export default function hook(key, update) {
-  let list = hooks[key];
-  if (!list) {
-    list = [];
-    hooks[key] = list;
+/**
+ * When an option is updated elsewhere (or when a yet unresolved options.ready will be fired),
+ * calls the specified `update` function or assigns the specified `prop` in `target` object.
+ * Also, when the latter mode is used, option.get() is called explicitly right away,
+ * but only if options.ready is resolved or `transform` function is specified.
+ * @param {string} key - option name
+ * @param {function(value) | { target, prop, transform }} update - either a function or the config object
+ * @return {function}
+ */
+export default function hookSetting(key, update) {
+  const { target } = update;
+  if (target) {
+    const { prop, transform } = update;
+    update = value => { target[prop] = transform ? transform(value) : value; };
+    if (transform || options.ready.indeed) update(options.get(key));
   }
+  const list = hooks[key] || (hooks[key] = []);
   list.push(update);
   return () => {
     const i = list.indexOf(update);
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -184,6 +184,15 @@ filterExecutionOrder:
 filterLastUpdateOrder:
   description: Label for option to sort scripts by last update time.
   message: last update time
+filterScopeAll:
+  description: Option in dashboard's search scope filter.
+  message: All
+filterScopeCode:
+  description: Option in dashboard's search scope filter.
+  message: Code
+filterScopeName:
+  description: Option in dashboard's search scope filter.
+  message: Name
 headerRecycleBin:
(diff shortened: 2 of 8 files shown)


.babelrc.js
```diff
@@ -10,5 +10,7 @@ module.exports = {
   ],
   plugins: [
     '@babel/plugin-proposal-function-bind',
+    // `loose:true` is fine because we don't use document.all
+    ['@babel/plugin-proposal-nullish-coalescing-operator', { loose: true }],
```
gera2ld: This is already included in the latest `@babel/preset-env`, so we just need an upgradation.

tophf: Rebased.

tophf:
I've force-pushed a couple of cosmetic changes and I think it's ready now.

src/common/hook-setting.js
```diff
@@ -13,7 +13,24 @@ options.hook((data) => {
...
+ * Also, when `prop` is specified, option.get() is called explicitly right away,
+ * but only if options.ready is resolved or a `handle` function is specified.
+ * @param {string} key - option name
+ * @param {function(value) | Object} update - either a function or the target object
+ * @param {Object} [config]
+ * @param {string} config.prop - property name to be assigned in the target
+ * @param {function(value)} [config.handle] - transformer function to call on the value
+ * @return {function}
+ */
+export default function hook(key, update, { prop, handle } = {}) {
```
gera2ld: It seems that `update` only appears as an object. Maybe we don't need to support it as a function any more.
ps. I think the relationship between `update` and `prop` is a little hard to understand. Though it's not a big deal, I prefer the original style.

tophf: > It seems that `update` only appears as an object.

vm-sync.vue uses a function but I can convert it to use the declarative syntax too

> ps. I think the relationship between `update` and `prop` is a little hard to understand. 

Do you mean the code inside the function is hard to understand or the usage?

> Though it's not a big deal, I prefer the original style.

You mean you like this:
```js
hookSetting('filters.searchScope', (value) => {
  filters.searchScope = value;
});
hookSetting('filters.showEnabledFirst', (value) => {
  filters.showEnabledFirst = value;
});
hookSetting('filters.sort', (value) => {
  filters.sort.set(value);
});
options.ready.then(() => {
  filters.searchScope = options.get('filters.searchScope');
  filters.sort.set(options.get('filters.sort'));
  filters.showEnabledFirst = options.get('filters.showEnabledFirst');
});
```
more than this:
```js
Object.keys(filters).forEach(name => {
  hookSetting(`filters.${name}`, filters, { prop: name });
});
```
right?

tophf: What about flattening it like `hookSetting(key, targetObj, propName, transformValueFunc)`?

gera2ld: Changing `update` to `targetObj` looks good to me.

> ps. I think the relationship between update and prop is a little hard to understand.

I mean if `update` is a function, `prop` does not make sense. If `update` should be kept any way, I prefer `hookSetting(key, update | { target, prop, transform })`.

tophf: Ah, yeah, I've considered that too, but then went the implicit/confusing route. Okay let's do `hookSetting(key, update | { target, prop, transform })`

tophf: Pushed the new version.

PR #810, opened 2020-01-18
fix(objectSet): set an array by default if key is number

gera2ld:
```js
objectSet({}, ['a', 1], 'v') // -> { a: [undefined, 'v'] }
objectSet({}, ['a', '1'], 'v') // -> { a: [undefined, 'v'] }
objectSet({}, ['a', 'b'], 'v') // -> { a: { b: 'v' } }
```

With this we don't have to use the double array hack.
This is also the default behavior of `lodash.set`, which this function is trying to mimic.

### DIFF
--- a/src/common/object.js
+++ b/src/common/object.js
@@ -25,7 +25,7 @@ export function objectSet(obj, rawKey, val) {
   let sub = root;
   const lastKey = keys.pop();
   keys.forEach((key) => {
-    sub = sub[key] || (sub[key] = {});
+    sub = sub[key] || (sub[key] = /^\d+$/.test(key) ? [] : {});
   });
   if (typeof val === 'undefined') {
     delete sub[lastKey];


tophf:
This might be interesting but I don't see where it can be used currently because the only case I found so far was for `frameId` but it's not absolutely necessary as the object works just fine as well. Maybe it's better to wait until the need arises?

gera2ld:
OK, I just want to keep the implementation for such functions in Violentmonkey almost the same as that in popular libraries so that we don't have to learn from source code before using it.

ps. `objectSet` is just a simplified implementation for `_.set` because the lodash version involves some code that are not needed by VM.

tophf:
I will remember this info. Using a numeric index in the path was among the ideas I've tried locally, but since I've never been using libraries/frameworks it didn't even occur to me to google up the existing solutions.

PR #811, opened 2020-01-18
refactor: restore simple hookSetting, prefetch options

tophf:
The benefit of this approach is that it's easier to see usages of a symbol, particularly via "Find usages" command in IDE.

Added: browser.js prefetches the options while an extension page loads to drastically reduce the probability of options being not ready. It reduces the UI flicker, for example [re]loading `options/index.html#settings` now shows the real settings in the first paint of the page.

### DIFF
--- a/src/common/ui/setting-check.vue
+++ b/src/common/ui/setting-check.vue
@@ -35,8 +35,8 @@ export default {
     },
   },
   async created() {
-    await options.ready;
-    this.revoke = hookSetting(this.name, { target: this, prop: 'value' });
+    if (!options.ready.indeed) await options.ready;
+    this.revoke = hookSetting(this.name, val => { this.value = val; });
     this.$watch('value', this.onChange);
   },
   beforeDestroy() {
--- a/src/common/ui/setting-text.vue
+++ b/src/common/ui/setting-text.vue
@@ -31,12 +31,13 @@ export default {
       error: null,
     };
   },
-  created() {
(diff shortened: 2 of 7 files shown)


src/common/ui/setting-check.vue
```diff
@@ -36,7 +36,7 @@ export default {
   },
   async created() {
     await options.ready;
-    this.revoke = hookSetting(this.name, { target: this, prop: 'value' });
+    this.revoke = hookSetting(this.name, val => { this.value = val; });
```
tophf: `val` is used to make it less similar to `this.value`

src/common/ui/setting-text.vue
```diff
@@ -32,11 +32,11 @@ export default {
     };
   },
   created() {
-    const transform = this.json
+    const handle = this.json
       ? (value => JSON.stringify(value, null, '  '))
       // XXX compatible with old data format
       : (value => (Array.isArray(value) ? value.join('\n') : value || ''));
-    this.revoke = hookSetting(this.name, { target: this, prop: 'value', transform });
+    this.revoke = hookSetting(this.name, val => { this.value = handle(val); }, true);
```
tophf: maybe it'd be better to use `await options.ready` instead of `true` like we do in other places where hookSettings is used?

tophf: Yep, did so.

PR #814, opened 2020-01-18
feat: per-script update notification option

### LINKED ISSUE #333: Per-script "notify updates" option
Scripts have different level of trust and frequency of updates, hence it can be convenient for scripts to have some settings that are different from global, e.g.
- Automatically check scripts for updates
- Notify script updates

tophf:
Fixes #333.

* A tri-state option to customize update notifications per script. The indeterminate state `?` means "not configured" so the current global value in Violentmonkey's settings will be used.

  image

* A global option to ignore the per-script option. Isn't really useful normally, I'd say it's an emergency kill-switch to disable the notifications globally and ignore the per-script option.

  image

Collateral:

* converted two related methods to async
* added value transformation function as a third parameter in `objectPick`

Didn't test it extensively yet.

### DIFF
--- a/src/common/object.js
+++ b/src/common/object.js
@@ -48,9 +48,10 @@ export function objectPurify(obj) {
   return obj;
 }
 
-export function objectPick(obj, keys) {
+export function objectPick(obj, keys, transform) {
   return keys.reduce((res, key) => {
-    const value = obj ? obj[key] : null;
+    let value = obj?.[key];
+    if (transform) value = transform(value);
     if (value != null) res[key] = value;
     return res;
   }, {});
--- a/src/options/views/edit/settings.vue
+++ b/src/options/views/edit/settings.vue
@@ -6,10 +6,10 @@
         <input type="checkbox" v-model="config.shouldUpdate">
         <span v-text="i18n('labelAllowUpdate')"></span>
       </label>
-      <label class="ml-2">
-        <span v-text="i18n('labelNotifyUpdates')"/>
+      <label class="flex">
+        <span v-text="i18n('labelNotifyUpdates')" class="mr-1"/>
         <select v-model="config.notifyUpdates">
-          <option value="" v-text="i18n('useGlobalSetting')"></option>
+          <option value="" v-text="i18n('useGlobalSetting')"/>
           <option value="1" v-text="i18n('buttonEnable')"/>
           <option value="0" v-text="i18n('buttonDisable')"/>
         </select>
--- a/src/options/views/edit/settings.vue
+++ b/src/options/views/edit/settings.vue
@@ -6,6 +6,11 @@
         <input type="checkbox" v-model="config.shouldUpdate">
         <span v-text="i18n('labelAllowUpdate')"></span>
       </label>
+      <label class="ml-2">
+        <input type="checkbox" v-model="config.notifyUpdates"
+               :indeterminate.prop="config.notifyUpdates == null">
+        <span v-text="i18n('labelNotifyUpdates')"/>
+      </label>
     </div>
     <h4 v-text="i18n('editLabelMeta')"></h4>
     <div class="form-group flex">
@@ -113,6 +118,11 @@ export default {
       };
     },
   },
+  watch: {
+    'config.notifyUpdates'(val, oldVal) {
+      if (val && oldVal === false) this.config.notifyUpdates = null;
+    },
+  },
 };
 </script>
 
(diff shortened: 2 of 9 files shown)


src/common/object.js
```diff
@@ -48,10 +48,10 @@ export function objectPurify(obj) {
   return obj;
 }
 
-export function objectPick(obj, keys) {
+export function objectPick(obj, keys, transform) {
   return keys.reduce((res, key) => {
     const value = obj ? obj[key] : null;
-    if (value != null) res[key] = value;
+    if (value != null) res[key] = transform ? transform(value) : value;
```
tophf: Maybe when transform is specified it should be called even if `value` is null?

```js
if (transform) res[key] = transform(value);
else if (value != null) res[key] = value;
```

gera2ld:
I think the tri-state checkbox is a little hard to understand, what do you think of select box? like the custom `Run-at` select box in script settings.

tophf:
I certainly agree a selector would be more obvious, I just didn't want to add three more localizable strings... So you're saying I don't need to worry about that?

tophf:
Ah, I can reuse messages buttonEnable='Enable' + buttonDisable='Disable' and add just one 'N/A' or 'Not set' or 'Use global setting'?

gera2ld:
> Ah, I can reuse messages buttonEnable='Enable' + buttonDisable='Disable' and add just one 'N/A' or 'Not set' or 'Use global setting'?

Good idea, maybe we can normalize the phrases to make them reusable as option values in many places.
And I don't think localization is a problem.

tophf:
>maybe we can normalize the phrases

These are fine as is because they express an action by design so the third option is also an action: "Use global setting".
I didn't know how to prefix this new "universal" message though so didn't add any prefix at all. I've added the changes in a new commit so if it's fine you can do "squash and merge", otherwise I'll force-push a consolidated commit after addressing your feedback.

src/options/views/edit/settings.vue
```diff
@@ -6,6 +6,14 @@
         <input type="checkbox" v-model="config.shouldUpdate">
         <span v-text="i18n('labelAllowUpdate')"></span>
       </label>
+      <label class="flex">
+        <span v-text="i18n('labelNotifyUpdates')" class="mr-1"/>
+        <select v-model="config.notifyUpdates">
+          <option value="" v-text="i18n('useGlobalSetting')"/>
+          <option value="1" v-text="i18n('buttonEnable')"/>
+          <option value="0" v-text="i18n('buttonDisable')"/>
```
gera2ld: `buttonEnable` and `buttonDisable` were supposed to be two actions, or verbs, I am not sure if they are proper here. Maybe it should be `enabled` and `disabled`?

@tophf What's your point of view?

tophf: Currently the choices are:

* Use global setting
* Enable
* Disable

All three are verbs so I don't think it's weird. For example the old option "Allow update" is also an imperative verb.

gera2ld: OK, get it.

tophf: If an opportunity presents itself like another similar option, we can add `On`, `Off` or something generic of the kind. As for `Use global setting`, right now I don't see how to convert it to a noun-based state description while keeping it short and self-evident, but that's a problem for the future.

tophf: @gera2ld hmm you got me thinking and now I'm not sure... What about this:

[x] Allow update + [notify on update | don't notify on update | use global notification mode]

Each string of these three will be an entire option inside the selector.
The plus sign is a hardcoded text.

gera2ld: I prefer the action + status version, i.e. `Notify on update: [true | false | (default)]`, because the options are shorter and easier to read in my opinion.

tophf: What **exactly** does `(default)` do from a user's viewpoint? It's too vague.

tophf:
```
[x] Allow update
[x] Notify on update or [x] use the global option in settings
```

The notification checkbox is automatically disabled when the global is checked.
Mm?

tophf: Another idea is a radio button:

```
Allow update [x]
Update notification: ( ) on ( ) off (x) use global setting
```

gera2ld: What's the difference between radio button and select box?
I think the last one looks good for me, because the options are simple enough so that I don't have to think much about what each option means behind the words.

tophf: The differences are:

* all the options are immediately visible
* there's no long empty rectangle when a shorter option is selected
* one click less to switch

gera2ld: Sounds great. Much better than hierarchical options, like one option depends on or affects another.

PR #816, opened 2020-01-19
fix: GoogleDrive auth in Edge

### LINKED ISSUE #815: Chromium Edge: Stuck on authorizing Google Drive
# What is the problem?

Cannot sync scripts from Google Drive with Chromium based Edge.

### How to reproduce it?
1. Select Google Drive and click Authorize.
2. Grant the access from the popup page. 

It should start syncing.

It kept showing "Authorizing".

This log seems to relevant:
```
index.html:1 Uncaught (in promise) Object
index.js:1 POST https://www.googleapis.com/oauth2/v4/token 400
(anonymous) @ common.js:1
g @ common.js:1
(anonymous) @ index.js:1
Promise.then (async)
loadData @ index.js:1
authorized @ index.js:1
checkAuth @ index.js:1
(anonymous) @ index.js:1
(anonymous) @ index.js:1
(anonymous) @ index.js:1
index.html:1 Uncaught (in promise) {url: "https://www.googleapis.com/oauth2/v4/token", data: {…}, xhr: XMLHttpRequest, status: 400}
Promise.then (async)
checkAuth @ index.js:1
(anonymous) @ index.js:1
(anonymous) @ index.js:1
(anonymous) @ index.js:1
```
```json
{error: "invalid_grant", error_description: "Malformed auth code."}
```

[xhr_token.har.txt](https://github.com/violentmonkey/violentmonkey/files/4082245/xhr_token.har.txt)

### Environment
- Browser: Edge 
- Browser version: 79.0.309.68 (Official build) (64-bit)
- OS: Windows 10 18363

tophf:
Fixes #815.

### DIFF
--- a/src/background/sync/googledrive.js
+++ b/src/background/sync/googledrive.js
@@ -125,7 +125,7 @@ const GoogleDrive = BaseService.extend({
     if (url.startsWith(redirectUri)) {
       this.authState.set('authorizing');
       this.authorized({
-        code: url.split('#')[0].slice(redirectUri.length),
+        code: decodeURIComponent(url.split('#')[0].slice(redirectUri.length)),
       })
       .then(() => this.checkSync());
       return true;
--- a/src/background/utils/search.js
+++ b/src/background/utils/search.js
@@ -1,13 +1,13 @@
 export function loadQuery(string) {
   return string.split('&').reduce((data, piece) => {
-    const parts = piece.split('=');
-    data[decodeURIComponent(parts[0])] = decodeURIComponent(parts[1]);
+    const [key, val] = piece.split('=').map(decodeURIComponent);
+    data[key] = val;
     return data;
   }, {});
 }
 
 export function dumpQuery(dict) {
-  return Object.keys(dict)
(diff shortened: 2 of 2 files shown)


src/background/sync/googledrive.js
```diff
@@ -125,7 +125,7 @@ const GoogleDrive = BaseService.extend({
     if (url.startsWith(redirectUri)) {
       this.authState.set('authorizing');
       this.authorized({
-        code: url.split('#')[0].slice(redirectUri.length),
+        code: decodeURIComponent(url.split('#')[0].slice(redirectUri.length)),
```
gera2ld: Is this safe for other browsers?

tophf: I've verified it works in Chrome, yes.

tophf: Actually I'm surprised the old code worked...

gera2ld: Works in Firefox too.

gera2ld: > Actually I'm surprised the old code worked... because I would have expected that URL parameters are always encoded.

That's true. 🤣

PR #817, opened 2020-01-19
chore: enable babel loose mode

tophf:
Since we use object spread inside plain objects, its behavior is identical to Object.assign, so we can get rid of the huge polyfill added by Babel in each module where spread is used. Not chunks but modules, meaning several kB of slow and bloated and unnecessary polyfills overall.

Edit: we'll enable babel loose mode globally

the only affected transforms/plugins for our browserlist:
* object-rest-spread will use Object.assign
* nullish-coalescing-operator will use == with null

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -6,6 +6,7 @@ module.exports = {
         modules: false,
       },
       useBuiltIns: false,
+      loose: true,
     }],
   ],
   plugins: [


gera2ld:
I'm considering using `loose` mode for all plugins in `@babel/preset-env`.

```js
    ['@babel/preset-env', {
      ...process.env.BABEL_ENV !== 'test' && {
        modules: false,
      },
      useBuiltIns: 'usage',
      corejs: 3,
      loose: true,
    }],
```

tophf:
Uhm, `useBuiltIns: 'usage'` adds the even bigger polyfill so injected-web.js becomes 73kB.

gera2ld:
My fault, wrong copy.

tophf:
How do you verify it doesn't cause problems? Is there a list of all affected transforms/plugins?

tophf:
I've manually checked the documentation on https://babeljs.io/docs/en/plugins/ and it looks like the only affected plugin/transform are `object-rest-spread` (configured in this PR) and `nullish-coalescing-operator` (not configured here). All the rest are natively supported in browser versions from our browserlist.

So I guess it's safe to use the global loose mode.
But why don't you want to explicitly configure the transforms?

gera2ld:
> But why don't you want to explicitly configure the transforms?

Because it is a little complex.
I think we both prefer the compilation mode that generates the least code, which is exactly what `loose` mode aims at. Since there is an easy way to enable it, I think we should make use of it.

tophf:
Updated.

PR #821, opened 2020-01-21
fix: rearrange and simplify notify update option

tophf:
image

### DIFF
--- a/src/options/views/edit/settings.vue
+++ b/src/options/views/edit/settings.vue
@@ -6,13 +6,13 @@
         <input type="checkbox" v-model="config.shouldUpdate">
         <span v-text="i18n('labelAllowUpdate')"></span>
       </label>
-      <label class="flex">
-        <span v-text="i18n('labelNotifyUpdates')" class="mr-1"/>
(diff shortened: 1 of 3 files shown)


src/options/views/edit/settings.vue
```diff
@@ -6,13 +6,13 @@
...
-          <option value="1" v-text="i18n('buttonEnable')"/>
-          <option value="0" v-text="i18n('buttonDisable')"/>
-        </select>
+      <span v-text="i18n('labelNotifyThisUpdated')"/>
+      <label class="ml-1" :key="value" v-for="([text, value]) of [
+        [i18n('genericOn'), '1'],
+        [i18n('genericOff'), '0'],
+        [i18n('genericUseGlobal'), ''],
+      ]"><!-- make sure to place the input and span on one line with a space between -->
+        <input type="radio" :value="value" v-model="config.notifyUpdates"> <span v-text="text"/>
```
gera2ld: FYI there is another way to force add a space:

```html
{{' '}}
```

I guess this is a common case both in Vue template and React JSX.

PR #824, opened 2020-01-22
Fix webdav

### LINKED ISSUE #823: WebDAV Sync button is disabled
### What is the problem?
WebDAV Sync button is disabled

I'm not able to test for successful connection to my local webdav

Sync button remains disabled even after filling all the info (ip, user, pass)
I tried using cadaver on a WSL session from another computer and it connected without issues

### Environment
- Browser: Firefox
- Browser version: 72.0.1
- Violentmonkey version: v2.12.4
- OS: Windows 10

tophf:
Fixes #823: `http://` and `/` will be added to WebDAV URLs as needed.

### DIFF
--- a/src/options/views/tab-settings/vm-sync.vue
+++ b/src/options/views/tab-settings/vm-sync.vue
@@ -71,7 +71,7 @@
 </template>
 
 <script>
-import { sendCmd } from '#/common';
+import { noop, sendCmd } from '#/common';
 import options from '#/common/options';
 import SettingCheck from '#/common/ui/setting-check';
 import hookSetting from '#/common/hook-setting';
@@ -144,7 +144,7 @@ export default {
   },
   methods: {
     onSaveUserConfig() {
-      sendCmd('SyncSetConfig', this.state.userConfig);
+      sendCmd('SyncSetConfig', this.state.userConfig).catch(noop);
     },
     onSyncChange(e) {
       const { value } = e.target;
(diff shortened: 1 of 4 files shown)


src/options/views/tab-settings/vm-sync.vue
```diff
@@ -20,29 +20,20 @@
     <p v-if="state" class="mt-1" v-text="state.message"></p>
     <fieldset class="mt-1" v-if="state && state.authType === 'password'">
       <div class="sync-server-url">
-        <label v-text="i18n('labelSyncServerUrl')"></label>
-        <input
-          type="text"
-          v-model="state.userConfig.serverUrl"
-          :disabled="!state.canAuthorize || state.userConfig.anonymous"
-        />
+        <!-- make sure to place the label and span on one line with a space between -->
```
gera2ld: The space should be added in `_locale/en` since not all languages need this. (e.g. CJK)

tophf: Putting this in the locale files is very fragile as it depends on the good will and attention to details by both the translators and whoever merges the locales, currently you. If this space is soooo bad in CJK, I'll make it a `computed` thing in Vue, okay?

gera2ld: Actually there is already a space in the locale files, but not visible due to `white-space: normal`.

> If this space is soooo bad in CJK, I'll make it a `computed` thing in Vue, okay?

Sounds good, but how can it be `computed`?

I think a simple solution is to remove colons in all locales and add spaces between labels and form controls. Just looked around in my installed extensions and see no colons.

tophf: Oh, I see.
I've removed this commit and we'll deal with that in a separate PR.
Someday.

PR #827, opened 2020-01-22
use Proxy for sandboxing

tophf:
Benefits:

* faster to initialize: ~10ms delta (45 -> 35) for 10 scripts on my fast desktop computer so the difference could be bigger on a slower/older laptop. 
* less memory used: 1MB delta (5 -> 4)
* proxy handler is preserved even after defineProperty

Collateral:

* fix: sandboxing of `window.onXXX` events so one script doesn't overwrite the value set in another one
* fix: unforgeable and readonly properties to mimic native behavior of `window`
* minor simplifications

~I'll test it during the next few days with more random popular scripts.~
Seems to be working fine.

You were against `Proxy` because it can't be polyfilled but I don't see why we would need that. Do you think you'll be using this repo to update the Maxthon version?

### DIFF
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -42,9 +42,8 @@ Object.assign(commands, {
     return data;
   },
   /** @return {Promise<Object>} */
-  async GetInjected(url, src) {
-    const { id: tabId } = src.tab || {};
-    if (src.frameId === 0) resetValueOpener(tabId);
+  async GetInjected(_, { url, tab, frameId }) {
+    if (frameId === 0) resetValueOpener(tab.id);
     const data = {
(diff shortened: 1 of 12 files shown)


tophf:
Seems ready.
As usual the forced pushes were mostly code cosmetics.

tophf:
Instead of forced pushes I'll be adding fixup commits (if needed) and will do a final forced push after this is reviewed.

gera2ld:
This is great.

> Do you think you'll be using this repo to update the Maxthon version?

I don't have much time doing this now, maybe I'll look for some other ways someday.

PR #829, opened 2020-01-23
Update messages.yml RUS

wvxwxvw:


gera2ld:
They are already translated in [Transifex](https://www.transifex.com/violentmonkey-1/violentmonkey-nex/viewstrings/#ru/messagesjson) and will be updated in a few days.

wvxwxvw:
OK. I will not bother you anymore.

tophf:
You can [use Transifex directly](https://www.transifex.com/violentmonkey/violentmonkey-nex/), though. I find it extremely convenient, and robust in the sense that it automatically checks for the presence of special placeholders and tags, there's also no need to deal with the pesky YAML syntax.

wvxwxvw:
Apparently you yourself do not translate anything, so you have this opinion. I don't think so. Instead of a single window with two translations, in which I can see everything at a glance, you offer me to use a buggy web muzzle with annoying animation. From which I can't even copy anything normally, since it (web muzzle) adds line translations to the copied one.

And advise that translator to learn the Russian language before taking up translations.

tophf:
I didn't mean to annoy you further (it seems very easy to do though), but just so you know: I've translated tons of stuff there.

tophf:
Actually, Transifex allows to download/upload the entire file so you can edit it offline, will this work for you? It'd be helpful if you could maintain the Russian language since you seem to be passionate about quality, which is a very good thing.

image

BTW I've just compared the translations and there are tons of differences, which is rather perplexing as some of those differences were translated 2 months ago. Looks like the transifex version wasn't synchronized to github for a long time. Hopefully, it'll be faster in the future.

wvxwxvw:
Well, Yes, I'm a little annoyed by useless changes, especially those related to the transition to anime sites.

There are many inconsistencies and errors in that translation. I thought that someone had already "fixed" it. But maybe it's just an old translation that came before me.

Translated own .yml in the .json and uploaded to Transifex.

PR #835, opened 2020-01-29
Update tooltip style in dark mode

gera2ld:
image

## Collateral changes

- Add a link to contributors list instead of personal email

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "gulp-uglify": "^3.0.2",
     "husky": "^4.0.10",
     "js-yaml": "^3.13.1",
-    "jsdom": "^16.0.0",
+    "jsdom": "^16.0.1",
     "plugin-error": "^1.0.0",
     "tape": "^4.13.0",
     "through2": "^3.0.1",
@@ -57,7 +57,7 @@
     "core-js": "^3.6.4",
     "tldjs": "^2.3.1",
     "vue": "^2.6.11",
-    "vueleton": "^1.0.3"
+    "vueleton": "^1.0.4"
   },
   "engines": {
     "node": ">=10"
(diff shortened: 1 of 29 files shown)


tophf:
Mmm, I really liked the black tooltip in the dark mode. You don't?

gera2ld:
> Mmm, I really liked the black tooltip in the dark mode. You don't?

The black tooltip in a dark background looks not so clear to me since the colors are similar. I can hardly tell where the arrow is and how large the box is.
So I need to lighten either the background color or the tooltip. After some investigation I took the style from Firefox, which looks good to me.

Snapshot from Firefox:
image

tophf:
The arrow doesn't seem important to me as  the tooltip is close to the target, but the important thing for those who use the dark mode is, AFAIK, the overall ambiance, which is now disrupted by the bright tooltip. Firefox is even brighter and hence worse in my opinion. The black tooltip looked really slick and cool.

tophf:
If you like the bright tooltip, maybe you could delay it so it doesn't flicker in the dark page on mouse movement:

```css
@keyframes fade-in {
  from { opacity: 0 }
  to { opacity: 1 }
}

body .vl-tooltip {
  animation: .25s fade-in .25s both;
}
```

tophf:
BTW in my Windows the standard tooltip is dark:

image

The fact that it appears suddenly nearby is enough to draw my attention so I'm perfectly fine with its color being similar to the surroundings.

gera2ld:
The light border makes it clearer. I'll see if I can keep the black background and add a light border.

tophf:
The simplest way to add a border is probably `box-shadow`

tophf:
Yay I like it but you wanted to enhance the arrow so maybe add the `border` there too?

gera2ld:
The arrow itself is built with `border`s so it's not easy to add another `border` for it. I'll leave it as is. Does it look good to you?

tophf:
Well, ideally I'd like the arrow to have a border. It can be handled separately though, I'm not a perfectionist.

gera2ld:
Maybe I'll find a perfect way someday.

PR #840, opened 2020-01-31
fix: value change notifications + broadcast only changes

### LINKED ISSUE #839: GM_addValueChangeListener works in an unexpected way
## What is the problem?

I'm not exactly sure, but it acts in a really weird way, which messes up my userscript

### How to reproduce it?

```
// ==UserScript==
// @name        New script 
// @match       *://*/*
// @grant       GM_setValue
// @grant       GM_getValue
// @grant       GM_addValueChangeListener
// @version     1.0
// @author      -
// ==/UserScript==

console.log("Running");

function updatetest(number) {
  var varname = "test" + number;
  console.log("(old)" + varname + " =", GM_getValue(varname));
  var newvalue = Date.now();
  GM_setValue(varname, newvalue);
  console.log("(new)" + varname + " =", newvalue);
}

function addvcl(number) {
  GM_addValueChangeListener("test" + number, function(name, oldValue, newValue, remote) {
      console.log("Updated value", name, oldValue, newValue, remote);
  });
}

for (var i = 0; i < 3; i++) {
  updatetest(i);
}

for (var i = 0; i < 3; i++) {
  addvcl(i);
}
```

Tampermonkey returns this:

```
Running
(old)test0 = undefined
(new)test0 = 1580436199438
(old)test1 = undefined
(new)test1 = 1580436199439
(old)test2 = undefined
(new)test2 = 1580436199439
```

The first run, it works fine. After that it works strangely:

```
Running
(old)test0 = 1580436041163
(new)test0 = 1580436102001
(old)test1 = undefined
(new)test1 = 1580436102001
(old)test2 = 1580436001255
(new)test2 = 1580436102002
Updated value test0 1580436102001 undefined true
Updated value test1 1580436102001 undefined true
Updated value test2 1580436102002 1580436001255 true
Updated value test0 undefined 1580436041163 true
Updated value test1 undefined undefined true
Updated value test2 1580436001255 1580436001255 true
Updated value test0 1580436041163 1580436041163 true
Updated value test1 undefined undefined true
Updated value test2 1580436001255 undefined true
```

I can confirm that no other tabs were running that userscript at the time. I can also confirm that Tampermonkey doesn't exhibit this behavior.

### Environment
- Browser: Chromium
- Browser version: 79.0.3945.130
- Violentmonkey version: Latest git (https://github.com/violentmonkey/violentmonkey/commit/448e3d163a7cdd2271e5a7b3dc3ce545c6be9347)
- OS: Linux

tophf:
Fixes #839.
The problem was in changedRemotely - it didn't check if the key is actually hooked.
There was also a null/undefined mix-up.

Collateral: reduces the amount of tab messages and their size by broadcasting only the changed values in response to GM_xxxValue, whereas previously the entire script's value storage was sent every time.

### DIFF
--- a/src/background/utils/values.js
+++ b/src/background/utils/values.js
@@ -1,12 +1,10 @@
 import { isEmpty, sendTabCmd } from '#/common';
-import {
-  forEachEntry, forEachKey, objectPick, objectSet,
-} from '#/common/object';
+import { forEachEntry, forEachKey, objectSet } from '#/common/object';
 import { getScript, getValueStoresByIds, dumpValueStores } from './db';
 import { commands } from './message';
 
 const openers = {}; // { scriptId: { tabId: { frameId: 1, ... }, ... } }
-let cache; // { scriptId: { key: [{ value, src }, ... ], ... } }
+let cache = {}; // { scriptId: { key: { last: value, tabId: { frameId: value } } } }
 let updateScheduled;
 
 Object.assign(commands, {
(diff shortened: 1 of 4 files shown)


tophf:
Rebased on master.

PR #842, opened 2020-01-31
Fix sandbox

tophf:
Problem:

Currently VMInitInjection is called after GetInjected response which happens after the page started running its inline scripts so the globals (and prototypes that we use) may be already compromised, or the page could steal contentId and abuse it, see the team chat for more info.

Fix:

* VMInitInjection will run immediately at document_start before any page scripts.

Collateral:

* sandboxing is refactored a bit so it doesn't use unsafe Object.defineProperty from the page.
* Proxy trap for ownKeys is fixed + Symbols are handled properly

### DIFF
--- a/scripts/webpack.conf.js
+++ b/scripts/webpack.conf.js
@@ -62,15 +62,24 @@ const modify = (extra, init) => modifyWebpackConfig(
   },
 );
 
+// avoid running webpack bootstrap in a potentially hacked environment
+// after documentElement was replaced which triggered reinjection of content scripts
+const skipReinjectionHeader = `if (window[Symbol.for('${INIT_FUNC_NAME}')] !== 1)`;
+const skipReinjectionConfig = (config, test) => config.plugins.push(
+  new WrapperWebpackPlugin({
+    header: skipReinjectionHeader,
+    ...test && { test },
+  }));
+
 module.exports = Promise.all([
-  modify(),
+  modify(null, config => skipReinjectionConfig(config, /^browser\.js$/)),
   modify({
     pages: {
       injected: {
         entry: './src/injected',
       },
     },
-  }),
+  }, skipReinjectionConfig),
   modify({
     pages: {
       'injected-web': {
@@ -81,8 +90,8 @@ module.exports = Promise.all([
     config.output.libraryTarget = 'commonjs2';
     config.plugins.push(
       new WrapperWebpackPlugin({
-        header: `\
-          window.${INIT_FUNC_NAME} = function () {
+        header: `${skipReinjectionHeader}
+          window[Symbol.for('${INIT_FUNC_NAME}')] = function () {
             var module = { exports: {} };
           `,
         footer: `
--- a/src/injected/content/inject.js
+++ b/src/injected/content/inject.js
@@ -5,9 +5,8 @@ import {
   INJECT_PAGE,
   browser,
 } from '#/common/consts';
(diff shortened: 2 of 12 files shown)


src/injected/content/inject.js
```diff
@@ -8,14 +8,16 @@ import {
...
 } from '../utils/helpers';
 import bridge from './bridge';
 
 // Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1408996
-const VMInitInjection = window[process.env.INIT_FUNC_NAME];
-delete window[process.env.INIT_FUNC_NAME];
+const VMInitInjection = window[Symbol.for(process.env.INIT_FUNC_NAME)];
+// To avoid running repeatedly due to new `document.documentElement`
+// (the symbol is undeletable so a userscript can't fool us on reinjection)
+defineProperty(window, Symbol.for(process.env.INIT_FUNC_NAME), { value: 1 });
```
gera2ld: What's the benefit of using a symbol instead of the string, especially when the symbol is created in a global list, i.e. userscripts can also get it by `Symbol.for`?

tophf: The benefit is that it's non-enumerable by default so I don't have to use Object.defineProperty which may be hacked by a content userscript on reinjection of documentElement.

tophf: To clarify: I would have to use a compromised Object.defineProperty in webpack.conf:

```js
      new WrapperWebpackPlugin({
        header: `if (window[Symbol.for('${INIT_FUNC_NAME}')]!==1)
          window[Symbol.for('${INIT_FUNC_NAME}')] = function () {
            var module = { exports: {} };
          `,
```

tophf: Another benefit is that it won't clash with user keys or symbols when those are used in the intended, standard fashion (as for the ability of userscripts to access it using Symbol.for, this won't help them because the symbol will non-configurable after first init).

gera2ld: I am convinced.

tophf: Actually, the **only** benefit is that it won't clash with user keys/symbols. This is an important benefit though.

scripts/webpack.conf.js
```diff
@@ -70,6 +70,13 @@ module.exports = Promise.all([
         entry: './src/injected',
       },
     },
+  }, (config) => {
+    config.plugins.push(
+      new WrapperWebpackPlugin({
+        header: `if(window[Symbol.for('${INIT_FUNC_NAME}')]!==1)`,
+        afterOptimization: true,
+      }),
```
gera2ld: Why is the condition added this way instead of writing in the source code?

tophf: This will prevent webpack bootstrap code from running in a potentially hacked environment.

tophf: I'd like to do the same in `browser.js` but I didn't know how to do it easily...

tophf: Yay, I found a way, see the [`fixup!` commit](https://github.com/violentmonkey/violentmonkey/pull/842/commits/1edc0ade7208971bb2d381ceebc33406c215847a): added `test: /^browser\.js$/` in the main config.

gera2ld: It seems that `afterOptimization: true` is not necessary?

tophf: You're right. It's a leftover form my experiments.

PR #850, opened 2020-02-02
feat: disable page sandbox automatically in some cases

tophf:
This is rather a proof-of-concept than a real PR, I'm not sure it's useful in its current form. The ultimate goal is to skip creating page sandbox on irrelevant URLs by converting matches/includes/blacklist into a set of rules for [Chrome's declarativeContent API](https://developer.chrome.com/extensions/declarativeContent#type-PageStateMatcher) or the standard content script matches for [Firefox's contentScripts API](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/contentScripts).

Currently the `page` mode sandbox is disabled automatically in these cases:

* case A (both conditions required):

  * default injection mode in VM settings is `content`;
  * all enabled scripts (regardless of where they run) either don't specify `@inject-into` or it's `content`.

* case B:

  * all enabled scripts (regardless of where they run) specify `@inject-into content`

The unavoidable problem with the automatic analysis is that it will fallback to `<all_urls>` in case of `@include` with a regexp which are often too complex to be analyzed so here are some simpler and more robust alternatives:

* add a textbox "Don't try page mode on URLs/patterns" in global settings, without regexps, so the patterns can be easily converted to chrome API or used directly in firefox API
* don't disable page sandbox at all - until someone reports a real problem

Thoughts?


tophf:
I know how to make this thing useful: I'll try to support optional instant injection mode (the real `document-start`) both in Chrome (via webRequest + sync XHR) and Firefox (UserScripts API).

tophf:
Not a pressing issue for now. Will revisit later. Maybe when tackling browser.userscripts API.

PR #856, opened 2020-02-04
fix: prevent data loss on exiting/reloading editor

tophf:
* Autosaves the editor state every minute and when the user attempts to a) leave the page by reloading it via F5 or b) actually leaves it by using the back button.
* Displays a standard browser confirmation in case `a` or showConfirmation in case `b`
  >Modifications are not saved!  
  >You can reopen the editor now and click the "save" button.
* Restores the recovery info on editing the same script
* Restores the "clean" state when the user undoes all changes manually so the UI is updated accordingly

P.S. Now I know why history.replaceState was used instead of pushState - moving back in history loses the edits.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -512,6 +512,13 @@ msgNewVersion:
 msgNoUpdate:
   description: Message shown when there is no new version of a script.
   message: No update found.
+msgNotSaved:
+  description: >-
+    Message shown when there are unsaved script modifications and the user clicks browser Back button,
+    which we cannot prevent due to technical reasons so we can only show this reminder.
+  message: |-
+    Modifications are not saved!
+    You can reopen the editor now and click the "save" button.
 msgSavedBlacklist:
   description: Message shown when blacklist are saved.
   message: Blacklist updated.
--- a/src/common/object.js
+++ b/src/common/object.js
@@ -79,3 +79,28 @@ export function forEachKey(func) {
 export function forEachValue(func) {
   if (this) Object.values(this).forEach(func);
 }
+
+// Needed for Firefox's browser.storage API which fails on Vue observables
+export function deepCopy(src) {
+  return src && (
+    Array.isArray(src) && src.map(deepCopy)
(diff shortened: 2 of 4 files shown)


tophf:
Another solution would be to decouple the editor from the options page into a separate page so `beforeunload` would work out of the box.

gera2ld:
> Another solution would be to decouple the editor from the options page into a separate page so `beforeunload` would work out of the box.

Sounds good. The only problem is, a new tab will be created every time user clicks editing scripts from the options page, which is bad experience to me, and since this is a multi-tab thing, we can hardly ensure which tab will be active after closing the editor tab, which might be confusing.

tophf:
> a new tab will be created every time user clicks editing scripts from the options page

Why? No such thing happens in Stylus extension. Links are opened n the current tab by default in the browsers.

gera2ld:
> Links are opened n the current tab by default in the browsers.

That's good to me. But how can we get back to the last position after closing the editor?

tophf:
We can save scrollTop in the storage and assign it back during initialization when it's present in the storage.
Another approach is using an iframe for the editor page so we don't have to restore anything.

gera2ld:
`iframe` does not support `beforeunload` AFAIK.
The `scrollTop` way might work but we have to deal with asynchronous loading which may be tricky.

Actually I have no idea why we need to keep data after exiting / reloading. I would expect to get the stored version in that case. For example, when I want to discard all my modifications, I can simply reload the page.

tophf:
beforeunload works in iframes, I've tested it before suggesting. There [was a bug](https://crbug.com/960360) in Chrome but it got fixed.

>Actually I have no idea why we need to keep data after exiting / reloading

We won't have to anymore if we make a separate page or an iframe because onbeforeunload will be sufficient. In this PR the data was stored only because it was impossible to prevent onpopstate.

tophf:
So I guess I'll try the iframe approach.
I'll probably end up with a messy integration of Vue-related stuff though :-)

gera2ld:
You are really efficient! 👍

PR #863, opened 2020-02-08
feat: inject earlier than document_start in Chrome

### LINKED ISSUE #828: Generate script code string in background
Currently the script's code and its requires are transferred a lot:

* `bg -> content (triage) -> page (build) -> content (inject)`
  (triage->build->inject is unnecessary for code/requires)

* `bg -> content (triage) -> page (build) -> content (send) -> bg (executeScript)`
  (only the last executeScript with code + sendMessage with meta is needed)

The extraneous transfers also cause garbage collection as each step requires internal JSON stringification (extension messaging in Chrome definitely works like that) or structured cloning (probably used for page<->content DOM messaging). The only exception is the direct handler invocation used for page-content communication in `content` injection mode.

Goal:

* background script should generate the injected code (or 99% of it) as it's almost entirely independent of injection site specific things like mode that's being determined dynamically
* executeScript can be used directly to load scripts that specify `@inject-mode content` (plus sendMessage with the window property id so the content script will make a sandbox and run it with the correct arguments).

Secondary goals:

* cache the generated code for like a minute so it can be reused, also limit the cache by size, and invalidate the contents on script updates/changes

* use Firefox API like browser.userScripts to deliver this code

* remove the declared injected-web.js content script and run it only when needed e.g. when there are scripts with `@inject-mode content` or when the page turned out to have a restrictive CSP.

This will also improve the security model of Violentmonkey by eliminating InjectScript command that theoretically can be abused by a very sophisticated hack that uses Spectre or Meltdown to compromise the tab's isolated world of our content script. Doesn't seem very realistic but it'd be nice to remove the theoretical possibility just the same.

### LINKED ISSUE #887: Requests on my.mail.ru fail
Test script:

```
// ==UserScript==
// @name        Mail.ru test 
// @match       *://my.mail.ru/*
// @grant       none
// @version     1.0
// @author      -
// ==/UserScript==

(function() {
    var url = "https://www.google.com/";
    
    console.log("Sending a request from", document.location.href, document.documentElement);
    GM_xmlhttpRequest({
        url: url,
        method: "HEAD",
        onload: function(resp) {
            console.log(document.documentElement, document.location.href, resp);
        }
    });
})();
```

I get the following result when run on https://my.mail.ru/:

```
Sending a request from https://my.mail.ru/ ...
Sending a request from https://my.mail.ru/static/myrb2.html ...
... "https://my.mail.ru/static/myrb2.html" [response object]
```

Using chrome's network inspector, it will show the following network request:

`chrome-extension://.../background/undefined`

Firefox doesn't show the first request, only the second (successful) one.

The issue appears under both Firefox and Chrome, using the latest git version (https://github.com/violentmonkey/violentmonkey/commit/33c468a03afe9780b5e8f79c4a9b2af5c7af7afd).

tophf:
A limited implementation of instant-inject (#420) for Chrome, only in main page (not in frames).

Also closes #828, closes #887.
 
In my tests it consistently ran userscripts at real document_start (before `<head>` appears) on fast-loading sites, but not on those that take more than a second to load. Well, this is a big improvement anyway that doesn't rely on a deprecated synchronous XHR used by Tampermonkey (their approach is usually reliable but spams console with the deprecation warnings).

Interestingly, super fast navigation like back/forward/reopen are actually more reliable here and I haven't seen it fail even once during the few days I've been testing this PR, whereas TM has problems in these cases.

```js
// ==UserScript==
// @name        test injection mode
// @match       *://*/*
// @grant       GM_notification
// @run-at      document-start
// @noframes
// ==/UserScript==
console.log([...document.all]);
if (!document.head) GM_notification(location.href, 'No <HEAD>');
```


tophf:
Added more stuff I've been working on recently, see the commit names. It'd be painful to separate them into different PR as they would have conflicts on slightly changed `import` statements and so on.

PR #868, opened 2020-02-10
feat: group .user.js in FF devtools

tophf:
Firefox lists .user.js among our own content scripts so a space at start will group them:

image

Another solution would be to place all userjs in a subfolder but that's less convenient as the list will be collapsed so an additional click would be required.


tophf:
Will include it in #863.

PR #871, opened 2020-02-12
fix: plugin.list() does not return a Promise

### LINKED ISSUE #870: Failed syncing: Google Drive
### What is the problem?
Failed syncing: Google Drive
### How to reproduce it?
我点了开始同步。当同步到最后时，提示“同步出错”。
DevTools 提示“TypeError: fe.list(...).then is not a function”

### Environment
- Browser: chrome
- Browser version:  81.0.4040.5 dev （64 位）
- Violentmonkey version: 2.12.6
- OS: windows 10 64位

tophf:
Fixes #870.
Didn't test it though.
@gera2ld, this particular PR is worth releasing a new version ASAP, I think.

### DIFF
--- a/src/background/plugin/index.js
+++ b/src/background/plugin/index.js
@@ -2,15 +2,27 @@ import { commands } from '../utils/message';
 import { getScripts } from '../utils/db';
 
 export const script = {
-  // Update an existing script identified by the provided id
-  /** @param {{ id, code, message, isNew, config, custom, props, update }} data */
+  /**
+   * Update an existing script identified by the provided id
+   * @param {{ id, code, message, isNew, config, custom, props, update }} data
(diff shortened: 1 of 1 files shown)


gera2ld:
I think this is not the expected fix. `plugin.list()` is supposed to return a Promise so that the source is extensible. The plugin system was designed to support sync service with the file system, which was mentioned in one of the issues. In that case `plugin.list()` reads from file system and is asynchronous.

gera2ld:
I'm curious why it was not found before. :smile:

tophf:
Updated.

gera2ld:
It was introduced in 8275184868136067ad59cfaf539826f25d9995c2, since v2.12.5.

tophf:
Yeah I know, it was me!

PR #879, opened 2020-02-16
update resources on script update/save and display errors

### LINKED ISSUE #876: Make a config that disables script cache entirely.
I am trying to implement a hot reload feature in user scripts, but as I could notice, all scripts from @require and resources from @resource are cached forever, unless I go manually to the user script and  update it's version.

I have control of the server that serves these external scripts and even when response HTTP header <code>Cache-Control</code> is set to <code>Cache-Control: no-cache</code> these scripts still won't be fetched again.

Tampermonkey has an option named <code>Externals -> Update Interval</code> that can be set to <code>Always</code>.

This feature request is just for development purposes, production scripts are always updated by it's @version tag.

tophf:
Fixes #876.

Main changes:

* always fetch from network and rely on browser caching  
  (but the installer page still uses cache for resiliency)
* force no-cache HTTP header on script update (both manual or automatic)
* display errors for resource fetches when saving in editor (modal message) or updating in dashboard or in a notification for an updated script

Minor changes:
* keep modal message on screen while it's hovered
* preserve line breaks in modal message + show the first line in bold font

Collateral:
* ESLint: allow more props in one-line literal objects.

  This one really has been driving me mad all this time because with air-bnb's limit at `4` I constantly have to break imports into three lines just because a fourth import was added. Ditto for object literals. Such a low limit would only make sense if it was based on line length or visual complexity (e.g. presence of value specifiers or sub-properties), but we mostly use { shorthand } syntax anyway.

### DIFF
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -473,14 +473,14 @@ function buildPathMap(script, base) {
 }
 
 /** @return {Promise<?string>} resolves to error text if `resourceCache` is absent */
-export async function fetchResources(script, fetchOptions, resourceCache) {
+export async function fetchResources(script, reqOptions, resourceCache) {
   const { custom: { pathMap }, meta } = script;
   const snatch = (url, type, validator) => {
     url = pathMap[url] || url;
     const contents = resourceCache?.[type][url];
     return contents != null && !validator
       ? storage[type].set(url, contents) && null
-      : storage[type].fetch(url, fetchOptions, validator).catch(err => err);
+      : storage[type].fetch(url, reqOptions, validator).catch(err => err);
   };
   const errors = await Promise.all([
     ...meta.require.map(url => snatch(url, 'require')),
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -1,18 +1,15 @@
-import {
-  i18n, getFullUrl, isRemote, getRnd4, sendCmd,
-} from '#/common';
+import { i18n, getFullUrl, isRemote, getRnd4, sendCmd, trueJoin } from '#/common';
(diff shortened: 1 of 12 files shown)


src/background/utils/db.js
```diff
@@ -433,7 +447,7 @@ export async function parseScript(src) {
   if (src.position) script.props.position = +src.position;
   buildPathMap(script, src.url);
   await saveScript(script, src.code);
-  fetchScriptResources(script, src);
+  fetchScriptResources(script);
```
gera2ld: Does this mean that cache will not be used for `@require`d resources?

tophf: Yep, the idea is to rely on browser caching. Do you think there are cases where it may be detrimental?

gera2ld: This is ideal if the resource has a long time cache. Other than that, this may cause problems for us with bad connections. Even a 304 request could fail if connection cannot be made.

Currently the confirm page allows users to retry loading dependencies by reloading the page. Once the dependencies are loaded, users will expect the installation to be successful. But if the background has to send another request due to different cache controls, the installation might fail.

I have suffered from such cases and that is why I added this cache.

tophf: Okay, I'll add error handling and fallback on VM's cache.

tophf: Wait, there should be no problem with the current code. It stores the dependencies in storage only on success so errors or delays on subsequent retries (if tracking is enabled) don't influence anything.

tophf: Ah, I understand now. There's a duplication of the process in the confimation page and fetchScriptResources.

PR #880, opened 2020-02-18
fix: `chrome` is cleared by injected/web

tophf:
Follow-up to df2e856ee6c03e4fae751d173293778fa8e76b3e to fix its bug in content mode.


tophf:
Er, lol, it's not needed as the code only runs in our own tabs.

PR #881, opened 2020-02-19
Update Japanese translation

ScratchBuild:


tophf:
This label was already translated on [Transifex](https://www.transifex.com/violentmonkey/violentmonkey-nex/). Is your translation better?

PR #884, opened 2020-02-20
fix: keep original Blob

### LINKED ISSUE #883: Blob requests don't work on Bing with Nano Adblocker
Userscript:

```
// ==UserScript==
// @name        Blob test
// @match       *://*/*
// @version     1.0
// @author      -
// ==/UserScript==

GM_xmlhttpRequest({
    url: "https://www.google.com/",
    method: "GET",
    responseType: "blob",
    onload: function(resp) {
        console.log(resp.response);
    }
});
```

It returns `cValue {}` on Bing, but a proper Blob on other websites. When disabling Nano Adblocker, it works. This is because Nano Adblocker replaces Blob.

That being said, Tampermonkey somehow works regardless of whether Nano Adblocker is enabled or not.

### Environment
- Browser: Chromium
- Browser version: 80.0.3987.87
- Violentmonkey version: Latest git (https://github.com/violentmonkey/violentmonkey/commit/ece0c09f78980006ec845add3044e7d9173d0b11)
- OS: Linux

gera2ld:
fixes #883

### DIFF
--- a/src/injected/utils/helpers.js
+++ b/src/injected/utils/helpers.js
@@ -3,7 +3,7 @@
 // eslint-disable-next-line no-restricted-properties
 export const {
   // types
-  Boolean, Error, Promise, Uint8Array,
+  Blob, Boolean, Error, Promise, Uint8Array,
   // props and methods
   atob, isFinite, setTimeout,
 } = global;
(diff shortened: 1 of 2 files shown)


src/injected/utils/helpers.js
```diff
@@ -3,7 +3,7 @@
 // eslint-disable-next-line no-restricted-properties
 export const {
   // types
-  Boolean, Error, Promise, Uint8Array,
+  Boolean, Error, Promise, Uint8Array, Blob,
```
tophf: And here I tried to sort them... lol

gera2ld: Updated

PR #902, opened 2020-03-07
feat: prettify JSON value, show parsing errors

tophf:
Error state is shown on the `Save` button in order not to frighten the user by repeated reddening of the huge textarea while they type. The button also gets disabled on error so it can't be pressed until the error is fixed.

image

The values are properly parsed as JSON now so the spaces are trimmed, for example. Previously `     "foo"   ` would save all this unnecessary space. There was also a trivial UI bug that an all-space value like `      ` wouldn't be recognized as an attempt to delete the value (it was properly deleted though internally so it vanished on reopening of the editor).

I've also "undeprecated" simple types in storage (`n`, `b`, `s`) because there should be no need to re-encode everything via JSON. Ideally, I want to store all values directly so numbers are stored as numbers, strings as strings, objects as objects, and so on as there's no need to store the type. The motivation is that there is a double or triple re-encoding going on on each access to db because most/many scripts still use JSON.stringify on objects in GM_setValue just in case. The only thing's stopping me is that this will preclude easy downgrades to a previous version of the extension and it makes me sad...

### DIFF
--- a/src/options/views/edit/values.vue
+++ b/src/options/views/edit/values.vue
@@ -32,7 +32,10 @@
       <div class="flex mb-1">
         <h4 class="flex-auto" v-text="i18n('labelEditValue')"></h4>
         <div>
-          <button v-text="i18n('editValueSave')" @click="onSave"></button>
+          <button v-text="i18n('editValueSave')" @click="onSave"
+                  :class="{'has-error': current.error}"
+                  :title="current.error"
+                  :disabled="current.error"/>
           <button v-text="i18n('editValueCancel')" @click="onCancel"></button>
         </div>
       </div>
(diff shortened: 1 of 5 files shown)


src/options/views/edit/values.vue
```diff
@@ -168,6 +180,19 @@ export default {
     onCancel() {
       this.current = null;
     },
+    onChange() {
+      this.reparseJson(this.current.value);
```
gera2ld: I'm not sure if this would work well when the value is very large. I'll do some tests later.

tophf: Fixed by disabling the automatic check once it exceeds 10ms.

PR #908, opened 2020-03-09
fix/simplify/speed-up sandbox

### LINKED ISSUE #916: GM_getResourceText breaks with surrogate pairs
### What is the problem?
```
// ==UserScript==
// @name              Resource Test
// @version           1.0.0
// @match             *
// @run-at            document-end
// @charset           UTF-8
// @resource        cedict https://raw.githubusercontent.com/twifty/ff-translate/master/cedict_1_0_ts_utf-8_mdbg.txt
// @grant           GM_getResourceText
// ==/UserScript==

const tmp = `㞗 㞗 [qiu2] /(dialect) penis/dick/
㞙 㞙 [niao4] /old variant of 尿[niao4]/
㞞 𪨊 [song2] /(coll.) semen/weak and incompetent/
㟁 岸 [an4] /variant of 岸[an4]/
㠀 岛 [dao3] /variant of 島|岛[dao3]/
㠯 以 [yi3] /old variant of 以[yi3]/`

var expected = tmp.split(/\r?\n/).filter(line => line[0] !== '#')
console.log(expected)

var cedict = GM_getResourceText("cedict").split(/\r?\n/).filter(line => line[0] !== '#')
cedict = cedict.slice(159, 165)
console.log(cedict)

expected = [...expected[2]]
cedict = [...cedict[2]]

console.log(expected, expected[2].length, expected[2].charCodeAt(0).toString(16), expected[2].charCodeAt(1).toString(16))
console.log(cedict,   cedict[2].length,   cedict[2].charCodeAt(0).toString(16),   cedict[3].charCodeAt(0).toString(16))
```

### How to reproduce it?
The example script fetches a resource which contains a Chinese character from the unicode supplementary plane (U+010000 → U+10FFFF) represented as the surrogate pair (\uD86A\uDE0A). I expect the string returned from `GM_getResourceText` to contain that surrogate pair, instead it has been transformed into the byte sequence (\u0AA8\uA81B).

### Environment
- Browser: firefox
- Browser version: 74
- Violentmonkey version:  2.12.7
- OS: Linux

I've had a quick look into the source:
https://github.com/violentmonkey/violentmonkey/blob/470a92d29d37e1aafe8f7dd66a0e6f999de40121/src/injected/utils/helpers.js#L42
This only accounts for a sequence of 3 bytes, but UTF8 encoding of the supplementary plane always produces 4 byte sequences.

tophf:
* added the missing methods to `unforgeables` and `readonlys`

* Fixed an exception when a userscript wanted a particular injection mode (in its meta or VM's global mode) but the mode was non-injectable.

* `unforgeables` is defined now in the wrapper's init to let JS natively handle subsequent attempts to override it
* `readonlys` is correctly depopulated on redefines
* `globalKeys` is used again to prevent userscripts from seeing page vars directly

* simplified stepping though code in debugger by removing the need for webpack import wrappers like `n.d(t, "d", ()=>s)` so stepping into a sandboxed function call like `console.log(document)` will require pressing F11 key only 21 times instead of 42.

* removed an unsafe call to regexp#test in isUnforgeableFrameIndex and fixed it so only int32 numbers are handled.
* inlined single-use imports in helpers.js

* Switched to TextDecoder so now GM_getResourceText will be ~15 times faster: 0.15s instead of 2.5s on 9MB text.
* Reverted to plain INIT_FUNC_NAME because Symbol.for function can be spoofed.
* fixes #916.
* more trivial fixes and speedups

### DIFF
--- a/scripts/sandbox-globals.html
+++ b/scripts/sandbox-globals.html
@@ -1,0 +1,52 @@
+<p>Run it both in Chrome and FF</p>
+<textarea id="results" style="width:30em; height: 90%" spellcheck="false"></textarea>
+
+<script>
+const el = document.getElementById('results');
+el.value = [
+  function boundMethods() {
+    const keys = [];
+    /* global globalThis */
+    for (const k in globalThis) {
+      if (k >= 'a' && k <= 'z') {
+        const { value } = Object.getOwnPropertyDescriptor(globalThis, k)
+        || Object.getOwnPropertyDescriptor(EventTarget.prototype, k)
+        || Object.getOwnPropertyDescriptor(Object.prototype, k)
+        || {};
+        if (typeof value === 'function') {
+          if (k === 'createImageBitmap' || k === 'fetch') {
+            keys.push(k);
+          } else {
+            try {
+              globalThis[k].call({});
+            } catch (e) {
+              if (/Illegal invocation|implement interface/.test(e)) {
+                keys.push(k);
+              }
+            }
+          }
+        }
+      }
+    }
+    return keys.sort().map(k => `'${k}',`).join('\n');
+  },
+  function unforgeables() {
+    return Object.entries(Object.getOwnPropertyDescriptors(window))
+    .filter(([, v]) => !v.configurable)
+    .map(([k]) => `'${k}',`)
+    .sort()
+    .join('\n');
+  },
+  function readonlys() {
+    return Object.entries(Object.getOwnPropertyDescriptors(window))
+    .filter(([k, v]) => k >= 'a' && k <= 'z'
+      && k !== 'webkitStorageInfo' // deprecated
+      && v.get && !v.set && v.configurable)
+    .map(([k]) => `'${k}',`)
+    .sort()
+    .join('\n');
+  },
+].map(fn => `// ${fn.name}\n${fn()}\n`).join('\n');
+el.focus();
+el.select();
+</script>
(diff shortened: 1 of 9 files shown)


tophf:
@gera2ld the build failure is caused by something on their server:

```
gpg: BAD signature from "Yarn Packaging <yarn@dan.cx>"
> GPG signature for this Yarn release is invalid! This is BAD and may mean the release has been tampered with. It is strongly recommended that you report this to the Yarn developers.
The command "curl -o- -L https://yarnpkg.com/install.sh | bash" failed and exited with 1 during .
```

gera2ld:
> simplified stepping though code in debugger by removing the need for webpack import wrappers like `n.d(t, "d", ()=>s)` so stepping into a sandboxed function call like `console.log(document)` will require pressing F11 key only 21 times instead of 42.

Just curious, where should I put the `debugger` to reproduce this?

tophf:
Set a breakpoint on `console.log(document)` and when it hits, step through by pressing F11.

tophf:
Hmm, I have an idea: now that `globalKeys` are back I can probably get rid of `deleted` and maybe even switch to `Map` as its lookup should be faster than both Object::has or Array::includes. Gonna try that.

tophf:
Done.

tophf:
Switched to an array+join instead of progressive concatenation in GM_getResourceText so now it's ~2.5 times faster: 1.1sec to decode 9MB text instead of 2.7sec.

tophf:
* Switched to TextDecoder so now GM_getResourceText will be ~15 times faster: 0.15 s instead of 2.5 s on 9MB text.
* Reverted to plain INIT_FUNC_NAME because Symbol.for function can be spoofed.

PR #910, opened 2020-03-10
speedup: gmxhr with very large responses

tophf:
Turns out buffer2string is unbelievably slow with large payloads: it takes several seconds to encode/transfer/decode ~60MB video on a 4GHz i7 CPU, which is just ridiculous.

* Blobs are back for non-incognito tabs, and they're essentially *instantaneous*
* blob/buffer is sent into the page realm directly as DOM messaging uses structured cloning
* `responseText` will be sent just once (like `response`) and remembered in web/*
* `responseText` that's equal to `response` will be deduplicated internally in the message
* `responseHeaders` will be sent only when changed
* removed try/catch in background's httpRequest because handleCommandMessage catches all exceptions anyway and reports to the caller

image

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -5,7 +5,6 @@ import { forEachEntry, objectPick } from '#/common/object';
 import ua from '#/common/ua';
 import cache from './cache';
 import { isUserScript, parseMeta } from './script';
-import { getScriptById } from './db';
 import { extensionRoot } from './init';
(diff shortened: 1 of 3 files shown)


tophf:
Also enabled `concatenateModules` in webpack config so `injected-web.js` is reduced from 20.3k to 19k, `injected.js` from 10.8k to 10.3k, which is ~5%. Other files aren't that impressive.

PR #912, opened 2020-03-14
type checks and jsdoc

thesummit:


.gitignore
```diff
@@ -2,3 +2,4 @@
 node_modules/
 *.log
 /stats.json
+.idea/
```
tophf: I think you should use your own global git exclusion config - as you can see .gitignore here doesn't list any IDE artifacts.

src/common/download.js
```diff
@@ -11,6 +16,11 @@ export function downloadUrl(url, name, callback) {
   }, 3000);
 }
 
+/**
+ * @param {object} blob
```
tophf: While not a mistake but we already use `Object` so it should be used everywhere consistency.

tophf: Also, it should be `Blob`, not `Object`.

PR #931, opened 2020-03-24
Fix window and globals

gera2ld:
Fixed a few bugs:

- `window` is supposed to be the wrapper when there are `@grant`s, but it's `unsafeWindow` now.

  ```js
  > console.log(this, window, unsafeWindow)
  < Proxy {...}  Window {...}  Window {...}
  // But the expected result is
  < Proxy {...}  Proxy {...}  Window {...}
  ```

- The wrapper is a Proxy with a target object of `local`. When calling `Object.keys(wrapper)`, a `TypeError` is thrown:
  ```js
  > console.log(Object.keys(wrapper));
  < Uncaught TypeError: 'getOwnPropertyDescriptor' on proxy: trap returned descriptor for property 'top' that is incompatible with the existing property in the proxy target
      at Function.keys (<anonymous>)
  ```

  This is caused by violation of [invariants](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/getOwnPropertyDescriptor#Invariants):
  > A property cannot be reported as non-configurable, if it does not exists as an own property of the target object or if it exists as a configurable own property of the target object.

  The reason is that some non-configurable properties returned in `ownKeys()` does not exist in `local` (aka. the target object).

- `window` should be mapped to `wrapper` at initialization stage, so that we don't get confused by values set by user.
  For example, `self` is a configurable property whose initial value is `window`, which should be mapped to `wrapper`. However `self` can be assigned to `unsafeWindow` by users, in that case we should not map it to `wrapper` any more.

  ```js
  > console.log(window.self === window);
  < true
  > window.self = unsafeWindow;
  > console.log(window.self === unsafeWindow);
  < true
  ```

### DIFF
--- a/src/injected/web/gm-wrapper.js
+++ b/src/injected/web/gm-wrapper.js
@@ -247,9 +247,6 @@ function makeGlobalWrapper(local) {
   readonlys.delete = setDelete;
   readonlys.has = setHas;
   local.has = hasOwnProperty;
-  for (const [name, desc] of unforgeables) {
-    defineProperty(local, name, desc);
-  }
   if (bridge.isFirefox) {
     // Firefox returns [object Object] so jQuery libs see our `window` proxy as a plain
     // object and try to clone its recursive properties like `self` and `window`.
@@ -282,7 +279,16 @@ function makeGlobalWrapper(local) {
     getOwnPropertyDescriptor(_, name) {
       const ownDesc = describeProperty(local, name);
       const desc = ownDesc || globals.has(name) && describeProperty(global, name);
-      if (desc && desc.value === window) desc.value = wrapper;
+      if (!desc) return;
+      if (desc.value === window) desc.value = wrapper;
+      // preventing spec violation by duplicating ~10 props like NaN, Infinity, etc.
+      if (!ownDesc && !desc.configurable) {
+        const { get } = desc;
+        if (typeof get === 'function') {
+          desc.get = (...args) => global::get(...args);
+        }
+        defineProperty(local, name, mapWindow(desc));
+      }
       return desc;
     },
     has(_, name) {
@@ -305,6 +311,16 @@ function makeGlobalWrapper(local) {
       return true;
     },
   });
+  for (const [name, desc] of unforgeables) {
+    defineProperty(local, name, mapWindow(desc));
+  }
+  function mapWindow(desc) {
+    if (desc && desc.value === window) {
+      desc = assign({}, desc);
+      desc.value = wrapper;
+    }
+    return desc;
+  }
   function resolveProp(name) {
     let value = boundMethods.get(name);
     const canCopy = value || inheritedKeys.has(name) || globals.has(name);


src/injected/web/gm-wrapper.js
```diff
@@ -305,15 +302,38 @@ function makeGlobalWrapper(local) {
       return true;
     },
   });
+  const mapWindow = desc => {
+    if (desc && (desc.value === global || desc.value === window)) {
+      desc = {
+        ...desc,
```
tophf: Babel will be using unsafe calls to Object.xxx methods so it's safer to use `assign()` instead of `...`

src/injected/web/gm-wrapper.js
```diff
@@ -305,15 +302,38 @@ function makeGlobalWrapper(local) {
...
+  }
+  // non-configurable properties in `global` should also be defined in `local` (target object)
+  // Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/getOwnPropertyDescriptor#Invariants
+  //
+  // > A property cannot be reported as non-configurable,
+  // > if it does not exists as an own property of the target object
+  // > or if it exists as a configurable own property of the target object.
+  for (const name of globals) {
+    if (!unforgeables.has(name)) {
+      const desc = describeProperty(global, name);
```
tophf: This will be slow. I did that initially but it increases the startup time of page sandbox several times from 10-15ms to up to 50ms on my machine so it'll be slower for a lot of people.

We were checking the configurability on individual properties previously before my tweaks so you can just restore that (in `getOwnPropertyDescriptor()`).

src/injected/web/gm-wrapper.js
gera2ld: done

src/injected/web/gm-wrapper.js
gera2ld: This has to be done if we are using *Proxy*. See the second section described above.

tophf: We weren't enumerating the entire globals in 2.12.7 and Proxy worked correctly.

tophf: I see the source of the bug is my new code:
```js
if (desc && desc.value === window) desc.value = wrapper;
```
Turns out I can't change the descriptor's value - it should be set on `local` initially in this loop:
```js
  for (const [name, desc] of unforgeables) {
    defineProperty(local, name, desc);
  }
```
Gonna poke around a little more...

gera2ld: Ah I see, you are right.

tophf: Here's a quick/dirty fix that illustrates a possible solution which I didn't test a lot though so see if you can come up with a better solution: https://puu.sh/Fomps/02f18e5873.txt

tophf:
Seems fine!

Issue #374, opened 2018-03-18
Use Content scripts API


kekkc:
### What is the problem?
Current implementation of tabs.executeScript() causes many problems, e.g. no proper run-at document-start support.

Any plans to use the new, preferred Content scripts API? 
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_scripts
https://bugzilla.mozilla.org/show_bug.cgi?id=1332273

gera2ld:
I don't think the new API will make a difference to Violentmonkey, and we are not using `tabs.executeScript`.

losnappas:
Doesn't this circumvent the Firefox CSP issues, as well as allow injecting even when JS is disabled? It would be real sweet to get that to work. This would close a large number of current issues.

gera2ld:
This will cause other issues and break scripts using `unsafeWindow`. I think that's a larger number of issues.

losnappas:
That is probably true. Then, doesn't first running a content script like `const unsafeWindow = window.eval('window')` solve all problems? Have you experimented with `tabs.executeScript` or the content scripts API yet? Any plans?

I've seen more and more websites use a pretty strict CSP which completely wrecks my userscripts. I can see it as being an uptrend, as well.

gera2ld:
It's a bit complicated. In Firefox, `window` in a context script is different from `unsafeWindow`. Similarly, all global objects have a different copy, e.g. `window.Array` and `unsafeWindow.Array`. Operating such objects from a different environment causes DOM Exceptions in Firefox.

p.s. I have tried to solve this by `browser.tabs.executeScript` in v2.9.2 and reverted in v2.9.3.

losnappas:
Okay, it's a tough cookie to crack. Please consider including `executeScript` via a new `@rule` (or whatever) to deal with CSP. That would be super cool.

tophf:
Closing in favor of the newer userScripts API, see #604.

PR #941, opened 2020-03-29
feat: use browser.contentScripts API

tophf:
The complementary feature to early injection in Chrome, this one is using Firefox's browser.contentScripts API to pass the script data directly and thus implement the _real_ `document_start` for page mode scripts in Firefox.

Content mode seems to require more trickery (maybe rewriting `InjectionFeedback`) so I don't want to do it in this simple PR. The content mode gets somewhat faster anyway because the data is received earlier now.

Related: #374, #420.


tophf:
Found some bugs, investigating, will close for now.

tophf:
Fixed them.

PR #943, opened 2020-03-31
Use contentScripts API in FF + various fixes/tweaks

tophf:
The complementary feature to early injection in Chrome, this one is using Firefox's browser.contentScripts API to pass the script data directly and thus implement the _real_ `document_start` for page mode scripts in Firefox.

Content mode seems to require more trickery (maybe rewriting `InjectionFeedback`) so I don't want to do it in this simple PR. The content mode gets somewhat faster anyway because the data is received earlier now.

Related: #374, #420.

### Additional fixes:

* sourceURL can't use normal space for grouping
* FF bug workaround for sourceURL

### DIFF
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -3,6 +3,7 @@ import { TIMEOUT_24HOURS, TIMEOUT_MAX } from '#/common/consts';
 import ua from '#/common/ua';
 import * as sync from './sync';
 import { commands } from './utils';
+import cache from './utils/cache';
 import { getData, checkRemove } from './utils/db';
 import { setBadge } from './utils/icon';
 import { initialize } from './utils/init';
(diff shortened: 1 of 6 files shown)


tophf:
@gera2ld, do you see any obvious problems here?

gera2ld:
Nope. Do you think it's necessary to create a develop branch for such experimental features? Maybe we should keep master stable so that I can release new versions more frequently.

tophf:
From my experience with Stylus, a beta extension in the store is more helpful. As for a separate branch I don't see how it'd help us because no one will test it just like no one (except qsniyg) tests the master branch. Maybe at least upload the zip from the last build job for more people to sideload it and test?

tophf:
FWIW, I've tested this branch in FF and Chrome with most popular userscripts on greasyfork and it seems to work. It's not like this PR changes anything substantially.

tophf:
Okay, merged based on implicit okay in https://github.com/violentmonkey/violentmonkey/pull/943#issuecomment-607597819 :-)

PR #964, opened 2020-04-13
Add French translation messages

clement-escolano:


gera2ld:
Thanks @clement-escolano !

I'll keep this open until we can sync translations from repo to Transifex so that we can keep them consistent. Currently only Transifex to repo is supported.
Before that you can still update the translations from [Transifex](https://www.transifex.com/projects/p/violentmonkey-nex/resource/messagesjson/).

tophf:
Do French language rules require spaces before `:` when used in the internet and computer software? It looks so weird especially if `:` gets wrapped onto the next line due to width constraints. Maybe you could use the [non-breaking thin space](https://en.wikipedia.org/wiki/Thin_space)?

clement-escolano:
> Thanks @clement-escolano !
> 
> I'll keep this open until we can sync translations from repo to Transifex so that we can keep them consistent. Currently only Transifex to repo is supported.
> Before that you can still update the translations from [Transifex](https://www.transifex.com/projects/p/violentmonkey-nex/resource/messagesjson/).

I will port this pull request to transifex then :-)

> Do French language rules require spaces before `:` when used in the internet and computer software? It looks so weird especially if `:` gets wrapped onto the next line due to width constraints. Maybe you could use the [non-breaking thin space](https://en.wikipedia.org/wiki/Thin_space)?

Indeed this is required. And you are even more correct: it requires a non-breaking space (not a thin one however). The [Wikipedia page about French punctuation](https://fr.wikipedia.org/wiki/Ponctuation#Signes_de_ponctuation_occidentaux) explains it. I will update the translations accordingly.

clement-escolano:
@gera2ld @tophf I filled the French translations in Transifex. Can you see it on your side?

tophf:
It'll be merged automatically on schedule.

tophf:
The scheduled update was merged in #977.

PR #965, opened 2020-04-14
Update Japanese translation

ScratchBuild:


tophf:
Quoting https://github.com/violentmonkey/violentmonkey/pull/964#issuecomment-612838961

>I'll keep this open until we can sync translations from repo to Transifex so that we can keep them consistent. Currently only Transifex to repo is supported.
Before that you can still update the translations from [Transifex](https://www.transifex.com/projects/p/violentmonkey-nex/resource/messagesjson/).

PR #967, opened 2020-04-14
fix: set `this` to `window` for grant-none scripts

### LINKED ISSUE #966: Option to disable the injected code wrapping
I need the option to disable the code wrapping.

```js
(function(){with(this)((define,module,exports)=>{

...my code

})()}).call(VMink8zbw6t62bef)
```

I use third-party code, and this wrapping breaks the work. I have no problem with "User JavaScript and CSS", that injects the code without any changing.

`console.log(this);`
It prints `GM` object. I need `this` to be `window` object.

tophf:
Fixes #966 (only in `@grant none` script of course).

The actual execution of script code is moved into wrapGM to avoid transferring and unpacking of arguments. Any objections? Maybe wrapGM should be renamed?

### DIFF
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -1,5 +1,7 @@
 import { getUniqId } from '#/common';
-import { INJECT_CONTENT, INJECTABLE_TAB_URL_RE, METABLOCK_RE } from '#/common/consts';
+import {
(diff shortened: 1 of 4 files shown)


tophf:
Uhm, I went with a frankenstein compromise: `wrapGmAndRun`

PR #969, opened 2020-04-16
add privacy policy in About tab

### LINKED ISSUE #959: Action required by AMO reviewer
> Please add a privacy policy to this add-on that details which user data is being sent and to what services. The critical things to describe in the policy are how your extension collects, uses, stores, and shares or discloses information about people.If your add-on makes it apparent to websites that it is installed, this must also be mentioned. The policy should be about the extension only, not a copy of the website's privacy policy. It should also be the actual text, as opposed to a link to a privacy policy on a website. The privacy policy can be added in the add-on settings under “Manage Authors & License” on AMO.

gera2ld:
Add a link to privacy policy

### DIFF
--- a/src/_locales/cs/messages.yml
+++ b/src/_locales/cs/messages.yml
@@ -352,6 +352,9 @@ labelNotifyUpdatesGlobal:
 labelPopupSort:
   description: Label in the VM settings tab for script list sort order in popup
   message: ''
+labelPrivacyPolicy:
+  description: Label of link to privacy policy
+  message: ''
 labelRelated:
   description: Label of related links.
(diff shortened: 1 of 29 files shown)


tophf:
The border seems unnecessarily bright, maybe `border: 1px solid var(--fill-4)` ? And I would remove `max-height` because when the window is small there will be two scrollbars, when the window is very tall there will be empty space.

tophf:
Wait... the way I read #959 it says you need to add the privacy policy on AMO site, not in the extension.

gera2ld:
Yeah, it's an advice. I think I'll just add a link to the privacy policy.

Issue #971, opened 2020-04-16
Add support for Prettier


dutzi:
Hey,

This is a feature request for [Prettier](https://prettier.io/) support.

It would be awesome if once saved the document will be auto-formatted using Prettier.

That's it :)

tophf:
No, it's a big library, ~700kiB minified.
The entire js code in Violentmonkey extension is ~700kiB.

The built-in editor is only for trivial modifications.
Use an [external editor or IDE](https://violentmonkey.github.io/posts/how-to-edit-scripts-with-your-favorite-editor/) as your main editor.

PR #972, opened 2020-04-17
Add code auto formatting on save using Prettier

dutzi:
This PR implements a common behavior adopted by developers of having the document automatically formatted once hitting Save.

Hooking up to CM's save I was able to run prettier on the code right before actually saving the content. Pretty simple.

It's pretty rudimentary and probably requires more polishing. I come from React and have no Vue experience, so sorry if what I did doesn't make perfect sense.  

I expressed my concerns in the code, looks for `// TODO:` within the diff.

This addresses the [following isssue](https://github.com/violentmonkey/violentmonkey/issues/971).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "editor.formatOnSave": false
+}
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@babel/runtime": "^7.9.2",
     "codemirror": "^5.52.2",
     "core-js": "^3.6.4",
+    "prettier": "^2.0.4",
     "tldjs": "^2.3.1",
     "vue": "^2.6.11",
     "vueleton": "^1.0.4"
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -67,6 +67,8 @@ import 'codemirror/addon/search/searchcursor';
 import 'codemirror/addon/selection/active-line';
 import 'codemirror/keymap/sublime';
 import CodeMirror from 'codemirror';
+import prettier from 'prettier/standalone';
+import parserBabel from 'prettier/parser-babel';
 import Tooltip from 'vueleton/lib/tooltip/bundle';
 import ToggleButton from '#/common/ui/toggle-button';
 import { debounce } from '#/common';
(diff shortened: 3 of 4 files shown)


dutzi:
Still missing is:

- [ ] Run Prettier once the user clicks on the Save or Save & Close buttons (I tried doing it but didn't get too far)
- [ ] Fix [this nasty hack](https://github.com/violentmonkey/violentmonkey/pull/972/files#diff-0ae86d8642c359c02ee7e7e933a9aa29R124-R126)

tophf:
No, it's a big library, ~700kiB minified.
The entire js code in Violentmonkey extension is ~700kiB.

The built-in editor is only for trivial modifications.
Use an [external editor or IDE](https://violentmonkey.github.io/posts/how-to-edit-scripts-with-your-favorite-editor/) as your main editor.

dutzi:
I don't see why size matters as we are talking about a browser extension that gets loaded once.

dutzi:
What if it was loaded dynamically? Upon user hitting "Format". Would that suffice?

dutzi:
Plus VM's editor is very advanced, as it should be. Supports multiple cursors, implements Sublime Text-like keyboard shortcuts...

tophf:
>I don't see why size matters

With that kind of logic where do we stop? Some people already suggested shipping Monaco editor which is 10MB minified.

There are several reasons, the big size of Prettier being only one of them. The main reason is that Violentmonkey is a lightweight extension with a single purpose of running userscripts. Its built-in editor (CodeMirror) is only for trivial stuff. This is intentional. The preferred method of editing is an [external editor or IDE](https://violentmonkey.github.io/posts/how-to-edit-scripts-with-your-favorite-editor/).

As for those shortcuts you mention are provided by a small built-in CodeMirror addon, only basic stuff, it's nowhere near the level I'd call "advanced" if compared to a proper IDE or specialized editor.

>What if it was loaded dynamically?

That won't work. Running remote code in the extension context is unsafe so we would have to use a sandboxed iframe or relax the built-in `content_security_policy`, which we won't do, of course.

dutzi:
> Some people already suggested shipping Monaco editor

I was going to suggest that :)

It also depends where you draw the line for what is expected out of an editor nowadays. As I use it daily, to me it seems obvious that Prettier is included.

I noticed that the VM's repo isn't using it, which is unfortunate, I highly recommend it as it has two uses:

1. Formatting the code, obviously
2. Receiving quick feedback for syntax errors. When Prettier encounters one it breaks and keeps the document in its unformatted code, so when hitting save you get immediate feedback as to whether or not you code compiles.

tophf:
We don't need Prettier because we use ESLint, it's sufficient. This topic was already discussed multiple times by lots of people so no need to try to advocate here.

gera2ld:
Running remote code is definitely not allowed by all web stores.

As tophf said, Violentmonkey is a lightweight extension with a single purpose of running userscripts. So the editor is not our focus at any time.

I think a better idea is to communicate with a cloud IDE or editor that supports linting, formatting, building out of the box. The cloud IDE can be built as a PWA so it will work offline after first load.

PR #996, opened 2020-06-01
fix: switch page->auto mode on upgrade from pre-2.12.8

tophf:
Turns out most of existing users are still using the `page` injection mode instead of `auto` which is the default one since dea7de01defb3cfcada84593f0c2837ee5885569 in 2.12.2. It means we still don't help our longtime Firefox users who are affected by the Firefox bug that applies page CSP to our script elements.

This PR forcefully switches from `page` to `auto` when upgrading from 2.12.7 or older in Firefox.

### DIFF
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -1,7 +1,12 @@
-import { i18n, getFullUrl, isRemote, getRnd4, sendCmd, trueJoin } from '#/common';
-import { CMD_SCRIPT_ADD, CMD_SCRIPT_UPDATE, TIMEOUT_WEEK } from '#/common/consts';
+import {
+  compareVersion, i18n, getFullUrl, isRemote, getRnd4, sendCmd, trueJoin,
+} from '#/common';
+import {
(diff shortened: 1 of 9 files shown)


src/background/utils/db.js
```diff
@@ -158,6 +161,13 @@ preInitialize.push(async () => {
       return map;
     }, {}),
   });
+  // Switch defaultInjectInto from `page` to `auto` when upgrading VM2.12.7 or older
+  if (version !== lastVersion
+  && ua.isFirefox
+  && data.options?.defaultInjectInto === INJECT_PAGE
+  && `${lastVersion}`.replace(/\b\d\b/g, s => `0${s}`) <= '02.12.07') {
```
gera2ld, line 168: Can we use `compareVersion` here?

tophf: Done, thanks. I've also just added an indication of the failure and a clickable action to fix it. The indication is a red strike-through, intentionally weird to be noticeable (too weird?). You can try it on this github page when VM's default injection mode is `page`.

gera2ld:
The retrying behavior looks great.

PR #1014, opened 2020-07-07
fix: semver-compliant comparison of the prerelease chunk

### LINKED ISSUE #1013: [Feature request] Support for semver pre-releases in metadata
### What is the problem?
Currently only numbers separated by points can be compared with function compareVersion.
According to [semver specifications](https://semver.org), pre-releases may be denoted by appending a hyphen, leading e.g. to following order:
1.0.0 < 1.0.1-0 < 1.0.1-1 < 1.0.0

In compareVersion this leads to:
1.0.0 < 1.0.1 = 1.0.1 = 1.0.1

### How to reproduce it?
1. Upload a pre-release script with a version string according to semver specifications, e.g. 1.0.1-0
2. Upload a newer version of the script, with a newer preloaded (1.0.1-1) or a release version (1.0.1)
=> The newer version is not found as newer version

### Desired behavior
compareVersion supports version strings according to semver.

tophf:
Fixes #1013.

Spec: https://semver.org/#spec-item-11


tophf:
I think it's ready for a review.

PR #1018, opened 2020-07-11
feat: [editor] support autocomplete

orionlee:
Support basic autocomplete using `Ctrl-Space`.

The main logic comes from CodeMirror's auto complete: https://codemirror.net/demo/anywordhint.html

UI on violentmonkey's script editor:
image
image

### DIFF
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -494,6 +494,26 @@ $selectionDarkBg: rgba(73, 72, 62, .99);
   }
 }
 
+/* CodeMirror show-hints fix to work here */
+.CodeMirror-hints {
+  z-index: 9999;
+  /* consistent font with the rest, copied from .monospace-font rule from style/style.css */
+  font-family: "Fira Code", Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", Courier, monospace;
+}
+
+@media (prefers-color-scheme: dark) {
+  .CodeMirror-hints {
+    background: var(--bg);
+  }
+  .CodeMirror-hint {
+    color: var(--fg);
+  }
+  li.CodeMirror-hint-active {
+    background:white ;
+    color: #08f;
+  }
+}
+
 /* fix contenteditable selection color bug */
 .cm-s-eclipse .CodeMirror-line {
   ::selection {
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -507,8 +507,8 @@ $selectionDarkBg: rgba(73, 72, 62, .99);
     color: var(--fg);
   }
   li.CodeMirror-hint-active {
-    background:white ;
-    color: #08f;
+    background: white ;
+    color: var(--fg);
   }
 }
 
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -102,19 +102,12 @@ CodeMirror.commands.insertTab = cm => (
   cm.options.indentWithTabs ? insertTab(cm) : insertSoftTab(cm)
 );
 
-function autoHintWithFallback(cm, options) {
-  var result = null;
-  var hintFn = cm.getHelper(cm.getCursor(), 'hint');
-  if (hintFn) {
-    result = hintFn(cm, options);
-  }
-  // fallback to anyword if default returns nothing
-  if (!result || result.list.length < 1) {
-    result = CodeMirror.hint.anyword(cm, options);
-  }
-  return result;
+function autoHintWithFallback(cm, opts) {
+  const result = cm.getHelper(cm.getCursor(), 'hint')?.(cm, opts);
+  // fallback to anyword if default returns nothing (or no default)
+  return result?.list.length ? result : CodeMirror.hint.anyword(cm, opts);
 }
-CodeMirror.registerHelper('hint', 'autoHintWithFallback',  autoHintWithFallback);
+CodeMirror.registerHelper('hint', 'autoHintWithFallback', autoHintWithFallback);
 
 CodeMirror.commands.autocomplete = (cm) => {
   cm.showHint({ hint: CodeMirror.hint.autoHintWithFallback });
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -497,9 +497,6 @@ $selectionDarkBg: rgba(73, 72, 62, .99);
 /* CodeMirror show-hints fix to work here */
 .CodeMirror-hints {
   z-index: 9999;
-  /* consistent font with the rest, copied from .monospace-font rule from style/style.css */
-  font-family: "Fira Code", Menlo, Monaco, Consolas, "Liberation Mono"
-                , "Courier New", Courier, monospace;
 }
 
 @media (prefers-color-scheme: dark) {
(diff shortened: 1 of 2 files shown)


src/common/ui/code.vue
```diff
@@ -486,6 +494,27 @@ $selectionDarkBg: rgba(73, 72, 62, .99);
   }
 }
 
+/* CodeMirror show-hints fix to work here */
+.CodeMirror-hints {
+  z-index: 9999;
+  /* consistent font with the rest, copied from .monospace-font rule from style/style.css */
+  font-family: "Fira Code", Menlo, Monaco, Consolas, "Liberation Mono"
+                , "Courier New", Courier, monospace;
```
tophf, line 502: Instead of copying simply add `.CodeMirror-hints,` to the original rule.

src/common/ui/code.vue
```diff
@@ -486,6 +494,27 @@ $selectionDarkBg: rgba(73, 72, 62, .99);
...
+@media (prefers-color-scheme: dark) {
+  .CodeMirror-hints {
+    background: var(--bg);
+  }
+  .CodeMirror-hint {
+    color: var(--fg);
+  }
+  li.CodeMirror-hint-active {
+    background:white ;
+    color: #08f;
```
tophf, line 514: Dark themes don't use the white background color usually so this should be

```css
  li.CodeMirror-hint-active {
    background: var(--fill-6);
  }
```

src/common/ui/code.vue
```diff
@@ -66,6 +66,9 @@ import 'codemirror/addon/search/match-highlighter';
 import 'codemirror/addon/search/searchcursor';
 import 'codemirror/addon/selection/active-line';
 import 'codemirror/keymap/sublime';
+import 'codemirror/addon/hint/show-hint.css';
+import 'codemirror/addon/hint/show-hint';
+import 'codemirror/addon/hint/anyword-hint';
```
tophf, line 71: It would make sense to include `codemirror/addon/hint/javascript-hint` too.

src/common/ui/code.vue
orionlee: just to make sure: the background is for the selected autocomplete option. In this case, it corresponds the white background of selected `querySelector` in dark theme screen shot. 

Using `--fill-6`, a light grey against dark background make the highlighting not too clear:

image

src/common/ui/code.vue
orionlee: We could add it. But in my experience, the bundled `javascript-hint` is somewhat limited and could be confusing.

E.g., it has no troubles in providing proper selection while completing on  browser's globals, e.g., providing all function when completing `document.`, but it doesn't infer much more:

```javascript
  var el = document.querySelector('#some-id');
  // cannot auto complete, javascript-hint cannot tell el if an HTMLElement.
  el.
```

src/common/ui/code.vue
orionlee: It we add `.CodeMirror-hints` to the original rule in `style/style.css`, it will be overriden by bundled `codemirror/addon/hint/show-hint.css` with `font-family: monospace;`

We could add `!important` to the original rule, but I'm uncertain whether it might has other unintended consequence. Thoughts?

src/common/ui/code.vue
orionlee: One can try out `javascript-hint` in CodeMirror's online demo to see if it's worth adding:
  https://codemirror.net/demo/complete.html

src/common/ui/code.vue
tophf: The foreground color should be specified as `var(--fg)`

src/common/ui/code.vue
tophf: It provides completion for the standard JS methods and keywords so it makes sense and I don't understand the reluctance.

src/common/ui/code.vue
tophf: Let's simply use a specificity hack e.g. `.CodeMirror .CodeMirror-hints`

orionlee: I used the `.CodeMirror-hints.eclipse` instead, because hints UI elements are directly under `<body>`, outside `.CodeMirror`, or violentmonkey UI specific `.editor-code`

Other alternatives inlcude `body > .CodeMirror-hints` , `ul.CodeMirror-hints`, but they depend on `shot-hints` addon internals in what and where HTMLs are generated. Adding `.eclipse` has the advantage that it is the style violentmonkey explicitly uses.

HTMLs for the generated hints
```html
   ...
   <ul class="CodeMirror-hints eclipse" style="left: 62.4px; top: 438.4px;">...
</body>
```

src/common/ui/code.vue
orionlee: I changed selected foreground to `var(--fg)`. I feel using `var(--fg)` does not standout enough though.

####  Use `--fg` as foreground
```css
  li.CodeMirror-hint-active {
    background: white ;
    color: --var(fg);
  }
```
image


####  initially in the pull request
```css
  li.CodeMirror-hint-active {
    background: white ;
    color: #08f;
  }
```

image

src/common/ui/code.vue
orionlee: I just feel `javascript-hint` is a bit limited and could be counter-intuitive that it works sometimes but not other times from end users perspective.

E.g., it can complete `querySelector` when the code is `document.q` (`document` is a global), but it cannot complete when the code is `someElt.q` (`someElt` is an HTML element object).

orionlee: If we include `javascript-hint`, what'd be the logic? One proposal is to first use `javascript-hint`, if it returns nothing, fallback to `anyword-hint`.

tophf: Ah, I see. Yes, js+any sounds like it could solve the problem.

src/common/ui/code.vue
tophf: Now it's washed out and not readable at all.
Let's go with a simple inversion:

```css
background: var(--fg);
color: var(--bg);
```

src/common/ui/code.vue
```diff
@@ -98,6 +102,24 @@ CodeMirror.commands.insertTab = cm => (
   cm.options.indentWithTabs ? insertTab(cm) : insertSoftTab(cm)
 );
 
+function autoHintWithFallback(cm, options) {
```
tophf, line 105: Please use ESLint.

* `options` is already declared in the global scope so use `opts` or `cfg` or `config` and so on.
* we don't use `var`, we use `let`
* there are two spaces

It's also possible to rewrite the entire function using `?.`:
```js
CodeMirror.registerHelper('hint', 'autoHintWithFallback', (cm, opts) => {
  const result = cm.getHelper(cm.getCursor(), 'hint')?.(cm, opts);
  return result?.list.length ? result : CodeMirror.hint.anyword(cm, opts);
});
```

src/common/ui/code.vue
orionlee: Just committed the changes:

image

src/common/ui/code.vue
orionlee: Thanks! I didn't realize pre-push lint check was bypassed. And I didn't know `?.` is available.

tophf:
LGTM.

PR #1021, opened 2020-07-12
let script installed from http localhost support check for update

### LINKED ISSUE #1020: support check for update for userscript from local http server
### What is the problem?
When installing a userscript a local http server (the main workaround of installing local script on Firefox), the installed script doesn't support check for update. Users have to manually re-enter the script's URL in the Download URL of the script's settings.

### How to reproduce it?
1. Assume there is a local user script available at http://localhost:8000/lib/acme.user.js
2. Open the script in the browser
3. Click Confirm Installation to install the script.
4. In ViolentMonkey's script list, users can not check for update. (without manually entering Download URL in the settings)

The check for update is grayed out:
image

- Violentmonkey should be able to check for update of the installed userscript.

### Environment

orionlee:
Close #1020 

I'm a tad uncertain whether there is some unintended consequence, because the old logic explicitly blocks scripts from `http://localhost` from supporting check for update (by the check `isRemote()`).

https://github.com/violentmonkey/violentmonkey/blob/a466ca34233ef3f58aee59f1ccb6bfc8ba697263/src/common/index.js#L112-L114

### DIFF
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -455,7 +455,7 @@ export async function parseScript(src) {
   if (!meta.homepageURL && !script.custom.homepageURL && isRemote(src.from)) {
     script.custom.homepageURL = src.from;
   }
-  if (isRemote(src.url)) script.custom.lastInstallURL = src.url;
+  if (/^https?:\/\//.test(src.url)) script.custom.lastInstallURL = src.url;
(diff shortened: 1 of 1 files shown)


tophf:
Depends on https://github.com/violentmonkey/violentmonkey/issues/1020#issuecomment-657266805

PR #1022, opened 2020-07-13
feat: [editor] highlight html /css

### LINKED ISSUE #1019: editor: support html/css higglighting
It'd be great if the editor can highlight html/css in the userscript. E.g.,

image

Note: I have something working in my personal dev version (that generates the above screenshot). If there are interests, I could further polish them and create a pull request.

orionlee:
Closes #1019 

It supports:
- highlight html /css string template when the context can be deduced
  - `insertAdjacentHTML`, `GM_addStyle`, `innerHTML`, `outerHTML`
- arbitrary string template prefixed with `/* html */` and `/* css */` comment.

Example:
image

---
Note: If PR #1018 autocomplete and this one are accepted, it would be trivial to support autocomplete html/css keyword as well (simply importing `html-hint` and `css-hint` in `code.vue`).


src/common/ui/code.vue
```diff
@@ -52,7 +52,7 @@
 <script>
 import 'codemirror/lib/codemirror.css';
 import 'codemirror/theme/eclipse.css';
-import 'codemirror/mode/javascript/javascript';
+import '#/common/ui/code_extras/mode/javascript-mixed/javascript-mixed';
```
tophf, line 55: We usually group all `#` imports at the top.

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,304 @@
...
+  // eslint-disable-next-line no-undef
+  } else if (typeof define === 'function' && define.amd) { // AMD
+    // eslint-disable-next-line no-undef
+    define(['codemirror/lib/codemirror', 'codemirror/mode/xml/xml', 'codemirror/mode/javascript/javascript', 'codemirror/mode/css/css'], mod);
+  } else { // Plain browser env
+    // eslint-disable-next-line no-undef
+    mod(CodeMirror);
+  }
+}((CodeMirror) => {
+  function dbg() {
```
tophf, line 18: FWIW you can use process.env.DEBUG, otherwise an empty function looks like an accidental leftover.

```js
  function dbg(...args) {
    // eslint-disable-next-line no-console
    if (process.env.DEBUG) console.debug(...args);
  }
```

On the other hand, maybe it's time to remove all the `dbg` calls to cleanup the code (in at least one case it'll eliminate the need for an intermediate variable).

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,304 @@
...
+        const rules = [
+          // <current-context>, <match-criteria>, <next-context-if-matched>,
+          // <optional-side-effects-if-matched>,
+          // <optional-side-effects-not-matched>
+          //
+          // - side-effects-not-matched: if not specified, defaulted to reset local mode matching
+
+          // for pattern GM_addStyle(`css-string`);
+          new Rule('<start>', () => tokTyp === 'variable' && tokStr === 'GM_addStyle',
+            'css-1'),
```
tophf, line 132: This wrapping looks weird. Since there are many parameters it might make sense to use named props so the call becomes:

```js
new Rule({
  curContext: '<start>',
  nextContext: 'css-1',
  matchFn: () => tokTyp === 'variable' && tokStr === 'GM_addStyle',
}),
```

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,304 @@
...
+    const cssMatcher = (function () {
+      const cssMode = CodeMirror.getMode(config, { name: 'css' });
+
+      function maybeToken(stream, state, jsTokStyle) {
+        let tokStyle = STYLE_PASS;
+        const tokTyp = state.jsState.lastType;
+        const tokStr = stream.current();
+
+        // define the transition rules to enter local CSS mode;
+        const rules = [
```
tophf, line 123: This code will run lots of times, needlessly recreating the same object. You can create it one time at defineMode beginning.

Same for the other cases below.

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,304 @@
...
+        if (this.caseNotMatchedFn) {
+          this.caseNotMatchedFn();
+        } else { // default not matched logic: reset local mode matching
+          state.maybeLocalContext = null;
+        }
+        return false;
+      }
+    }
+
+    const cssMatcher = (function () {
```
tophf, line 114: I don't see the *physical* need for this encapsulation so I'd say `maybeToken` should be just `maybeCssToken` in the upper scope. I understand you copied the way it was structured in CodeMirror but does it warrant an extra scope?

src/common/ui/code.vue
tophf: Ah, and `code_extras` is using the wrong convention, it should be `-`, but I'd like to simplify the path e.g. maybe `#/common/ui/code/codemirror-js-mixed`

orionlee: done

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: - how do I make the extension codes running on Firefox see `process.env.DEBUG`? Do I start firefox from command line and add the `DEBUG` to the environment?

- I'd feel more comfortable with removing the `dbg` calls after the logic has been tired out more thoroughly, say, in some beta. They are quite useful in tracing/debugging.

tophf: AFAIK The usage depends on IDE or the way you run the build so refer to the respective documentation or try googling it.

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: Using named props would be clearer but the rule set would become more verbose, harder to glance.

```javascript
          new Rule({
            curContext: 'css-1',
            matchFn: () => tokTyp === '(' && tokStr === '(',
            nextContext: 'css-2'
          }),
          new Rule({
            curContext: 'css-2',
            matchFn: () => tokTyp === 'quasi', // if it's a string template
            nextContext: 'css-in',
            caseMatchedFn: () => prepareReparseStringTemplateInLocalMode(cssMode, stream, state)
          }),
          new Rule({
            curContext: 'css-in',
            matchFn: () => stream.peek() === '`', // if it hits ending backtick for string template
            nextContext: null, // then exiting local css mode
            caseMatchedFn: () => { tokStyle = exitLocalModeAndTokenEndingBacktick(stream, state); },
            caseNotMatchedFn: () => { tokStyle = tokenInLocalMode(stream, state) // else stay in local mode
          }),
```

Should I go ahead and make the changes?

tophf: Well, to you it may look overly verbose but to me (or other people unacquainted with this code) the named props variant looks more meaningful. It's also easier to maintain/extend. Performance won't suffer because the object will be created just once.

tophf: BTW you can drop the `fn` suffix to simplify the names as it's unnecessary.

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: One difficulty in moving the rules outside the function is that the logic defined in the rules requires access to various parameters/variables passed. With the rules inside the function, the lambdas can simply access them locally. 

If the rules are moved outside the function, the lambdas need other ways to bind/access. 

One way is to pass them via some an object as a context to run the lambdas, but it'd make the lambdas more verbose. E.g.,

```javascript
          new Rule('css-2', (ctx) => ctx.tokTyp === 'quasi', // if it's a string template
            'css-in',
            (ctx) => prepareReparseStringTemplateInLocalMode(cssMode, ctx.stream, ctx.state)),

          new Rule('css-in', () => ctx.stream.peek() === '`', // if it hits ending backtick for string template
            null, // then exiting local css mode
            (ctx) => { ctx.tokStyle = exitLocalModeAndTokenEndingBacktick(ctx.stream, state); },
            (ctx) => { ctx.tokStyle = tokenInLocalMode(ctx.stream, ctx.state); }), // else stay in local mode
```

Existing rules for reference:
```javascript
          new Rule('css-2', () => tokTyp === 'quasi', // if it's a string template
            'css-in',
            () => prepareReparseStringTemplateInLocalMode(cssMode, stream, state)),

          new Rule('css-in', () => stream.peek() === '`', // if it hits ending backtick for string template
            null, // then exiting local css mode
            () => { tokStyle = exitLocalModeAndTokenEndingBacktick(stream, state); },
            () => { tokStyle = tokenInLocalMode(stream, state); }), // else stay in local mode
```

Thoughts? Any other ideas?

tophf: I'd say pass this stuff in `state`.

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: assume we move the rules outside the function, I'd think the encapsulation would still serve some purpose (safety, but also readability), grouping things needed by css matching only. E.g.,

```javascript
const cssMatcher = (function() {
  const cssMode = ... ;
  const cssRules = ...; // assuming we move it outside maybeToken()

  function maybeToken() { ... } // that uses cssMode/rules
  return { maybeToken };
})();
```

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
tophf: Er, I mean your `ctx` variant is fine.

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
tophf: Well, keep it if you like it but use ES6 `(() => {`.

Anyway, what I dislike in this setup is that the function is declared separately and is exported in an object so when jumping to definition or quickly looking up its code at a call site the extra step becomes an unnecessary pain. It's also unclear to me why use the extra complexity if there's only one function - why not export the function itself without all these indirections?

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
tophf: Hmm, I'd probably even remove `tokStr` and just use `ctx.stream.current()` because not all rules need it...

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: I added `process.env` check, though I don't really understand how nodejs's `process.env` got simulated in web extension. 

I also added a global check `window.DEBUG_CM_JAVASCRIPT_MIXED` for now, so that I can turn on /off from devtools.

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: done

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: done

orionlee: done

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: > if there's only one function - why not export the function itself without all these indirections?
By that do you mean move the codes to a separate file and import the top level functions?

Anyway, I've removed the scope. The exposed structures are rather small in the end anyway.
```javascript
const cssMode
const cssRules
function maybeCssToken
// html counter part
```

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
tophf: Run node.js with a command line that sets the DEBUG variable - it's usually possible to set via some config in an IDE, see the documentation. Don't add a `window` check because it preserves this code in in the production build whereas with just process.env check all dbg code and calls are eliminated.

src/common/ui/code.vue
```diff
@@ -50,9 +50,9 @@
 </template>
 
 <script>
+import '#/common/ui/code_extras/mode/javascript-mixed/javascript-mixed';
```
tophf, line 53: `code_extras` is using the wrong convention, it should be `-`, but I'd like to simplify the path e.g. maybe `#/common/ui/code/codemirror-js-mixed`

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: `Run node.js with a command line that sets the DEBUG variable`
But violentmoneky, being a web extension, does not run in `node.js`. The process.env it got is empty - the `process` is some npm implementation for a a browser that returns empty `env`.
I confirmed it by starting firefox from command line with `DEBUG` added to the envrionment. The `env` remains empty.

tophf: You need to build the extension with this variable defined. The build process runs node.js internally but that's probably an unnecessary info so I don't know which IDE you are using but I suppose it should be easy to find a tutorial on setting up environment variables for any current IDE, as well as in case you build it using a command line.

src/common/ui/code.vue
orionlee: done

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,361 @@
...
+        }
+        return false;
+      }
+
+      // holds input parameters and return values for a rule execution
+      static createRunContext(stream, state, jsTokStyle) {
+        const ctx = {};
+
+        let _tokStr = null;
+        Object.defineProperties(ctx, {
```
tophf, line 114: Why bothering with read/write access? This is used only internally, it's not a public API, so there should be no need for this.

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,361 @@
...
+          stream: { value: stream, writable: false },
+          state: { value: state, writable: false },
+          jsTokStyle: { value: jsTokStyle, writable: false },
+          tokTyp: { get() { return state.jsState.lastType; } },
```
tophf, line 118: `tokenType`

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,361 @@
...
+          tokStr: {
```
tophf, line 119: `tokenStream`

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,361 @@
...
+            get() {
+              if (_tokStr === null) {
+                _tokStr = stream.current();
+              }
+              return _tokStr;
+            },
+          },
+
+          // holds the output of a rule execution - the only writable property
+          tokSyle: { writable: true },
```
tophf, line 129: `tokenStyle`

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    const cssMode = CodeMirror.getMode(config, { name: 'css' });
+
+    // define the transition rules to enter local CSS mode;
+    const cssRules = [
+      // for pattern GM_addStyle(`css-string`);
+      new Rule({
+        curContext: '<start>',
+        match: c => c.tokTyp === 'variable' && c.tokStr === 'GM_addStyle',
```
tophf, line 159: I understand the desire to make it compact but it's not a trivial parameter like e.g. `s` for a string or `c` for a character or `i` for index so `ctx` would be more appropriate especially since you're using it already in createRunContext.

src/common/ui/code.vue
orionlee: i just noticed in the same file, `import #/` lines are put at the end of the imports. Should they be moved to the top as well? 

https://github.com/violentmonkey/violentmonkey/blob/033282f44eccc5dd5cf8749704b4f56323eb8e0e/src/common/ui/code.vue#L69-L73

I poked around a few other `.vue` files. There do not appear to have a convention of placing `import #/` at the top of imports.

tophf: I've just checked and I see `#` being grouped at the top right after `vue` import in every file except one. I guess code.vue treats CodeMirror block as a framework which is why it's at the top. Well, I guess then let's move it at the end of the import block.

src/common/ui/code_extras/mode/javascript-mixed/javascript-mixed.js
orionlee: Ah, I see what you mean: webpack does the magic at build time.
The code is changed.

src/common/ui/codemirror-js-mixed/javascript-mixed.js
orionlee: It's primarily meant as both a defensive measure and documentation. The use of context makes it unclear what are meant to be inputs and what are meant to be outputs.

In the equivalent logic before using context to pass around input/output. It's something like the following:
```javascript
function maybeToken(stream, state) {
  let tokStyle = STYLE_PASS;
  const tokTyp = state.jsState.lastType;
```
It's clear what are inputs and what are outputs (and enforced by the engine). 

Defining read/write access in the context is to serve the same purpose, to prevent developers accidentally misuse them in the future.

tophf: I understand the general idea, I just don't see how it benefits us in this case. This code is small, self-contained, and is rather a finished product that hopefully won't see any changes so there's for so called defensive programming. As for the documentation, there's a better solution that doesn't incur a performance cost: TypeScript-compatible JSDoc comments.

orionlee:
done.

     On Wednesday, July 15, 2020, 10:46:17 AM PDT, tophf <notifications@github.com> wrote:  
 
 


@tophf commented on this pull request.

In src/common/ui/code.vue:
> @@ -50,9 +50,9 @@ </template>  <script>+import '#/common/ui/code_extras/mode/javascript-mixed/javascript-mixed';
I've just checked and I see # being grouped at the top right after vue import in every file except one. I guess code.vue treats CodeMirror block as a framework which is why it's at the top. Well, I guess then let's move it at the end of the import block.

tophf:
Just a reminder, I can handle all these nitpicky/stylistic changes myself after merging as they are trivial.

src/common/ui/codemirror-js-mixed/javascript-mixed.js
orionlee: How about replacing it with a class? Something like the following:

```javascript
    class RunContext {
      constructor(stream, state, jsTokStyle) {
        /**
         * @readonly
         */
        this.stream = stream;

        /**
         * @readonly
         */
        this.state = state;

        /**
         * @readonly
         */
        this.jsTokStyle = jsTokStyle;

        /**
         * The output of a rule execution - the only writable property
         */
        this.tokStyle = STYLE_PASS;
      }

      get tokTyp() { return this.state.jsState.lastType; }
    }
```

orionlee:
I've addressed the `Object.defineProperties` performance issue and various naming changes.

Additionally, I've slipped in html/css autocomplete support, given the base autocomplete PR #1018 has been merged.

From a release standpoint, we might want to consider to provide an option (user-facing or a backdoor) to use plain javascript mode in initial release, in case the new codes unexpectedly some unexpected errors for some users.

A minimal backdoor could be allowing allow switching back to plain javascript mode by setting a value in localStorage.
```javascript
    mode: window?.localStorage['editor.usePlainJavacript'] ? 'javascript' : 'javascript-mixed',
```

tophf:
There should be no need for an emergency switch because people can use an external editor easily.

Looks fine now except the import line, which I still want either at the end of the import block or right after another one from `/ui/` and a shorter directory name e.g. just `code`, but I can do it myself if you're fed up :-)

However, the real deal-breaker I see in devtools profiler is 33% worse performance compared to the built-in `javascript` mode on a big script. This must be alleviated to 5% or less on scripts with embedded grammar and almost 0% on those without. The primary cause is apparently `jsToken` function: why does it re-check all known tokens like `keyword` from the built-in js mode? If it's determined to be a valid js keyword then there's no way it could be a part of embedded css/html, AFAICT. I hope it's possible to limit the work only to `string-2` tokens contents.

src/common/ui/code.vue
```diff
@@ -69,7 +68,10 @@ import 'codemirror/keymap/sublime';
 import 'codemirror/addon/hint/show-hint.css';
 import 'codemirror/addon/hint/show-hint';
 import 'codemirror/addon/hint/javascript-hint';
+import 'codemirror/addon/hint/html-hint';
+import 'codemirror/addon/hint/css-hint';
```
tophf, line 72: We won't use html/css addons as they add ~10kB, too big for such a rarely used feature. People either use devtools to write CSS/HTML or the specialized editors.

tophf:
> highlight html /css string template when the context can be deduced

I wonder would it be viable to detect html in standalone strings as they seem to be very popular, even a simple regex like `/^\s*<[-a-z0-9]+(\s|\/?>)/i` might work.

orionlee:
Regarding performance issue: I'd look into it. Limiting matching to only some tokens could work.

Having said that, I think that as it is, the overhead comparing to *overall* performance is acceptable.
- empirically, I load and edit some moderately large scripts (~500 lines js + html/css, ~900 lines of plain js) and didn't really notice any lag on a laptop and an entry-level android tablet (1.4Ghz quad-core, 2GB ram) [*]. 
- looking at profile results, the overhead comparing to *overall* performance (Scripting + Rendering from devtools performance summary) is about 1.5% of pure overhead (`matchRule` - `tokenInLocalMode`, and another 1.5% for html/ css tokenization  (`tokenInLocalMode`) for scripts with quite some html/css.

[*] I try it on a mobile device by loading `javascript-mixed` in a basic CodeMirror demo web page. 
See https://gist.github.com/orionlee/d038eaabe6451023c5bd97812796c312

src/common/ui/code.vue
orionlee: It's just a nice to have but it's no big loss.

I'm curious on the reasons that the extra ~10k. Given all the files are served locally rather than network, are you concerned about the extra load / parse time or memory?

tophf: Violentmonkey's built-in editor is just a stop-gap measure to perform quick edits and it's meant to be as bare-bones as possible so only the more frequently used features are added.

tophf:
I've rechecked in production mode and indeed there's almost no difference. Previously I tried the dev mode. So, I guess there's urgent need to look into token scoping.

tophf:
I've rechecked in production mode and indeed there's almost no difference. Previously I tried the dev mode. So, I guess there's no urgent need to look into token scoping.

orionlee:
I tried short-circuiting local mode match: it shaved off a lot of calls, e.g, matching tokens such as `(` and `)`, but actual time reduction is not significant (< 1% of overall scripting time).

```javascript
    function jsToken(stream, state) {
......
      if (state.maybeLocalContext == null
        && !['variable', 'property', 'comment'].includes(tokenStyle)) {
        return tokenStyle;
      }

      // existing local mode code
      // match to see if it needs to switch to local html mode, return local mode style if applicable
      let maybeLocalStyle = maybeHtmlToken(stream, state, tokenStyle);
      if (maybeLocalStyle === STYLE_PASS) {

```

I'd also like to add a small regression test.

src/common/ui/code.vue
orionlee: okay. I've backed out the commit.

tophf:
No need to complicate the code for that 1%.

tophf:
1. Single-quoted or double-quoted strings don't work for me.

2. Would it be viable to detect html in standalone strings as they seem to be very popular? Even a simple regex like `/^\s*<[-a-z0-9]+(\s|\/?>)/i` might work.

orionlee:
Plain (single-quoted and double-quoted) strings were not in the scope. It'd require more logic (and some workaround if we use codemirror's html mode for it). I could take a stab to look into it.

```javascript
elt.insertAdjacentHTML('beforebegin', '<div class="chapterWords">this chapter has '
    + numWords + ' words (time: ' + countTime(numWords) + ')</div>');
```

- The logic need to detect all string are part of `insertAdjacentHTML` calls and should be parsed in HTML mode, while correctly with string concatenation, etc.
- if codemirror's html mode is used, there might be additional issue that it might incorrectly detect that the ending `</div>` has no matching beginning one.

Another possibility is to check *all* standalone strings, ignoring context and such. In a userscript, any string started with `<` is likley to be html. Is that what you mean?

orionlee:
I had some logic that matches any HTML-like string. 

There is some minor limitation (on both false positives and false negatives) but overall should be acceptable for userscript environment. I did some sanity profiling and it doesn't seem to have much impact (I was a tad concerned about running regexp against every string but it appears to be okay). 

Thoughts?

image

tophf:
Yeah, that's what I meant. This should cover like 99% of cases.

orionlee:
latest commits include support for Single-quoted or double-quoted strings that look like HTML.

I also extend the same support to string template for parity. E.g., it can auto-highlight, without outside of context (`insertAdjacentHTML`, `innerHTML`, etc.)
```javascript
let someHTML = `<span>foo</span>`;
```

Codewise, I considered removing the context rules for HTML (`insertAdjacentHTML`, `innerHTML`, etc.). I decided to keep them for now, as they are still needed to support a common case that cannot be auto-detected:

```javascript
// the html of the string template starts in the next line, it cannot be detected by the auto-detect heuristics.
someElt.insertAdjacentHTML('beforened', `
<div>
  bar
</div>`);
```

tophf:
Why isn't it possible to detect HTML in the last snippet? It'd be very nice because such formatting is convenient and I'd say common.

tophf:
Ahem, so I've measured performance again in devtools profiler using production builds and I kinda don't like it: the tokenizing task of CodeMirror (the second big task) previously took ~80 ms, now it takes ~115 ms (same for the yesterday's PR)... This task is performed whenever something is changed in the code so it must be as fast as possible. I'm probably overly obsessed with micro-performance but on the other hand given the goal of keeping our built-in editor primitive I would decline merging this PR...

tophf:
Would it be possible to limit all processing behind a check for `string-2` token type?

orionlee:
I feel the performance is acceptable: I tested the end-to-end performance on an entry level android tablet (see https://github.com/violentmonkey/violentmonkey/pull/1022#issuecomment-659576923), and could not perceive any lag in scrolling / editing, even for a large script ([AC-Baidu-...Redirect](https://greasyfork.org/en/scripts/14178-ac-baidu-%E9%87%8D%E5%AE%9A%E5%90%91%E4%BC%98%E5%8C%96%E7%99%BE%E5%BA%A6%E6%90%9C%E7%8B%97%E8%B0%B7%E6%AD%8C%E5%BF%85%E5%BA%94%E6%90%9C%E7%B4%A2-favicon-%E5%8F%8C%E5%88%97), ~2400 lines, the one on greasyfork). The setup is an approximation of what happens on a low-end machine, running on a dev build so I feel it's more than sufficient.

Having said that, I do understand the unease of the extra time in tokenization, and will do a few different experiments and report back. [*] 

Regarding only handling `string-2` style tokens., it would lead to some feature loss: handling only string templates (but not plain single/double quoted strings), can't handle css (requiring `GM_addStyle` of `variable` type context check, ), etc. I will look into it as well.

[*] I don't know what takes up the time, and will do some experiments to narrow it down. it could be:
- extra overhead of the lambda calls in using `rule.match(ctx)`
- regex calls against the text / comments of some rules
- creating 1-2 `RunContext` objects for every single token
- sheer numbers of rules
- the tiny overhead just all add up.

orionlee:
I have some tentative improvement (not yet in PR), in a separate branch: [javascript-mixed-v2.js](https://github.com/orionlee/violentmonkey/blob/e-h-html-css-performance/src/common/ui/codemirror-js-mixed/javascript-mixed-v2.js). I'd like to get some feedback before merging.

- the new version almost match existing ones, except it doesn't cover a corner case where html-like string detection fails, e.g., 
```javascript
  // failed to detect cases that some free-form text before html tag, even though it must be HTML
  // given it's insertAdjacentHTML
  elt.insertAdjacentHTML('beforeend', ` | <a href="...">`);
```

I saw some improvement with from chrome devtools on a dev build. 
BTW, how do you get measurement from a prod build in chrome devtools, where all the function names are scrambled?

Improvement incorporated:
- the biggest cut come from short-circuiting rule match (e.g., operators). See `javascript-mixed-short-circuited` below.
- reduced number of rules (looking for magic `insertAdjacentHTML`, `innerHTML`, `outerHTML`) also helped (see `js-mixed-skip-html-context-rules-shortcircuited` below).
- note: some amount of time increase is unavoidable due to the additional tokenization of css / html for scripts with lots of them. My estimate is that roughly half of the time increase come from the css/html tokenization for scripts that do lots of them. (Basically, time spent in `tokenInLocalMode()` and `tokenInLocalModePlainString()`) 

---
Potential further improvement:
- remove the magic `/*html*/` and `/*css*/` highlight hint: it'd allow skipping all `comments`, but it'd make highlighting css more restrictive for users who know about them, e.g.,
```
  // addCSS is user's own helper to add CSS
  // the current rules allows user to add /*css*/ hint before the string template to make the highlighting work.
  addCSS(`.div {
  padding: 1em;
}`);
```
- optimize the core rule match logic: the results show that the rule matching (that match nothing) takes up quite a bit of time. There might be rooms to improve, e.g. cut down the number of function calls by replacing `rule.match` lambdas with inline match.

---
**Time Measurement details**

Time measurement is obtained by repeatedly running the tokenization code standalone with [`test/.../runmode.html`](
https://github.com/orionlee/violentmonkey/blob/e-h-html-css-performance/test/common/ui/codemirror-js-mixed/runmode.html) against ~2400 line script (based on AC-Baidu-...Redirect script from greasyfork)


  mode   | median | numTokens | median/10k tokens | time details
 ------- | ------ | --------- | ------------------| ------------
javascript | 18 | 27344 | 6.6 | 16,16,16,17,17,17,17,17,17,17,18,18,19,19,32,51,52,52,59,83
javascript-mixed | 36 | 31732 | 11.3 | 34,35,35,35,35,35,36,36,36,36,36,37,37,37,38,48,49,56,58,58
javascript-mixed-v2 | 22 | 31732 | 6.9 | 21,21,21,21,21,22,22,22,22,22,22,22,23,23,24,29,30,30,31,31
*Other variations:* |
js-mixed-only-str | 25 | 30796 | 8.1 | 23,24,24,24,24,24,24,24,24,25,25,25,26,26,26,31,31,32,33,34
js-mixed-only-str-shortcircuited | 21 | 30796 | 6.8 | 19,19,20,20,20,20,20,20,20,21,21,21,22,22,26,29,31,32,38,41
javascript-mixed-shortcircuited | 25 | 31732 | 7.9 | 23,23,24,24,24,24,24,24,24,24,25,25,25,25,27,31,33,34,34,34
js-mixed-skip-html-context-rules-shortcircuited | 22 | 31724 | 6.9 | 20,21,21,21,21,22,22,22,22,22,22,22,22,22,23,29,29,30,30,32
js-mixed-skip-comments-n-html-context-rules-shortcircuited | 20 | 31243 | 6.4 | 19,19,19,19,20,20,20,20,20,20,20,21,21,22,22,23,25,25,26,27

tophf:
Regarding tokenization, I thought it's possible to look a few tokens back, at least one token?

Regarding performance, I'm using the light-weight `JavaScript profiler` in Chrome that can be opened from devtools menu or via Ctrl-Shift-P menu.

<details><summary>Instruction / screenshot</summary>

1. open a script editor in Violentmonkey
2. open the profiler
3. press Ctrl-E, Ctrl-R, and almost immediately Ctrl-E again
4. [one time] switch the profiler toolbar mode to `chart`

Remember the block duration and repeat to find the best/average times out of several runs.

image

</details>

branch | best | mean
-|-|-
master | 80 | 90
mixed | 115 | 125
mixed-v2 | 105 | 115

orionlee:
For further optimization, I think the next step I'd try is to replace rule-based execution with a set of mini `maybeTokenXxx` functions, similar to how the `javascript.js` switches between different tokens.

For measuring performance, to ensure we're talking about the same thing.
1. In the chart view above, how do you identify the block of time for tokenization?
2. I can find them in `Bottom Up` view, search for `token` and ensure the right token is selected by checking the source. 
image
3. Could you send me the script you loaded in the editor for profiling? Scripts with lots of HTML/CSS code would take some more time for tokenization that can't be much further reduced.

E.g., from my test using dev build on a script ~440 lines with quite a bit of HTML/CSS, ~40-50% of time increase are due to actual HTML/CSS tokenization (in profiler, search for time spent in `tokenInLocalMode` and `tokenInLocalModePlainString`). The number of tokens generated increased from 3000 to 3500. 

If the script you used share similar characterstics, the best possible mean time would be around `104ms` `(90 + (125-90) * 0.4)`

tophf:
Switch the profile into `chart` mode by clicking on `Heavy (bottom up)`, then you'll see something that's similar to my screenshot where I outlined the tokenization block.

The script was [my MPIV fork](https://github.com/tophf/mpiv/raw/master/script.user.js). I also tried the same script without CSS/HTML at the bottom of the script: mixed-v2 best 105 ms, avg 110 ms.

orionlee:
A question regarding performance measurement: the numbers you got are the `Kt` call (which is vue runtime's `flushCallbacks()` when comparing in with dev build) that involve call codemirror mode's `token()`. Right?

orionlee:
I have pushed a [`javascript-mixed-v3.js`](https://github.com/orionlee/violentmonkey/blob/e-h-html-css-performance/src/common/ui/codemirror-js-mixed/javascript-mixed-v3.js) that replaced the Rule matching with more efficient (but harder to change) of tiny maybeTokenXxx functions.

There are improvement. For the `Kt` time from javascript profiler (loading MPIV fork you referenced).

|  branch | best | mean | worst |
| --- | --- | --- | --- |
|  master | 70.4 | 76.9 | 84.6 |
|  mixed | 87.4 | 102.5 | 112.1 |
|  mixed-v3 | 80.2 | 88.3 | 98.0 |


I noticed something peculiar. 
Comparing to base javascript mode, the javascript-mixed variants somehow appear to spent much more time in the base javascript mode's token().

If I look at the breakdown from chrome dev tools > Performance tab.

- In plain javascript mode (master): the 'javascript.js'  token() is about 10ms.
- In javascript-mixed-v3: the mixed mode's token is about 27ms. But 23ms is spend in `javascript.js` token, doubling the time in plain javascript mode.
- I created a variation of the mixed mode that only forward the calls to underlying javascript mode [`js-mixed-v3-only-js.js`](https://github.com/orionlee/violentmonkey/blob/e-h-html-css-performance/src/common/ui/codemirror-js-mixed/js-mixed-v3-only-js.js#L172). The results are essentially the same. The profiler reports the time in `javascript.js` token() remain about 20ms.
- when I use the `runmode.html` helper to measure only the time spent in tokenization, I don't notice such discrepancy. The `javascript-mixed-v3`'s time is almost the same as plain `javascript`

<details>
<summary>Screenshot from performance tab</summary>

image

</details>

<details>
<summary>token time from runmode.html</summary>

 mode   | median | numTokens | median/10k tokens | time details
 ------- | ------ | --------- | ------------------| ------------
javascript | 19 | 35190 | 5.4 | 17,17,18,18,18,18,19,19,19,19,19,19,19,20,20,21,21,39,43,94
javascript-mixed | 38 | 35212 | 10.8 | 37,37,37,37,37,37,38,38,38,38,38,38,38,39,39,39,46,49,54,60
javascript-mixed-v3 | 20 | 36814 | 5.4 | 19,19,19,20,20,20,20,20,20,20,20,20,20,20,20,20,20,21,21,30

</details>

tophf:
I get roughly the same numbers with v3 when testing on this [script.txt](https://github.com/violentmonkey/violentmonkey/files/4957652/script.txt).

> the numbers you got are the `Kt` call

Yes, that block is almost entirely determined by the tokenizer function's prefgormance so might as well look at the topmost row for convenience.

orionlee:
> I get roughly the same numbers with v3 when testing on this [script.txt](https://github.com/violentmonkey/violentmonkey/files/4957652/script.txt).
> 

By roughly the same number, do you mean `v3` is roughly the same as `master`, or the same as `mixed` currently in this PR?

Anyhow, the numbers `Kt` time I reported above is done with the same script (the MPIV fork).

tophf:
Alas, I meant the PR.

orionlee:
Hmm, from my measurement, the `v3` code does run faster than the the one in PR.  I don't know why it was not reproduced on your end.

I did find the main reason that the `Kt` calls took longer during reload, after all the optimization done in my trial.  

---

The extra time happens during reload, triggered by CodeMirror initialization, but does not happen in adding codes/ scrolling.

1. mixed mode provides `innerMode()` function, which during reload (CodeMirror initialization with current script), CodeMirror makes more `token()` calls. 
  a. In the MPIV fork case, the presence of `innerMode` makes the number of javascript mode `token()` calls jumped from 778 to 2278. From profiler, the time of `token()` jumped from ~10ms to ~22ms, accounting for gap of the `Kt()` call time between `master` and `v3`.
  b. I also verified it by creating a plain javascript mode wrapper that forwards the call to javascript and implements a `innerMode()` method. (See the [wrapper code](https://github.com/orionlee/violentmonkey/blob/e-h-html-css-performance/src/common/ui/codemirror-js-mixed/javascript-wrap-only-innermode-copystate.js#L49)) The mere presence of the extra `innerMode` caused the issue.
2. I tried to see if similar increase number of calls happen in other operations, e.g., scrolling and editing. The answer is no. See [*] for details.
3. The use of `innerMode` is quite common in CodeMirror modes, most notably `htmlmixed`. So I'd be surprised if it is an issue.

<details>
  <summary>[*] Measuring performance impact on other operations</summary>
I patched CodeMirror's `javascript.js` to track number of calls to `token()` in a global `nJsToken`.

1. Scroll MPIV script all the way to the bottom. `nJsToken` is about 39,000 - 41,000, for for both `javascript` and `javascript-mixed-v3`.
```javascript
// In devtool
nJsToken = 0;

// in main editor window, press page down all the way to the bottom of the script
// then in devtool
console.log(nJsToken) // ~39,000 - 41,000
```

2. Adding some lines to MPIV script. `nJsToken` is ~5600, for both `javascript` and `javascript-mixed-v3`.
```javascript
// in devtool
_cm.setCursor({line: 343, ch: 0})

nJsToken = 0; // reset token() call track

// simulate edit
// _cm is an export of active CodeMirror instance in the performance branch code
_cm.replaceRange(`
let fooHtml = '<div id="bar">hello</div>';

function hello() {
  return 'world';
}

GM_addStyle(\` .bar {
  display: block;
}
\`);

`, {line:343, ch:0});

// show number of js token() calls due to edits
console.log(nJsToken) // ~5680
```

3. The `CodeMirror.setValue()` call, used during initialization, does cause increased number of token() calls.

```javascript
nJsToken = 0;
_cm.setValue(_cm.getValue()) // approximate code.vue's initialization

console.log(nJsToken) // ~778 for javascript, 2278 for javascript-mixed-v3
```

4. token() raw performance:

</details>

tophf:
Er, I actually meant the last before `v3` which is `v2`, not PR. I just didn't have time to assess your question properly. So, v3 is indeed faster than PR, but about the same as v2.

>The extra time [...] does not happen in adding codes/ scrolling

Thanks for the analysis, I was hoping you would find an explanation. Indeed, doing cut-paste of the entire code is almost as fast as the master branch.

So I guess we'll use the v3 version. 

<details><summary>In that case here's a few nits...</summary>

* check `GM.addStyle` too
* exclude `.vscode` directory from the PR and specify it in your own local git excludes, not in repo's .gitignore, but in .git/info/exclude, [more info](https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally).
* `maybeTokenBase` currently always performs all the `if` checks for a given token type when the second condition isn't satisfied even though only one `if` can possibly match due to the first condition restraint. Performance-wise there'll be no improvement but the structure/flow would be more obvious, I think.

```js
    function maybeTokenBase(stream, state, jsTokenStyle) {
      // TODO: might need to cache stream.current()
      const tokenText = stream.current();
      switch (state.jsState.lastType) {
      case 'string':
        if (RE_HTML_PLAIN_STRING.test(tokenText)) {
          // rule html-plain-string
          state.maybeTokenize = maybeTokenHtmlPlainString;
          prepReparsePlainStringInLocalMode(htmlNoMatchClosingMode, stream, state);
        }
        break;
      case 'quasi':
        if (RE_HTML_STRING_TEMPLATE.test(tokenText)) {
          // rule html-string-template
          state.maybeTokenize = maybeTokenStringTemplateInLocalMode;
          prepReparseStringTemplateInLocalMode(htmlMode, stream, state, true);
        } else if (/^`(\\)?\s*$/.test(tokenText)) {
          // rule html-string-template-2nd-line
          // first line is blank
          state.maybeTokenize = maybeTokenHtmlStringTemplateStarted2ndLine;
        }
        break;
      case 'variable':
        // rule css-gm-addstyle
        if (tokenText === 'GM_addStyle' || tokenText === 'GM.addStyle') {
          state.maybeTokenize = maybeTokenCssGMAddStyle1;
        }
        break;
      default:
      }
      if (jsTokenStyle === 'comment') {
        switch (tokenText.match(/^\/\*\s*(?:language=)?(css|html)\s*\*\/$/i)?.[1].toLowerCase()) {
        case 'css':
          // rule css-by-hint
          state.maybeTokenize = maybeTokenCssStringTemplateBegin;
          break;
        case 'html':
          // rule html-by-hint
          state.maybeTokenize = maybeTokenHtmlStringTemplateBegin;
          break;
        default:
        }
      }
      // else no match, no-op

      return STYLE_PASS;
    }
```

tophf:
Psst, I think we can eliminate the extra calls on init. Quoting CodeMirror documentation:

>**In a nested mode**, it is recommended to add an extra method, innerMode

So there's no need for `innerMode` in the primary javascript-mixed mode, only the nested ones should expose it i.e. cssMode, htmlMode, htmlNoMatchClosingMode:

```js
    const innerMode = state => ({
      state: state.localState || state.jsState,
      mode: state.localMode || jsMode,
    });
    const cssMode = CodeMirror.getMode(config, {
      name: 'css',
      innerMode,
    });
```

IDK about jsMode.

---

Found a problem: js-in-html nesting isn't properly handled (js isn't recognized, html highglight is broken after the nested `${}`):
```html
`<label style="flex: 1" title="${trimLeft(`
  Scale factors to use when “zooms to” selector is set to “custom”.
`)}">Custom scale factors:
  <input id=scales placeholder="${scalesHint}">
</label>`
```

orionlee:
Regarding `innerMode`: I'm afraid you misunderstood the documentation. It's the primary mode that need to expose the function. The nested mode in the doc means modes that contain other modes, i.e., the primary mode. The See Codemirror's `htmlmixed.js` for example.

The presence/absence of `innerMode` does not affect highlighting, but it could affect other features such as folding.
 
I think we can proceed with either `v2` or `v3` as the basis performance wise. Let me know your thoughts.

For the issue in html highlighting, I'd resolve that after we settle down on the performance issue. 
The current logic for detecting string template ending backtick is simplistic. In this case, it incorrectly treats the backtick at the beginning of line 3 as the string template ending backtick.
Moving it to the second line will work for now.  E.g,
```html
`<label style="flex: 1" title="${trimLeft(`
  Scale factors to use when “zooms to” selector is set to “custom”.`
)}">Custom scale factors:
  <input id=scales placeholder="${scalesHint}">
</label>`
```

tophf:
v3 is slightly faster so I see no reason not to use it.

orionlee:
It's a tad harder to understand / modify the matching rules with `v3` that I compensated with the comments in `maybeTokenBase()`.

I'll resolve other nits - I just noticed your earlier comments. Thanks for the tips of adding local git ignore as well.

tophf:
Still simple enough because there's not a lot of code.

tophf:
I actually just had a look at the v3 code and indeed it's not all that readable... So, given that v2 is on a par with v3 speed-wise, let's use v2 then.

tophf:
On the other hand, this turned out to be not as trivial as I thought at first so it feels out of the scope for Violentmonkey and somewhat unfair to other projects that may benefit from the same if you manage to get it merged into CodeMirror as an addon...

orionlee:
I'd start work on a refinement based on the v2 code.

Regarding putting the codes in CodeMirror project: it would make sense, but my understanding is that CodeMirror maintainer is reluctant to accept new add-ons/modes. So I think practically, we could first try to maintain the code here. Once it's used, we could ping CodeMirror maintainer again to see if he is willing to merge it.

orionlee:
I have pushed the optimization logic based on `v2` into this PR.

Furthermore, the backticks inside a string template is now handled. There is some caveats. See below.

Assuming backticks handling are acceptable, the things on list to do are:
- Support `GM.addStyle`
- Possibly support CSS in a tagged string template, e.g. (the `String.raw` threw off the matching)
```javascript
const myStyle = /* css */ String.raw`.foo {
  display: none;
}`;
```
- some minor nits

---

<details>
<summary>Limitations for backticks handling inside a string template</summary>

The logic now handles backticks within an expression. So the case you raised is handled properly:

image

But there are some limitations:
1. the expressions are tokenized as simple text (e.g., attribute value in the above example)
2. because it is treated as simple text, in some edge cases the tokenization is not quite correct (e.g., the `"` and `}" in the examples below threw off the tokenizer)
image
image

The limitations could be addressed if we properly handle those JS expressions as embedded javascript inside html/css. CodeMirror provides a mechanism to do it, e.g., [JSP / ASP support](https://codemirror.net/mode/htmlembedded/index.html). It'd require the multiplex addon which is another 4.8K uncompressed.

</details>

tophf:
>CodeMirror provides a mechanism to do it, e.g., JSP / ASP support. It'd require the multiplex addon which is another 4.8K uncompressed.

I'd say it's an integral/mandatory thing that must be implemented, otherwise people will report bugs with nested things shown above e.g. JS-template>HTML/CSS>JS-template>HTML/CSS.

tophf:
What will it take to support automatic handling of per-tag modes like `<style>foo { color: red }</style>`? Just plug in htmlmixed, hopefully?

orionlee:
> What will it take to support automatic handling of per-tag modes like `<style>foo { color: red }</style>`? Just plug in htmlmixed, hopefully?

It's on my bonus list (i.e., to try after done with other features). It should be plugging in `htmlmixed`, plus some adjustment on the codes tracking html tokenizer state.

orionlee:
- Added `GM.addStyle()` support
- tagged css string, e.g., 
```
  String.raw`some-css-string```
  // is supported with inline hint 
  String.raw /*css*/ `some-css-string`
```

Remaining issues
- Support JS expression in CSS
- Styling `<style></style>` in html (use htmlmixed)

---
<details>
<summary>Support JS expression in CSS</summary>

- Right now JS expressions are not recognized at all in CSS mode.
- They usually lead to some tokens are incorrectly styled as error.
- The most disruptive problem is when there are nested JS expression with nested backticks: it lead to premature exit of css mode, screwing up the tokenization of the remaining codes.
- We might need to live with some of the local problems.
```css
  border-${leftOrRightVal}: 1px solid;
  /* the tokenizer can't tell border-${leftOrRightVal} should form a valid css property */
```


image

</details>

orionlee:
With the latest commits, I have completed all that have been discussed:
- css in `<style>` tags are highlighted inside html (so as js in `<script>` tags)
- JS expressions are handled in CSS too.

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
```
tophf, line 1: Why is this necessary in the repo? Historically we don't add support for IDE artifacts.

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,756 @@
...
+      const attrContinuedStateDoubleQuote = dummyState1a.htmlState.state;
+      const tokenForAttContinuedDoubleQuote = dummyState1a.htmlState.tokenize;
+
+      const dummyStream1b = new CodeMirror.StringStream('<p class=\'someClass', 2, {});
+      const dummyState1b = htmlmixedMode.startState();
+      while (dummyStream1b.current() !== '<p class=\'someClass') {
+        htmlmixedMode.token(dummyStream1b, dummyState1b);
+      }
+      const attrContinuedStateSingleQuote = dummyState1b.htmlState.state;
+      const tokenForAttContinuedSingleQuote = dummyState1b.htmlState.tokenize;
```
tophf, line 85: Att -> Attr for consistency

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,756 @@
...
+      case 'javascript':
+        return -1; // let js mode handle ${ natively
+      default:
+        console.error('htmlmixedMode.indexOfJsExprStartInStream() - unrecognized mode:', modeName, modeState);
+      }
+      return -1; // should never reach here
+    };
+
+    // eslint-disable-next-line max-len, no-unused-vars
+    htmlmixedMode.ensureProperLocalModeStatePostJsExpr = function ensureProperLocalModeStatePostJsExpr(stream, state, style) {
```
tophf, line 220: You're not using `this` so no need for `function`. I guess you wanted it to have a name which can be supplanted differently I think:

```js
Object.assign(htmlmixedMode, {
  ensureProperLocalModeStatePostJsExpr(stream, state, style) {
    ........
  },
});
```

Or maybe even like this:

```js
const htmlmixedMode = CodeMirror.getMode(config, {
  name: 'htmlmixed',
  ensureProperLocalModeStatePostJsExpr(stream, state, style) {
    ........
  },
});
```

Bonus: there'll be no need for eslint overrides.

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,756 @@
...
+            state.tokenizePostJsExpr = state.localMode.tokenizePostJsExpr;
+          }
+        }
+      }
+      return style;
+    }
+
+    function indexOfBacktickInString(text) {
+      // whether encounter backtick ` (not those escaped)
+      const jsExprMatch = text.match(/[^\\][`]|^[`]/);
```
tophf, line 326: The first part ```[^\\][`]``` can be anywhere, not just at the beginning, is this intended?

orionlee: It's intentaional: the logic is to match backtick anywhere, but not those that are escaped, i.e., <code>\\`</code>.
 
The second part, <code>^[`]</code>, is to handle cases that the backtick happen to be the first character.

Examples for the match logic
```javascript
  someHTML=`<p>text</p>more text post last tag`;
```
After handling `</p>`, the html tokenizer would next tokenize <code>more text post last tag`;</code>, not knowing it is running in a JS string template and thus the backtick is meant to end the string template (and html). The logic here is to handle such cases.

.vscode/launch.json
orionlee: An accidental inclusion. It's removed. Thanks for catching it.

src/common/ui/codemirror-js-mixed/javascript-mixed.js
orionlee: Fixed. A typo I didn't catch

src/common/ui/codemirror-js-mixed/javascript-mixed.js
orionlee: Other than stylistic preferences, what's the reason the `function` is not preferrable?

Anyway, I changed to `Object.assign`, which also let me group related functions in a block, and keep the logic close to where they are called.

orionlee:
I also did some sanity check performance-wise (1. profile the loading in MPIV script in editor, 2.try editing the script in a standalone web page on a low-end tablet with Firefox Android), it remains acceptable.

There is some slight concern as the additional codes, primarily the JS expression handling, does increase the time.

I tried to measure the token time in isolation by using `runmode.html` helper, the median time to token 10k tokens.

branch | median time / 10k | overhead
---|---|---
master(javascript) | 12.06 |
mixed latest |  17.46 | 4.61
mixed-v2 (the old in performance branch, before js expression and htmlmixed) | 15.64 | 2.08

I think it remains acceptable, but would like to point out for just in case.

src/common/ui/codemirror-js-mixed/javascript-mixed.js
tophf: The problem I wanted to solve is the use of eslint overrides, duplication, and excessive length.

src/common/ui/codemirror-js-mixed/javascript-mixed.js
```diff
@@ -0,0 +1,810 @@
...
+          // will end up seeing previous token.
+          // dbg('   <--', `[${tokenStyle}]`, stream.current());
+          return null;
+        }
+      }
+
+      // optimization: short-circuit to skip local mode match when the rules won't cover
+      // Note: if the rules change (the <start> ones), the conditions here might need to be updated accordingly.
+      if (state.maybeLocalContext == null
+        && !['variable', 'comment', 'string', 'string-2'].includes(tokenStyle)) {
```
tophf, line 745: I've tested a big script without html/css in strings at all and it opens only slightly slower than on the master (150 vs 133 which is expected), pastes just as fast at 90ms.

One thing that may influence the overall time covertly is garbage collector and here I thought this temporary array is probably created a lot of times. Plain comparison is less elegant but just as readable:

```js
      if (state.maybeLocalContext == null
      && tokenStyle !== 'variable'
      && tokenStyle !== 'comment'
      && tokenStyle !== 'string'
      && tokenStyle !== 'string-2') {
```

orionlee: hmm, I assume js engines could recognize such array as a constant and avoid creating them repeatedly. Better be safe than sorry. The latest commits include your suggestion, plus a similar fix where the codes returned `[localMode, localState]` array.

Concerning garbage collector impact, I had checked in a few tweaks to cut down object creation with that in mind: cut down `new RunContext()`, and avoid calling `stream.current()` (that creates new substring).

tophf:
Could you create a separate repo and publish it as an npm package? We would just add it to our dependencies.

orionlee:
@tophf I should be able to get to it in a few days.

tophf:
Superseded by #1040.

PR #1045, opened 2020-08-20
 Update Korean translation

deunlee:
I added some missing translations.


tophf:
Thanks. 

We're still waiting for #1036 to make it possible to use translations like this one from GitHub so if takes a while you can use our [Transifex](https://www.transifex.com/projects/p/violentmonkey-nex/).

gera2ld:
Even with #1036 merged, the messages that are already translated in Transifex will not get updated because it's hard to decide which one is preferred in my script.

I'll see if I can do anything to improve it. Any help is welcome.

PR #1059, opened 2020-09-13
feat: option to open editor in new window from popup

### LINKED ISSUE #1058: Open editor in new window
I'd like to see an option to always open the editor in a new browser window, instead of a tab.

Usually I need to inspect the current page while I'm writing a user script for it, so I have to move the Violentmonkey editor to a new window by hand. Stylus has an option to open its editor in a new window, and it's very handy.

### Environment
- Browser: Firefox
- Browser version: 80
- OS: Ubuntu 18.04 LTS

tophf:
Fixes #1058.

The option is added in general section:

`[x] Open editor from popup in a new window`

The size/position of such a separate window is remembered automatically. In browsers that don't support chrome.windows.onBoundsChanged (only Chrome 86+ at the moment) a tooltip is displayed:

`Editor window position will be remembered only on resizing or saving`

### DIFF
--- a/src/background/utils/tabs.js
+++ b/src/background/utils/tabs.js
@@ -1,7 +1,8 @@
-import { getActiveTab, sendTabCmd, getFullUrl } from '#/common';
+import { getActiveTab, noop, sendTabCmd, getFullUrl } from '#/common';
 import ua from '#/common/ua';
 import { extensionRoot } from './init';
 import { commands } from './message';
+import { getOption } from './options';
 
 const openers = {};
 
@@ -12,6 +13,7 @@ Object.assign(commands, {
     active = true,
     container,
     insert = true,
+    maybeInWindow = false,
     pinned,
   }, src = {}) {
     // src.tab may be absent when invoked from popup (e.g. edit/create buttons)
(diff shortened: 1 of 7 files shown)


src/options/views/edit/index.vue
```diff
@@ -163,6 +192,9 @@ export default {
   created() {
     this.script = this.initial;
     document.addEventListener('keydown', this.switchPanel);
+    if (options.get('editorWindow') && global.history.length === 1) {
+      global.chrome.windows?.getCurrent({ populate: true }, setupSavePosition);
```
gera2ld, line 196: How about using the `browser` API which is supposed to be the standard?

ps. maybe we should also handle the case where `chrome` does not exist on `global`, it's exposed in Firefox now but the document says it's not guaranteed in all browsers.

src/background/utils/tabs.js
```diff
@@ -21,23 +23,40 @@ Object.assign(commands, {
...
+      const wndOpts = {
+        url,
+        incognito: canOpenIncognito && incognito,
+        ...getOption('editorWindowSimple') && { type: 'popup' },
+        ...ua.isChrome && { focused: !!active }, // FF doesn't support this
+        ...storeId,
+      };
+      const pos = getOption('editorWindowPos');
+      const hasPos = pos && 'top' in pos;
+      const wnd = await browser.windows.create({ ...wndOpts, ...pos }).catch(hasPos && noop)
```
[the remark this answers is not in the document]
tophf: When it doesn't like the provided position for whatever reason.

src/options/views/edit/index.vue
tophf: It was a micro-optimization to avoid adding getCurrent to browser.js.

src/options/views/edit/index.vue
```diff
@@ -163,6 +192,9 @@ export default {
   created() {
     this.script = this.initial;
     document.addEventListener('keydown', this.switchPanel);
+    if (options.get('editorWindow') && global.history.length === 1) {
+      browser.windows?.getCurrent({ populate: true }, setupSavePosition);
```
gera2ld, line 196: `browser.windows.getCurrent` should return a `Promise`.

Ref: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/getCurrent

tophf: Sorry, just woke up...

gera2ld (approved):
LGTM

tophf:
Combined the commits into one, no changes.

PR #1068, opened 2020-09-26
fix: gmxhr with partial responses while still loading

### LINKED ISSUE #1066: [Chromium] GM.xmlHttpRequest is very slow and buggy
### What is the problem?

`GM.xmlHttpRequest` is very slow and buggy.

### How to reproduce it?
Use this script to download some file (30+MB) from a file hosting (sorry, no URL example, I will add it later (if you would still need))
```js
// ==UserScript==
// @name        New script - GM.xmlHttpRequest
// @match       <PUT_HERE_THE_ADDRESS>   // TODO: add a demo URL
// @grant       GM.xmlHttpRequest
// @version     1.0
// @author      -
// @description GM.xmlHttpRequest slow chrome
// ==/UserScript==

const UNSENT           = 0;
const OPENED           = 1;
const HEADERS_RECEIVED = 2;
const LOADING          = 3;
const DONE             = 4;
const xml = GM.xmlHttpRequest({
  method: "GET",  
  url: "<PUT_HERE_THE_ADDRESS>",    // TODO: add a demo URL
  onload: function(response) {
    console.log("---onload---");
    console.log(response);
    console.log("response.loaded", response.loaded);
    console.log("response.responseText.length", response.responseText.length);
    console.log("response.response.length", response.response.length);
  },
  onreadystatechange: function(response) {
   console.log("onreadystatechange", response.readyState, response, response.responseText.length);
    if (response.readyState === HEADERS_RECEIVED) {
        console.log(response);
        console.log(response.responseHeaders);
    }
  },
});
console.log(xml);
```

`GM.xmlHttpRequest` works fast without freezing.

No problem in Firefox. And no problem with Tampermonkey _(but it has no `response.response` and `response.responseText === ""` for `onreadystatechange` event)_ _UPD: But it's normal for using with `responseType: "arraybuffer";_

Downloading in the background script work absolutely OK (I checked it with DevTools). So it's the problem with the sending data to the content script.

With little files ~5 MB it works ~OK.
With files ~15 MB I works slow.
For file 20 MB + it begins to work very slow, and probably freezing (no progress). Also the extension start to lag (Can't immediately open the popup). Memory consuming is increasing.

### Environment
- Browser: Chrome, Opera.
- Browser version: Chromium 85 based
- OS: Win 10

tophf:
Fixes #1066 (the bug with the length of response reported in comments).

The problem was that while XHR is still loading, response and responseText contain only partial info per the specification, for example 65536 bytes.

The fix seems to be working correctly ~at first try but I want to test it more because I haven't yet analyzed the content script (it may need some changes as well)~.

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -157,9 +157,9 @@ function xhrCallbackWrapper(req) {
   let contentType;
   let numChunks;
   let response;
-  let responseSent;
-  let responseTextSent;
+  let responseText;
   let responseHeaders;
+  const sent = new Set([undefined]);
   const { id, chunkType, xhr } = req;
   // Chrome encodes messages to UTF8 so they can grow up to 4x but 64MB is the message size limit
   const chunkSize = 64e6 / 4;
@@ -178,15 +178,6 @@ function xhrCallbackWrapper(req) {
       return { responseHeaders };
     }
   };
-  const getResponseText = () => {
-    try {
-      const text = xhr.responseText;
-      responseTextSent = !!text;
-      return { responseText: text === response ? ['same'] : text };
-    } catch (e) {
-      // ignore if responseText is unreachable
-    }
-  };
   const chainedCallback = (msg) => {
     lastPromise = lastPromise.then(() => req.cb(msg));
   };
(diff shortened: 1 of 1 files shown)


tophf:
Merging this to let our potential testers use it in an RC.

gera2ld:
LGTM

PR #1074, opened 2020-09-30
feat: add a note in UI about using custom CM themes

tophf:
Now that thanks to d1539d3 themes can be used as is, without editing them, this PR adds a simple how-to:

> Custom options for CodeMirror and addons in JSON object like `{"indentUnit":2, "smartIndent":true}` however note that some of them may not work in Violentmonkey. See [full list](https://codemirror.net/doc/manual.html#config). To use a custom CodeMirror theme, specify its file name like `"theme": "3024-day"` here and paste the [actual theme's CSS](https://github.com/codemirror/CodeMirror/tree/master/theme) in "Custom Style" input below.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -122,6 +122,10 @@ descEditorOptions:
     them may not work in Violentmonkey. See <a
     href="https://codemirror.net/doc/manual.html#config" target="_blank"
     rel="noopener noreferrer">full list</a>.
+    To use a custom CodeMirror theme, specify its file name like
+    <code>"theme": "3024-day"</code> here and paste the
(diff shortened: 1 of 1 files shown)


src/options/views/tab-settings/vm-editor.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <section>
     <h3 v-text="i18n('labelEditor')"></h3>
-    <p v-html="i18n('descEditorOptions')" />
+    <p v-html="i18n('descEditorOptions') + ' ' + i18n('descEditorTheme')" />
```
gera2ld, line 4: Can we use two different `<p>`s to support languages that don't like whitespaces better?

tophf: No, I will modify the existing message then.

Issue #1052, opened 2020-08-31
Violentmonkey beta in Chrome Web Store


darkred:
*(Continuing issue #5)*

My suggestion is to consider publishing Violentmonkey beta in Chrome Web Store (CWS).

Currently, with every new rc version (there can be a lot, 15 versions since last stable release v2.12.7 in last February), you have to manually install it as an unpacked extension in Chrome.

I believe it would be convenient and facilitate testing to have a separate Violentmonkey beta build. The stable version and the beta version will be installed as two distinct extensions: they will auto-update independently of each other, and won't share any settings or storage.

For reference, I use the beta/dev builds of uBlock Origin, Tampermonkey, and Stylus.
*(note: uBlock Origin publishes uBO beta builds for Firefox too, signed and self-hosted in its [GitHub repo](https://github.com/gorhill/uBlock/releases))*.

PS. Thank you for your tremendous work all this time!

gera2ld:
Thank you for your suggestion.

I think self-hosted version on AMO is a good idea for testing, since it can be easily updated to the stable version if the beta channel is no longer needed. The only concern is that data might be unexpectedly lost when something goes wrong with the extension. But I think power users that choose to use a testing version should be able to handle it.

> I believe it would be convenient and facilitate testing to have a separate Violentmonkey beta build. The stable version and the beta version will be installed as two distinct extensions: they will auto-update independently of each other, and won't share any settings or storage.

Does that mean two versions for two different groups of users? It's like maintaining two independent extensions, and both need to be reviewed 😅.

When coming to a stable release, the two versions may be exactly the same. Will it be a problem if two different extensions have the same or similar code? I think I've read from somewhere that this is not recommended or even not allowed in CWS, but I can't find the reference any more.

tophf:
In CWS you can have beta extensions with the same code without any problems.

darkred:
Yes, the two variants of the extension will both have to pass the CWS review procedure.
But, there's an effort to improve that procedure in the last months *( see [1](https://github.com/uBlockOrigin/uBlock-issues/issues/745#issuecomment-545654074), [2](https://www.reddit.com/r/chrome/comments/dgoymg/warning_ubo_ublock_origin_will_possibly_be/f3g3jlq/)* )

And, I believe that, if a problem occurs with the version being stuck in 'rejected' status,
[...]

PR #1078, opened 2020-10-05
Proposal: beta releases

gera2ld:
This is a proposal workflow for self-hosted releases or beta releases that can be published more frequently and tested by power users.

Since we don't have a lot of developers and we are unlikely to develop many features concurrently, I don't think it necessary to have multiple channels like in tremendous projects as Chrome and Firefox. So this proposal is actually prereleases + milestone like releases.

1. version numbers
    Web extensions do not support semver (maybe supported by some, but not all), we can use more dot separated numbers as version numbers. An extra number is added to make this work.
    For example, say `2.12.7` is a release, the following prereleases can be `2.12.7.1`, `2.12.7.2`, etc. And when it comes to a next release, the extra number can be removed (or reset to 0 and omitted), i.e. `2.12.8`.

2. making prereleases
    A new command is added to create prereleases: `yarn bump`.
    Under the hood it just increases `beta` (the extra number) in `package.json` and creates a commit with a prerelease tag.

3. making releases
    There is no change in the command, which should be `yarn version --patch`. Just that when `version` is changed, the `beta` field will be removed.

4. self-hosted release
    In Chrome, a separate extension is supposed to be created and named *Violentmonkey BETA*, released with an extra version number as mentioned above.
    In Firefox we can use self-hosted channel for testing purpose as recommended by AMO.
    The self-hosted version will be built with `yarn build:selfHosted`. The difference is that an `update_url` will be inserted to `manifest` for update checking.
    Another command `yarn update:selfHosted` will update the version list file on GitHub, so that the self-hosted extension can get a list of the latest (pre)releases there.

Caveats:

- Why not use semver prereleases?
    First, not all browsers support prereleases. Another problem is that, if we transform semver to a string with dot separated numbers, we need to change the first 3 numbers to that from the last release. For example, `2.12.8-1` should be transformed to `2.12.7.1` because `2.12.8` is not created yet. But I guess it's not so easy to find the last release number. So it's much simpler to use an extra number rather than change `version` directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-Violentmonkey
-=============
+# Violentmonkey
 
 [![Chrome Web Store](https://img.shields.io/chrome-web-store/v/jinjaccalgkegednnccohejagnlnfdag.svg)](https://chrome.google.com/webstore/detail/violentmonkey/jinjaccalgkegednnccohejagnlnfdag)
 [![Firefox Add-ons](https://img.shields.io/amo/v/violentmonkey.svg)](https://addons.mozilla.org/firefox/addon/violentmonkey)
@@ -12,14 +11,17 @@ It works on browsers with [WebExtensions](https://developer.mozilla.org/en-US/Ad
 
 More details can be found [here](https://violentmonkey.github.io/).
 
-Related projects
----
+## Related projects
+
 - [Violentmonkey for Opera Presto](https://github.com/violentmonkey/violentmonkey-oex)
 - [Violentmonkey for Maxthon](https://github.com/violentmonkey/violentmonkey-mx)
 
-Development
----
-Make sure [Node.js](https://nodejs.org/) greater than v10.0 is installed.
+## Workflows
+
+### Development
+
+Make sure [Node.js](https://nodejs.org/) greater than v10.0 and Yarn v1.x is installed.
+
 ``` sh
 # Install dependencies
 $ yarn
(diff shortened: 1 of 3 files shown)


tophf:
I'm still a noob in tooling so would you mind explaining if this is something that I can make use of?

gera2ld:
@tophf With this proposal, we can create a prerelease with `yarn bump` rather than manually create the RC tag, and create a patch release with `yarn version --patch` as before.

The main change in this PR is the convention to use an extra `beta` field in `package.json` for prereleases, as the fourth number of the version numbers. For example, in `package.json`:

```json
{
  "version": "2.12.7",
  "beta": "17",
}
```

`yarn build` will create a package of version `2.12.7.17`.

tophf:
Since I can't publish, it's something you'd have to do, preferably as a separate beta extension per #1052.

gera2ld:
The version number matters for beta releases. This PR decides how the version number is generated.
@tophf If you are not against it I'll merge it later.

tophf:
I'm not against it as I'm not the one to use it, apparently :-)

tophf:
@gera2ld, so it creates releases automatically when a tag is pushed? I've just pushed a fix (ce8a3a04) and accidentally included two old tags that I already deleted here online (2.12.7.18 and 2.12.7.19) but didn't delete locally.

gera2ld:
yep. maybe i should check the last version first?

tophf:
Yeah, I guess there will be no need to publish "in the past".

tophf:
Looks like it still creates releases... Turns out I didn't delete these tags locally and pushed them once again. This time only one release was created though.

gera2ld:
The problem is, the workflow definition in the commit/tag that is pushed will be used. So the new action definition will only affect tags in the future...

PR #1079, opened 2020-10-05
fix: dark mode theme of CodeMirror

gera2ld:
There is no `CodeMirror-wrap` in the codebase.

### DIFF
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -539,52 +539,48 @@ $selectionDarkBg: rgba(73, 72, 62, .99);
     background-color: hsla(40, 100%, 50%, 0.1);
   }
   // mostly copied from Monokai theme
-  .CodeMirror-wrap {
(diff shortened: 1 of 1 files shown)


tophf:
LGTM

tophf:
BTW CodeMirror-wrap is added by CM itself but the old code was incorrect anyway because `.CodeMirror-wrap .CodeMirror` doesn't match anything since these two classes belong to the same element. It was broken recently by my #1069.

gera2ld:
> CodeMirror-wrap is added by CM itself

Maybe something is changed in the latest CodeMirror because I could not find it in the DOM either, `document.querySelector('.CodeMirror-wrap')` gives `null`.

Screen Shot 2020-10-06 at 15 13 13

tophf:
Oh, today I learned it's added only when line wrapping option is enabled... Previously I thought it's a class for the cm.display.wrapper element.

PR #1084, opened 2020-10-17
fix: add placeholders on pasting/duplicating/sorting/etc.

tophf:
* fix: add placeholders on pasting/duplicating/sorting/etc.

  Also speed up it a little by using CodeMirror's generation instead of comparing the re-assembled code string which can be long. This also avoids stressing the garbage collector after each change.

* fix: catch regexp compilation error in Ctrl-F and show it as `title`.

  Uses the standard DOM `title` which is somewhat unusual but I think it makes sense to be different from the other UI tooltips.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -141,6 +141,15 @@ editLabelMeta:
 editLabelSettings:
   description: Settings section in settings tab of script editor.
   message: Script settings
+editLongLinePlaceholder:
+  description: Shown in the editor in lines that were cut due to being too long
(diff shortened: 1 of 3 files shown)


tophf:
@gera2ld, the PR works so it's not like it needs an in-depth review, just a couple of potentially contentious points:

* child's component `getRealContent` is called directly by the parent component (to avoid unnecessary concatenation of code)
* the placeholder line's font is `italic` to show it's not an editable line
* the placeholder ends with `...[Line is too long]` text to make the effect obvious

PR #1088, opened 2020-10-21
fix: use RFC algo for UUID

### LINKED ISSUE #1087: Incorrect UUID generation
https://github.com/violentmonkey/violentmonkey/blob/386aa7c09d23d0c1e7e6e88ebe16b81581bf890c/src/background/utils/db.js#L358-L361

This function does not always generate a valid UUID. A random UUID `xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx` should have `M`=`4` and `N` chosen from `8` to `b`.

### See also
* [Algorithm for creating a random UUID in RFC 4122](https://tools.ietf.org/html/rfc4122#section-4.4)
* [UUID on Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier)
* [Implementation of Greasemonkey](https://github.com/greasemonkey/greasemonkey/blob/e0caac00f34f39fe87e23e43df7c33d847f46e11/src/user-script-obj.js#L68-L88)

tophf:
Fixes #1087.

The advantage is that it's specification-compliant and a single crypto.getRandomValues is faster than multiple Math.random.
The only drawback is that we lose 6 bits of randomness out of 96 we had previously but I think we can live with that :-)

### DIFF
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -1,5 +1,5 @@
 import {
-  compareVersion, i18n, getFullUrl, isRemote, getRnd4, sendCmd, trueJoin,
+  compareVersion, i18n, getFullUrl, isRemote, sendCmd, trueJoin,
 } from '#/common';
 import {
   CMD_SCRIPT_ADD, CMD_SCRIPT_UPDATE, INJECT_PAGE, INJECT_AUTO, TIMEOUT_WEEK,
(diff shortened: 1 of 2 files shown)


src/background/utils/db.js
```diff
@@ -356,8 +356,18 @@ export function checkRemove({ force } = {}) {
...
-  return `${idSec}-${getRnd4()}-${getRnd4()}-${getRnd4()}-${getRnd4()}${getRnd4()}${getRnd4()}`;
+  const rnd = new Uint16Array(6);
+  window.crypto.getRandomValues(rnd);
+  // xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx
+  // We're using UUIDv4 variant 1 so N=4 and M=8
+  // See format_uuid_v3or5 in https://tools.ietf.org/rfc/rfc4122.txt
+  rnd[1] = rnd[1] & 0x0FFF | 0x4000; // eslint-disable-line no-bitwise
+  rnd[2] = rnd[2] & 0x3FFF | 0x8000; // eslint-disable-line no-bitwise
+  return `${
+    (id + 0x10bd_e6a2).toString(16).slice(-8)
```
gera2ld, line 367: Maybe we can also get rid of `id` here.

tophf: Done.

src/background/utils/db.js
```diff
@@ -355,9 +355,15 @@ export function checkRemove({ force } = {}) {
...
-  const idSec = (id + 0x10bde6a2).toString(16).slice(-8);
-  return `${idSec}-${getRnd4()}-${getRnd4()}-${getRnd4()}-${getRnd4()}${getRnd4()}${getRnd4()}`;
+function getUUID() {
+  const rnd = new Uint16Array(8);
+  window.crypto.getRandomValues(rnd);
+  // xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx
+  // We're using UUIDv4 variant 1 so N=4 and M=8
+  // See format_uuid_v3or5 in https://tools.ietf.org/rfc/rfc4122.txt
+  rnd[3] = rnd[1] & 0x0FFF | 0x4000; // eslint-disable-line no-bitwise
+  rnd[4] = rnd[2] & 0x3FFF | 0x8000; // eslint-disable-line no-bitwise
```
tophf, line 365: typo fixed in b0669655.

PR #1091, opened 2020-10-25
fix: accept any MIME type on update check + show error info

### LINKED ISSUE #1090: Dashboard: "Failed fetching update information" for my userscript
I have a couple of userscripts, but Violentmonkey's dashboard can't seem to check for updates. I'm hosting these userscripts on my own domain. I'm confident there must be something wrong with the userscript, but I can't figure it out. There doesn't seem to be a way to debug it. There's no information in the console or network tab.

1. Install userscript: https://steamhunters.com/userscripts/steam-games-analyzer.user.js
2. Go to Violentmonkey dashboard
3. Click the "Check for updates" button underneath the userscript.

### Environment
- Browser: Chrome
- Browser version: 86.0.4240.111 (Official Build) (64-bit)
- Violentmonkey version: v2.12.7
- OS: Windows 10 Home

tophf:
* Fixes #1090.

* Restores the correct tooltip with the error details (the `error.url` was lost in 80cde6da0ac731d5a3d13aaedfe2c67db2df0d8a)

  image

* ~shows an instruction how to open devtools of the background page to debug the network problems:~

  <details>
  image

  In Firefox the instruction is slightly different, it also doesn't allow extensions to open `about:debugging`.
  </details>

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -205,6 +205,8 @@ filterScopeCode:
 filterScopeName:
   description: Option in dashboard's search scope filter.
   message: Name
+genericError:
+  message: Error
 genericOff:
   description: 'To indicate something is turned off or disabled, similar to "no".'
   message: 'off'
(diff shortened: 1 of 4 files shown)


gera2ld:
I think the instruction for debugging is trivial for normal users, and should not be shown in the extension. Maybe it's better to link to a page somewhere else, e.g. the Violentmonkey website.
I guess normal users would not try to debug the errors, so the error message should be enough most of the time.

tophf:
Okay, removed it.

PR #1092, opened 2020-10-25
Create icons with Jimp

gera2ld:
This PR is a follow-up to #615 (wow it's been more than a year) and preparation for the beta version. There should be a different icon for the beta version I think. The improvement for 16px icon is preserved.


tophf:
I'd say it's blurry compared to my hand-optimized version:

icon16

gera2ld:
@tophf It's true. 😅 Did you change the color a bit?

tophf:
I don't remember but that's probably the result of line sharpening: the lines get darker, the surrounding pixels get lighter.

PR #1093, opened 2020-10-25
chore: switch to Sharp to improve legibility of small icons

tophf:
Closely replicates the crispness of the hand-optimized small icons by switching to https://github.com/lovell/sharp

Hand-made (darker mouth) vs sharp's: new
Hand-made (darker mouth) vs jimp's: old

The mouth was darker apparently because I manually darkened its pixels, but there's no way to automate it without manually preparing a separate overlay image with just the contour of the mouth and then blend it on top once or a couple of times. I can do that but maybe it's unnecessary?


gera2ld:
That looks better. I think it's good enough, no need for the manual tweaks. We can improve it in the future if there are some simple declarative methods to emphasize the lines or something like that.

gera2ld:
The problem with sharp is that it's not pure JavaScript, hard to install in some environments. Hope Jimp will [support sharpen](https://github.com/oliver-moran/jimp/issues/809) someday.

tophf:
Oh I didn't know about installation problems. Well, maybe we could find a separate sharpen library in pure js and restore jimp + add sharpen + add [blit](https://github.com/oliver-moran/jimp/tree/master/packages/plugin-blit) to emphasize the outer edge (now done using `composite`). Or we can wait until someone complains...

tophf:
FWIW sharpening can be poorly imitated via jimp: `img.convolute([[0, -a, 0], [-a, a * 4 + 1, -a], [0, -a, 0]])` where `a` is the strength like `0.1`. The problem is that it processes the image in RGB so the colors get over-saturated and hence flattened. Theoretically we can convert the image to HSL or YUV by manually transforming each pixel, extract and sharpen the lightness channel, combine and convert back to RGB.

PR #1100, opened 2020-11-07
feat: instant inject via synchronous XHR

tophf:
Related: #420.

Our [upcoming 2.12.8 release](https://github.com/violentmonkey/violentmonkey/releases) is already sufficiently fast so until proven otherwise we probably shouldn't add this feature but I'm sharing this PR anyway, just in case. As they say, TBD.

FWIW, there seems to be a better alternative for Chrome: inject the prepared data in webNavigation.**onCommitted** via executeScript + run_at:document_start - turns out this runs before the main content scripts in default mode (that is without chrome.declarativeContent). It requires `webNavigation` permission though, which will disable the extension on update, most likely.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -351,6 +351,10 @@ labelInjectionMode:
 labelInstall:
   description: Shown in the title of the confirm page while trying to install a script.
   message: Installing script
+labelInstantInject:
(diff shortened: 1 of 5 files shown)


tophf:
Now that we inject much less data at document-start, I've implemented it as c6249327, disabled by default.

PR #1101, opened 2020-11-12
feat: faster load of popup, editor, dashboard

tophf:
It bugged me for a while that the popup is initially rendered with an empty script list for a fraction of a second. So, I've tried to find a way to improve it. The result is that now all UI loads faster by fetching the data directly from the background page which is synchronous.

Additional speedups:

* dashboard: replaced the very slow `cache2blobUrl` with a data URI.
* editor when launched from popup: `GetData` sends only the edited script at first
* popup: `SetPopup` data is prefetched by the background script when the popup starts loading.

The no longer needed prefetch for options in browser.js was removed.

### DIFF
--- a/src/common/load-script-icon.js
+++ b/src/common/load-script-icon.js
@@ -1,0 +1,43 @@
+const images = {};
+
+/**
+ * Sets script's safeIcon property after the image is successfully loaded
+ * @param {VMScript} script
+ * @param {Object} [_]
+ * @param {?string} [_.default]
+ * @param {string} [_.key]
+ * @param {Object} [_.cache]
+ * @returns {Promise<boolean>}
+ */
+export function loadScriptIcon(script, {
+  default: defaultIcon = null,
+  key = 'safeIcon',
+  cache = {},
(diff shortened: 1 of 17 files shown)


src/common/load-script-icon.js
```diff
@@ -0,0 +1,37 @@
...
+ * @returns {Promise<boolean>}
+ */
+export function loadScriptIcon(script, defaultIcon = null) {
+  const { key = 'safeIcon' } = loadScriptIcon;
+  const { icon } = script.meta;
+  const url = script.custom?.pathMap?.[icon] || icon;
+  const isNewUrl = url !== script[key];
+  let promise = isNewUrl && url ? images[url] : Promise.resolve(false);
+  if (!promise) {
+    promise = Promise.resolve(loadScriptIcon.store.cache[url] || fetchImage(url));
```
gera2ld, line 16: How about just attaching `cache` so that the structure of `store` does not matter?

tophf: Done.

PR #1105, opened 2020-11-15
feat: show script name in syntax errors during injection

### LINKED ISSUE #1104: [Firefox] Improve error reporting
### What is the problem?

Runtime errors in scripts are reported to the web console with the script's filename and line number(s), but the output for compile-time errors (e.g. syntax errors) omits everything but the error message.

vm-error

This makes it hard to see error messages even if they're [enabled](https://stackoverflow.com/a/62381648) (they're not by default) and filtered ("Error:" is a very common prefix).

### How to reproduce it?

Install the following script and run it on a page with no other scripts enabled.

#### Script

```javascript
// ==UserScript==
// @name          Syntax Error
// @version       0.0.1
// @namespace     violentmonkey-test
// @include       *
// @grant         GM_log
// @inject-into   content
// ==/UserScript==

await 42
```

An error in the web/browser console showing the message, filename and line number, e.g.:

<table>
<tr><th>message</th><td>[Violentmonkey] SyntaxError: await is only valid in async functions and async generators</td></tr>
<tr><th>source</th><td>vm-error.user.js#10:1</td></tr>
<table>

<table>
<tr><th>message</th><td>Error: await is only valid in async functions and async generators</td></tr>
<tr><th>source</th><td><code>undefined</code></td></tr>
<table>

### Environment

- **Violentmonkey**: v2.12.7
- **Browser**: Firefox v82.0.3
- **OS**: Linux (Arch)

tophf:
Fixes #1104 as much as possible.

With `@inject-into content` our own wrapper code won't even run so there'll be no error message at all. There's nothing we can do about it, the error doesn't seem to be reported anywhere so this is a Firefox bug I guess.

In normal mode the error source is incorrectly attributed by Firefox to the page itself. This can be intercepted. This PR explicitly prints such unrecoverable errors that occur during appendChild() call.

P.S. In Chrome the same happens when devtools was closed at the time the error was printed, but when devtools is opened the error is printed with correct attribution. However, there's nothing we can do about it because the error cannot be intercepted via `error` event or try-catch and anyway the user may just reload the page with devtools opened and see the correct message. FWIW Strictly speaking it's possible to catch the error event in `web` bridge and then send the error to `content` but that seems like an overkill for something that already has a simple solution (opening devtools and reloading).

### DIFF
--- a/src/injected/content/inject.js
+++ b/src/injected/content/inject.js
@@ -11,6 +11,8 @@ import { defineProperty, describeProperty, forEachEntry, objectPick } from '#/co
 import {
   forEach, push,
   append, createElementNS, remove, NS_HTML,
+  addEventListener, removeEventListener,
+  log,
 } from '../utils/helpers';
 import bridge from './bridge';
 
@@ -26,11 +28,12 @@ const { get: getHead } = describeProperty(Document.prototype, 'head')
   || describeProperty(HTMLDocument.prototype, 'head'); // old FF before 61;
 const { get: getDocElem } = describeProperty(Document.prototype, 'documentElement');
 const { appendChild } = Document.prototype; // same as Node.appendChild
+const stringIncludes = String.prototype.includes;
 
 bridge.addHandlers({
   // FF bug workaround to enable processing of sourceURL in injected page scripts
   InjectList(runAt) {
-    bridge.realms[INJECT_PAGE].lists[runAt]::forEach(item => inject(item.code));
+    bridge.realms[INJECT_PAGE].lists[runAt]::forEach(inject);
   },
 });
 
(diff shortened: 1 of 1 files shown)


src/injected/content/inject.js
```diff
@@ -117,12 +122,24 @@ function checkInjectable() {
...
 }
 
-function inject(code) {
+function inject(item) {
   const script = document::createElementNS(NS_HTML, 'script');
+  // Firefox ignores sourceURL comment when a syntax error occurs so we'll print the name manually
+  let onError;
+  if (bridge.isFirefox) {
+    onError = e => {
+      if (e.error.stack::stringIncludes(browser.runtime.getURL('/sandbox'))) {
```
gera2ld, line 131: Will `e.error.stack` be empty or something other than a string here?

tophf: Added a check in case the page broken it somehow.

PR #1114, opened 2020-11-24
fix: toggle save/reset buttons depending on value

### LINKED ISSUE #1113: Editor options bug
### What is the problem?
The editor options are lost when saved without modification.

### How to reproduce it?
1. Open the Settings page.
2. Expand the advanced options.
3. Click ‘Save’ in the ‘Editor’ section directly.

No options should be changed.

The input area displays `null` and user options are lost.

https://github.com/violentmonkey/violentmonkey/blob/44d1ffbb602d087c1d5fb9c4278bd14892957483/src/common/ui/setting-text.vue#L40
The problem seems to be that `parseJson` is not invoked until the input area is changed.

### Environment
- Browser: Firefox
- Browser version: 83.0
- Violentmonkey [pre-release behavior](https://github.com/violentmonkey/violentmonkey/releases): 2.12.7.18
- OS: Windows

tophf:
Fixes #1113.

### DIFF
--- a/src/common/ui/setting-text.vue
+++ b/src/common/ui/setting-text.vue
@@ -1,17 +1,26 @@
 <template>
-  <textarea
-    class="monospace-font"
-    :class="{'has-error': error}"
-    spellcheck="false"
-    v-model="value"
-    :disabled="disabled"
-    :title="error"
-    @change="onChange"
-  />
+  <div>
+    <textarea
+      class="monospace-font"
+      :class="{'has-error': error}"
+      spellcheck="false"
+      v-model="value"
+      :disabled="disabled"
+      :title="error"
+      @change="onChange"
+    />
+    <button v-if="hasSave" v-text="i18n('buttonSave')" @click="onSave"
+            :disabled="disabled || !canSave"/>
+    <button v-if="hasReset" v-text="i18n('buttonReset')" @click="onReset"
+            :disabled="disabled || !canReset"/>
+    <slot/>
+  </div>
 </template>
 
 <script>
+import { deepEqual, objectGet } from '../object';
 import options from '../options';
+import defaults from '../options-defaults';
 import hookSetting from '../hook-setting';
 
 export default {
@@ -23,12 +32,19 @@ export default {
       type: Boolean,
       default: true,
     },
+    hasSave: {
+      type: Boolean,
+      default: true,
+    },
+    hasReset: Boolean,
   },
   data() {
     return {
       value: null,
       jsonValue: null,
       error: null,
+      canSave: null,
+      canReset: null,
     };
   },
   created() {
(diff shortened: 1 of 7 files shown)


src/common/ui/setting-text.vue
```diff
@@ -23,39 +32,70 @@ export default {
...
-      try {
-        this.jsonValue = JSON.parse(this.value);
-        this.error = null;
-      } catch (e) {
-        this.error = e.message || e;
+    /** `event` is absent when calling from save() */
+    onChange(event) {
+      if (!this.error) {
+        const value = this.json ? this.jsonValue : this.value;
+        if (this.sync || !event) options.set(this.name, value);
```
gera2ld, line 85: Is this line virtually `onSave`?

src/common/ui/setting-text.vue
```diff
@@ -23,39 +32,70 @@ export default {
...
-      if (this.error) return;
-      const value = this.json ? this.jsonValue : this.value;
-      if (this.sync) options.set(this.name, value);
-      this.$emit('change', value);
+    onInput(val) {
+      this.canSave = this.hasSave && !this.error && !deepEqual(val, this.savedValue || '');
+      this.canReset = this.hasReset && !deepEqual(val, this.defaultValue || '');
+    },
+    onSave() {
+      if (this.onChange()) this.$emit('save');
```
gera2ld, line 95: Why is `onChange` called here? I think data should have been updated when `save` is emitted.

gera2ld (changes requested):

src/common/ui/setting-text.vue
tophf: onSave does nothing except sending an event to the parent, which is why it calls onChange to actually write the value.

src/common/ui/setting-text.vue
tophf: It's for deduplication. The only purpose of a dedicate onSave method is to inform the parent so it could do something. I guess I can extract the actual saving into a separate method to reduce confusion.

tophf: BTW, saving is performed automatically when the text field loses focus, so there's no need for the save button strictly speaking, we could do without it by using an indicator for unsaved state and `beforeunload` listener to prevent the loss of data.

tophf: Extraction example:
```js
    onChange() {
      this.commit({ change: true });
    },
    onInput(val) {
      this.canSave = this.hasSave && !this.error && !deepEqual(val, this.savedValue || '');
      this.canReset = this.hasReset && !deepEqual(val, this.defaultValue || '');
    },
    onSave() {
      if (this.commit({ save: true })) {
        this.$emit('save');
      }
    },
    onReset() {
      options.set(this.name, this.defaultValue);
    },
    commit({ change, save }) {
      if (!this.error) {
        const value = this.json ? this.jsonValue : this.value;
        if (this.sync || save) options.set(this.name, value);
        if (change) this.$emit('change', value);
        return true;
      }
    },
```

gera2ld: We can update data on `input` and save/sync on `change`.
A simple way to handle `click` of save button is to wrap in `setTimeout` and avoid calling `change` manually.

gera2ld: > BTW, saving is performed automatically when the text field loses focus, so there's no need for the save button strictly speaking, we could do without it by using an indicator for unsaved state and `beforeunload` listener to prevent the loss of data.

Good idea, maybe showing a different border color or some icon for unsaved state is enough.

tophf: > update data on input

What data?

> setTimeout 

It can be argued that an indirect approach isn't better than a direct call but maybe I didn't understand what exactly you mean.

gera2ld: `data` refers to `this.jsonValue` here, which is not updated in time when the textarea is focused and save button is clicked.

`setTimeout` can be replaced with `this.$nextTick` here, to wait until reactive data is updated. In this way it does not need to call something that should not be done by it.

```js
onSave() {
  this.$nextTick(() => {
     // do stuff
  });
}
```

gera2ld:
The save button seems to do nothing indeed. 😅 The `sync` prop is never used.

src/common/ui/setting-text.vue
tophf: I've simplified the code by removing `sync`, see if you like it now. As for removing the save button, let's keep it so people don't get nervous :-) and anyway it's also useful as an obvious and understandable indicator of the modified state.

P.S. there was a problem in `master` fixed by 9fc8b6b2 - on browser startup (when "restore previous session" is selected in browser options) the Violentmonkey dashboard didn't initialize properly in the active tab.

gera2ld:
It should work but not quite Vue-flavor. I'll merge it now and try to see if I can polish it later.

PR #1116, opened 2020-11-25
fix: auto save only if there is no save button

gera2ld:
- make code a little more declarative
- change saving behavior a bit to make it more intuitive: reload page before clicking *Save* should discard unsaved changes

### DIFF
--- a/src/common/ui/setting-text.vue
+++ b/src/common/ui/setting-text.vue
@@ -2,11 +2,11 @@
   <div>
     <textarea
       class="monospace-font"
-      :class="{'has-error': error}"
+      :class="{'has-error': parsedData.error}"
       spellcheck="false"
       v-model="value"
       :disabled="disabled"
-      :title="error"
+      :title="parsedData.error"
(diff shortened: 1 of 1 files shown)


tophf:
I think this is a pretty big change in behavior so to prevent the inadvertent loss of data you should add `beforeunload` event handler.

gera2ld:
@tophf OK. It was a surprise to me that the save button did nothing. 😅

tophf:
Yeah it was a side-effect. Still, I'm worried that many people may be relying on this auto-saving...

gera2ld:
There are quite a few changes to apply the `beforeunload` sentry. I'll start a new PR for that.

tophf:
This is why I wanted to keep the old behavior, to avoid all that :-)

gera2ld:
But the old behavior is weird, isn't it?
Besides, I think it's reasonable to have a `beforeunload` sentry module since it's common and useful.

tophf (approved):

PR #1117, opened 2020-11-26
fix: extract a common unload sentry

gera2ld:
- move `showMessage` and `showConfirmation` to `common/ui` as they don't depend on page context
- add a common `getUnloadSentry` method to handle `beforeunload` event
- replace `router.pinned` with `router.confirmChange`, and remove `popstate` listener

So the workflow becomes:
- Set `router.confirmChange` when there are unsaved modifications and unset it when changes are gone.
- When `hash` changes:
  - If `router.confirmChange` is set, restore `hash` to the previous value (the same as `router.pinned` before), and call `router.confirmChange`.
    `router.confirmChange` should show a popup and decide whether to update `route` according to user's response.
  - Otherwise just update `route`.


src/common/ui/setting-text.vue
```diff
@@ -62,6 +63,11 @@ export default {
       return !deepEqual(this.parsedData.value, this.defaultValue || '');
     },
   },
+  watch: {
+    canSave(state) {
+      this.toggleUnloadSentry(state);
```
tophf, line 68: This should be relaxed a bit to allow the sentry when there's an error. Use case: I type something in the json options which is invalid until I properly finish it e.g. `"cursorBlinkRate": 100,`

tophf (commented):
Great! Just one glitch.

src/common/ui/setting-text.vue
gera2ld: done

tophf (approved):

PR #1128, opened 2020-12-15
feat: differentiate regexps from strings + tweak colors

### LINKED ISSUE #1127: Why colors of comments, regexp and strings are now very similar?
Why colors of comments, regexp and strings are now very similar?
изображение
**Can I change color theme?**

tophf:
Fixes #1127.

* greenish string colors
* desaturated comments
* regexps have their own color different from template strings

This is how it looks now:

image

image

### DIFF
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -281,6 +281,16 @@ export default {
       Object.assign(CodeMirror.keyMap.sublime, {
         'Shift-Ctrl-/': 'commentSelection',
       });
+      // Differentiate regexps and templates, TODO: remove when implemented in CodeMirror
+      const tokenizer = cm.doc.mode.token;
+      Object.assign(cm.doc.mode, {
+        token(stream, state) {
+          const res = this::tokenizer(stream, state);
+          return res === 'string-2' && state.jsState.lastType === 'regexp'
+            ? 'string-2 regexp'
+            : res;
+        },
+      });
       cm.on('keyHandled', (_cm, _name, e) => {
         e.stopPropagation();
       });
@@ -570,19 +580,6 @@ $selectionDarkBg: rgba(73, 72, 62, .99);
   z-index: 9999;
 }
 
-@media (prefers-color-scheme: dark) {
-  .CodeMirror-hints {
-    background: var(--bg);
-  }
-  .CodeMirror-hint {
-    color: var(--fg);
-  }
-  li.CodeMirror-hint-active {
-    background: var(--fg);
-    color: var(--bg);
-  }
-}
-
 /* fix contenteditable selection color bug */
 .CodeMirror .CodeMirror-line {
   ::selection {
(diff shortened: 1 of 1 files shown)


gera2ld:
Actually I think we shouldn't put much effort into this color thing. We'd better include several themes we like and let users choose, and add a better way to define custom theme. Otherwise there will always be someone asking for different styles.

Here is what I think we can do:

- remove any theme previously included
- add several picked themes as presets that can be easily switched to
- add a theme setting similar to custom style, showing the theme CSS applied, and can be easily tweaked by the user without handling CSS priorities
- the theme CSS will be applied dynamically just as custom styles

tophf:
> put much effort into this color thing

The amount of effort currently put in this PR is trivial, its aim is to fix a couple of objective deficiencies in the default theme, and if there'll be any further tweaks they'll be just as trivial or even simpler, but what you suggest sounds to me 100 times more complicated:

> remove any theme previously included

We already removed it (`eclipse`)

> add several picked themes 

Which ones and who will choose them? There's 60 themes and I expect at least half of them appeal to someone.

>add a theme setting similar to custom style, showing the theme CSS applied, and can be easily tweaked by the user without handling CSS priorities

Sounds like a very complex thing. Currently changing a theme is as easy as copypasting CSS, there's no need to handle any "CSS priorities".

gera2ld:
OK, it does seem to be over complexed.

> there's no need to handle any "CSS priorities"

Actually there is. I just figured out that the dark mode color definitions prevented me from switching themes.

tophf:
Wow, our dark mode definitions don't have a theme specifier! I'll add it here.

tophf:
Done.

I guess I'll try to suggest my changes to CodeMirror and if they accept then I'll remove them from the PR...

tophf:
Oh no, I won't deal with CodeMirror upstream because it has 120 language modes so in order to add regexp detection there I would have to deal with every mode that has regexps. I would also have to patch 63 themes...

So do we switch back to `eclipse`? Accept this PR's tweaks for some colors? Tweak them a bit?

gera2ld:
> So do we switch back to `eclipse`?

Why `eclipse`? I think it's not a problem if we can switch themes easily.
ps. I like `material-darker` more.

tophf:
Because Violentmonkey has been using `eclipse` for a long time.

tophf:
Hmm, the default theme is almost the same as `eclipse` - the most prominent difference is that `eclipse` has keywords in bold, see https://codemirror.net/demo/theme.html#eclipse. I guess we'll keep the default theme for the light scheme because its string colors are better differentiated (they're blue in `eclipse` same as identifiers).

The PR now overrides only two colors (gray comments, olive template literals) and adds orange regexps. The rest of the changes is the fix for the wrong dark scheme specificity.

tophf:
Regarding switching of the themes, it's a separate concept outside of the scope of this PR. Personally I don't mind if we include all CodeMirror themes or automatically download them when `theme` keyword is detected in json config (not sure we should add a visual selector though but I don't mind that either).

gera2ld:
LGTM.

There is no need to include all themes as most of them will be useless.

PR #1131, opened 2020-12-16
fix: auto-close .user.js tab when installer closes

### LINKED ISSUE #1129: Tabs created by dragged userscript files do not close
### What is the problem?
Tabs that the browser creates when a .user.js (userscript) file is dragged into it do not close.

### How to reproduce it?
1. Drag a .user.js file into the browser.

The tab created by dragging the .user.js file closes.

The tab created by dragging the .user.js file doesn't close. The 'Installing script' dialog opens in another tab.

The same as above. The latest release (v2.12.8) is being used.

### Environment
- Browser: Firefox
- Browser version: 83.0
- Violentmonkey version: v2.12.8
- OS: Windows

tophf:
Fixes #1129.

### DIFF
--- a/src/confirm/views/app.vue
+++ b/src/confirm/views/app.vue
@@ -14,7 +14,9 @@
               <span class="ml-1" v-text="i18n('installOptionClose')"></span>
             </label>
             <label>
-              <setting-check name="trackLocalFile" :disabled="closeAfterInstall" />
+              <setting-check name="trackLocalFile"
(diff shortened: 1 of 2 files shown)


tophf:
Hmm, gonna rework this a bit...

tophf:
Additionally, the tracking option toggles the actual tracking even after confirming the install whereas previously it did nothing when checked afterwards.

PR #1136, opened 2020-12-17
Import/export tweaks

### LINKED ISSUE #1095: [feature request] option for importing zipped scripts "don't overwrite the current GM_setValue values"
I think it would be useful to add an option for importing zipped scripts titled "Don't read the data saved by the script (don't overwrite the current GM_setValue values)" .
Thanks.

tophf:
* "[x] import script data" option - fixes #1095.
* download via Blob in modern FF

### DIFF
--- a/src/common/download.js
+++ b/src/common/download.js
@@ -1,14 +1,9 @@
 export function downloadUrl(url, name, callback) {
   const a = document.createElement('a');
-  a.style.display = 'none';
-  document.body.appendChild(a);
   a.href = url;
-  if (name) a.download = name;
-  a.click();
-  setTimeout(() => {
-    document.body.removeChild(a);
-    if (callback) callback();
-  }, 3000);
+  a.download = name || '';
+  a.dispatchEvent(new MouseEvent('click'));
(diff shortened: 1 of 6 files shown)


src/common/download.js
```diff
@@ -1,14 +1,9 @@
...
-  document.body.appendChild(a);
   a.href = url;
-  if (name) a.download = name;
-  a.click();
-  setTimeout(() => {
-    document.body.removeChild(a);
-    if (callback) callback();
-  }, 3000);
+  a.download = name || '';
+  a.dispatchEvent(new MouseEvent('click'));
```
gera2ld, line 5: Is there any difference between `a.click()` and `a.dispatchEvent(new MouseEvent('click'));`?

tophf: Yep, click() requires adding to DOM while dispatchEvent works in detached node.

tophf: Should I add it as a comment?

gera2ld: It's OK, we can leave it as is.

Issue #1137, opened 2020-12-17
Add New Syntax Option for Latest @run-at document-start Behavior


nyfair:
The feature is great, but it also break compatibility of many scripts that execute at document-start and do some DOM modification.

tophf:
Without a demo script that reproduces the problem there's nothing we can do here.

nyfair:
Are u kidding? this is fairly simple, just use document.querySelector to see the error message.
You can not persuade all the script contributors to update their scripts, right?

tophf:
I'm not kidding. Without a demo script I can only guess and the only guess I have so far is that you misunderstood what `document-start` means.

nyfair:
Yes, you are right. Let me say again, you can't persuade all the script contributors to update their scripts just because you are right.

tophf:
It's not about me. It's about how `document-start` works and always worked for the last 20 years. You need to provide an example script that shows the problem, otherwise there's nothing we will do.

nyfair:
And 20 years is much longer than 2days since the latest release. These scripts already existed for years.

tophf:
We'll wait for someone who is reasonable to show us at least one script. I'm locking this issue for now.

gera2ld:
So many useless messages without any snippet or a single link to one of your *many scripts*...

You shouldn't be able to `querySelector` at `document-start` by design. Maybe you can try `document-body`.
Please read the docs for more details.

**We don't patch for wrong use cases.**

PR #1142, opened 2020-12-18
implement run-at: document-body

tophf:
* implement `@run-at document-body`, related: #1137
* add the missing space in labelRunAt for letter-based languages

1. How do I/we update files on transifex for this change? I can do it manually I guess but I'd prefer to upload in one click.
2. Maybe we could instead just check for unicode range of the message in `i18n` and add the space after ASCII colon `:` automatically?

### DIFF
--- a/src/_locales/de/messages.yml
+++ b/src/_locales/de/messages.yml
@@ -408,7 +408,7 @@ labelReplace:
   message: 'Ersetzen mit: '
 labelRunAt:
   description: Label of script @run-at properties in custom meta data.
-  message: 'Ausführen bei (Run-At):'
+  message: 'Ausführen bei (Run-At): '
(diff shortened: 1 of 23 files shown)


gera2ld:
> How do I/we update files on transifex for this change?

They should be uploaded automatically if this PR is labeled with `i18n` when this PR is merged. But we need to make sure the current translations are up to date. I'll trigger an update manually first.

> Maybe we could instead just check for unicode range of the message in `i18n` and add the space after ASCII colon `:` automatically?

That's kind of tricky, we'd better not.

gera2ld (approved):
LGTM

tophf:
Oh, I forgot to add it to `content` mode... Will open a new PR soon.

PR #1147, opened 2020-12-22
feat: show extras menu in popup (remove + homepage)

### LINKED ISSUE #1000: Provide a way to quickly delete a script in using
### What is the problem?

Currently, if you want to delete a script that is used on the current website, you have to:

1. Look and memorize the name of the script;
2. Go to the dashboard;
3. Search for the name manually;
4. Remove it from the dashboard.

I think if we provide either:

1. a remove button in the dropdown list
2. a remove/delete button in editor

This process can be shortened a lot, and no memorizing needed. 

Instead of go to the dashboard and search, one can directly remove it without leaving the website, or alternatively, open that script in editor by right click (or click edit button) and then delete it from there.

### Environment
- Browser: Chrome
- Browser version: 83
- OS: Windows

tophf:
Fixes #1000.

* Homepage if present
* Remove (or Restore)

I wonder if showing the version number at the top of the new menu would make sense? As for update check, it would require quite a few changes apparently so ain't gonna add it here.

### DIFF
--- a/src/popup/style.css
+++ b/src/popup/style.css
@@ -184,7 +184,7 @@ footer {
   border-top: 1px dashed var(--fill-2);
   > * {
     position: relative;
-    &:not(:hover) > .submenu-buttons {
+    &:not(:hover):not(.extras-shown) > .submenu-buttons {
       visibility: hidden;
     }
     .menu-item {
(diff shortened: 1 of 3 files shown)


gera2ld:
Looks great!

Nitpick: menus for the last few items lead to resizing of the container, not a big problem though.

tophf:
I've switched the menu to a single instance so it can cover the footer.

PR #1158, opened 2020-12-30
fix: show warning if document-start is used improperly

gera2ld:
## Limitations

- Only direct usage of `document.head` can be detected, so `var doc = document; doc.head` will be ignored.
- The referred documentation has no enough information on this ATM.

### DIFF
--- a/src/injected/web/index.js
+++ b/src/injected/web/index.js
@@ -104,7 +104,13 @@ async function onCodeSet(item, fn) {
   if (process.env.DEBUG) {
     log('info', [bridge.mode], item.custom.name || item.meta.name || item.props.id);
   }
-  const run = () => wrapGM(item)::fn(logging.error);
+  const handleError = err => {
+    logging.error(err);
+    if (!stage && err?.name === 'TypeError' && err.message === 'document.head is null' && document.head === null) {
+      logging.warn('👆 This script may be using `document-start` improperly, if you are the script author, please read https://violentmonkey.github.io/api/metadata-block/#run-at for more details');
+    }
+  };
+  const run = () => wrapGM(item)::fn(handleError);
   const el = document::getCurrentScript();
   const wait = waiters[stage];
   if (el) el::remove();


tophf:
This won't work in Chrome though as its message is `Cannot read property '......' of null`...

gera2ld:
By now only Firefox supports `browser.contentScripts.register` so this issue is likely to appear only in Firefox. I did some quick tests and could not reproduce it in Chrome.

`Cannot read property '......' of null` is too general, I think it may cause false positive warnings. If checking for Chrome is necessary some day, I guess it's better to proxy `document` and see if `document.head` is read just before the error occurs.

tophf:
It happens in Chrome too, quoting the release info:

> Enabled on pages that open quickly: super fast sites, back-forward navigation, reopening of a closed tab.

tophf:
I'm not yet convinced we should be adding a warning because there might be just a handful of such incorrect scripts.

gera2ld:
Ah, I must be missing something in Chrome.

There would be many issues complaining about this since many authors don't read docs and users don't know what the expected behavior is, and the worst thing is that they work in tampermonkey but not in violentmonkey.
I hope users can know what is happening after reading the warning.

gera2ld:
I think it's OK to add a simple warning but if things become over complexed it would be better to pin a issue on this.

tophf:
Thing is, such incorrect scripts will also break in Tampermonkey when "instant" injection mode is enabled. This mode is relatively popular because it's used by various anti-ad scripts so I don't think there's a lot of scripts that use document-start and expect document.head to exist.

gera2ld:
I'm convinced, closing this.

tophf:
Greasyfork lists only [19 scripts using `document-start`](https://greasyfork.org/en/scripts/code-search?c=document-start). Out of ~100K it has overall. I've checked the source code of all these 19 scripts and there isn't a single one that uses document.head incorrectly.

PR #1166, opened 2021-01-12
fix: clear preinject cache for more db changes

tophf:
Now it checks everything (hopefully) that affects the preinject cache: autoupdates, requires, resources, values.


gera2ld (approved):
LGTM

PR #1177, opened 2021-01-26
ui: align editor settings

tophf:
+ align editor settings
+ split into multiple columns on wide displays
+ auto-fit textarea height
+ remove tooltips on textareas because their labels already indicate `@key`

chrome_2021-01-27_01-07-06

### DIFF
--- a/src/options/views/edit/settings.vue
+++ b/src/options/views/edit/settings.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="edit-settings" ref="container">
     <h4 v-text="i18n('editLabelSettings')"></h4>
-    <div class="form-group">
+    <div class="form-group condensed">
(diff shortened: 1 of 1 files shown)


src/options/views/edit/settings.vue
```diff
@@ -129,35 +97,56 @@ export default {
       }
     },
   },
+  methods: {
+    countRows(name) {
+      return 2 + (this.settings.custom?.[name]?.match(/\n/g)?.length || 0);
```
gera2ld, line 102: This does not work on long lines that wrap. Maybe it's better to use `scrollHeight` in a debounced callback? Or just leave it as is.

tophf: Fixed.

PR #1191, opened 2021-02-10
feat: show structured info in install confirm ui

### LINKED ISSUE #1121: Rework the installation dialog
Parse the metadata, analyze it, and display the important stuff above or to the left of the script code.

* `@match`, `@include` and so on
* `@resource`
* `@require`
* `@antifeature` - https://github.com/JasonBarnabe/greasyfork/commit/719e2f5042a75fda0c26410ccfdaeadc198c4584
* `@compat`
* `@connect`

Also, the `Confirm Installation` call to action button should stand out.

tophf:
Fixes #1121

Narrow window | Wide window
-|-
image | image

Clicking the `(i)` icon collapses the lists and makes their titles clickable to display only the clicked list:

image

I don't want to store this state persistently via `setting-check` because the entire point of showing the lists in the installer is to make them obvious at a glance, especially important if the script has the metablock somewhere down the code or it adds spaces to "hide" the urls.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -583,6 +583,9 @@ msgSavedScriptTemplate:
 msgScriptUpdated:
   description: Notification message for script updates.
   message: 'Script [$1] is updated!'
+msgShowHide:
+  description: Tooltip or text shown on a toggle that shows/hides stuff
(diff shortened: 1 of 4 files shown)


gera2ld:
Awesome!

PR #1199, opened 2021-02-15
fix: check preinject after promise resolved

gera2ld:
Fixes problem: Preinjects do not refresh in Firefox.

The cache for preinject contains promises but the checks do not wait for them to be resolved. As a result the checks always fail.

### DIFF
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -52,9 +52,9 @@ const propsToClear = {
   [storage.value.prefix]: 'withValueIds',
 };
 
-browser.storage.onChanged.addListener(changes => {
+browser.storage.onChanged.addListener(async changes => {
   const dbKeys = Object.keys(changes);
-  const cacheValues = cache.getValues();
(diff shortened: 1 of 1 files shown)


src/background/utils/preinject.js
```diff
@@ -153,7 +154,6 @@ function prepareScript(script, index, scripts) {
     // Firefox lists .user.js among our own content scripts so a space at start will group them
     `\n//# sourceURL=${extensionRoot}${ua.isFirefox ? '%20' : ''}${name}.user.js#${id}`,
   ].join('');
-  cache.put(dataKey, injectedCode, TIME_KEEP_DATA);
```
gera2ld, line 156: Removing this line as the cache seems to be unused.

tophf: It's used by InjectionFeedback.

gera2ld: Oh, my bad...

tophf (approved):

PR #1200, opened 2021-02-15
fix: use the db storage cache for icons in popup

tophf:
When I added icons in popup I thought reusing db cache would be complicated but turns out it's simple so here it is.

### DIFF
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -52,7 +52,7 @@ hookOptions((changes) => {
 });
 
 Object.assign(commands, {
-  /** @return {Promise<Object>} */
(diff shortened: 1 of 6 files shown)


tophf:
Hmm, fetching icons too soon seems to cause delays in some browsers :-) Investigating...

tophf:
There was a weird delay in one browser so now I set the icons in setTimeout, seems to look fine in all browsers.

tophf:
Will rework this to deduplicate all our `canvas` code.

tophf:
Reworked to deduplicate the code.

PR #1202, opened 2021-02-16
feat: add "exclude" to popup menu

### LINKED ISSUE #367: disable script in the popup for the current site
### What is the problem?
https://github.com/violentmonkey/violentmonkey/issues/366

### How to reproduce it?
1. disable single script with the popup panel

example. https://greasyfork.org/pl/scripts/30974-autoplay-disabled-for-html5-videos-pause-on-switch-tab

blocked this script per this single site

blocked this script per all sites

### Environment
- Browser: chrome
- Browser version: 64.0.3282.186
- Violentmonkey version: v2.8.29
- OS: win10

tophf:
Fixes #367.

<details><summary>Initial implementation, now rejected, see comments below</summary>

Menu | UI
-|-
image | image

</details>

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -499,6 +499,15 @@ menuMatchedScripts:
 menuNewScript:
   description: Menu item to create a new script.
   message: Create a new script
+menuExclude:
+  description: Shown in popup menu after clicking script's "..." dropdown so try to keep the label short.
+  message: Exclude...
+menuExcludeHint:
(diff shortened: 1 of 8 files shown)


gera2ld:
I think a content-rich popup inside another popup looks weird.
How about rendering the form as a submenu instead of a popup so that we don't need to care about the height and position of the popup?

tophf:
Indeed. Seems obvious now that you mention it :-) Gonna try.

tophf:
image

And this is after clicking `(i)`

image

gera2ld (approved):
Great work! LGTM.

PR #1205, opened 2021-02-20
deduplicate css

tophf:
* deduplicate `<setting-check>` by adding `label` prop and `slot`
* deduplicate classes like `ml-1` on children to a single class like `ml-1c` on parent
* ~deduplicate `flex flex-col` and `flex flex-wrap` to simply `flex-col` and `flex-wrap` by including `display:flex` in each one~

### DIFF
--- a/src/common/ui/setting-check.vue
+++ b/src/common/ui/setting-check.vue
@@ -1,5 +1,10 @@
 <template>
-  <input type="checkbox" v-model="value" :disabled="disabled">
+  <label>
+    <input type="checkbox" v-model="value" :disabled="disabled">
+    <slot>
+      <span v-text="label" />
+    </slot>
+  </label>
 </template>
 
 <script>
@@ -9,6 +14,7 @@ import hookSetting from '../hook-setting';
 export default {
   props: {
     name: String,
+    label: String,
     disabled: Boolean,
     sync: {
       type: Boolean,
--- a/src/common/ui/message.vue
+++ b/src/common/ui/message.vue
@@ -3,11 +3,10 @@
     <div class="mb-1" v-if="message.text" v-text="message.text"></div>
     <form v-if="message.buttons" @submit.prevent>
       <input class="mb-1" type="text" v-if="message.input !== false" v-model="message.input">
-      <div>
+      <div class="mr-1c">
         <button
           v-for="(button, index) in message.buttons"
           :key="index"
-          class="mr-1"
           :type="button.type || 'button'"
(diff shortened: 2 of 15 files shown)


gera2ld:
I prefer keeping separated classes, each of which is as simple as possible, as designed in [tailwindcss](https://v1.tailwindcss.com/docs/utility-first). Actually I even wanted to leverage tailwindcss to simplify CSS in this project.
In this way we can compose styles in a more accurate level, for example switching `display` between `flex` and `block` without touching other classes. But if we set `display: flex` in `flex-col`, we can no longer just remove `flex-col` to reset `flex-direction` since we need to add `display: flex` back in another way.

gera2ld:
Other changes look good to me.

tophf:
Makes sense, I've removed `flex` changes.

src/common/ui/setting-check.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-  <input type="checkbox" v-model="value" :disabled="disabled">
+  <label>
+    <input type="checkbox" v-model="value" :disabled="disabled">
+    <slot>
+      <span v-text="label" />
```
gera2ld, line 5: I think the `<span>` here can be stripped, and the component can be used like this:

```html
<setting-check v-text="label" />
```

In this way we don't need a new prop `label` either.

tophf: Yeah, looks tempting but it'll be a breaking change because it'll reduce the ability to customize these elements via custom CSS, like swapping elements via `order` trick, etc.

gera2ld: Do you mean that CSS depends on the `<span>` element? I thought we can embed a `<span>` when we need it:

```html
<setting-check><span>label</span></setting-check>
```

It's OK to always keep a `<span>` though, this is not a problem.

tophf: I didn't mean us, I meant the `custom CSS` in VM settings that some user might have used. On the other hand, this would be an extremely rare and possibly even weird thing to do so let's go with your idea.

tophf: So I tried it and it didn't work: `v-text` replaces the entire contents, not just `slot`.

gera2ld: `v-text` replaces the entire contents in `<setting-check>`, and the content will be put in `<slot>`.
Never mind, we should not break the custom CSS.

gera2ld (approved):

PR #1211, opened 2021-02-25
fix: close popup on command click

gera2ld:
I guess the popup should be closed when a command is clicked, which is the default behavior of most menus.


tophf (approved):
Yeah, I wondered about that myself since the day I switched to VM.
Tampermonkey closes it so having the same here looks reasonable.

PR #1212, opened 2021-02-25
fix: ensure content script to be unregistered

gera2ld:
In Firefox, when a content script registration is removed from our cache because of expiration, it won't be unregistered, leading to multiple registrations for a URL and script updates cannot take effect any more.
This PR adds an `onDispose` callback to explicitly dispose every cached registration when it is removed from cache.


tophf (approved):
I've been always obsessed with performance so I'd keep `cache = {}` when `onDispose` is not provided. The difference is negligible though, deleting 1000 keys takes just 0.1ms, but I kinda feel bad for CPU :-)

gera2ld:
Update: delete keys one by one only if `onDispose` is provided.

Issue #1221, opened 2021-03-08
[feature request] List of scripts


wvxwxvw:
Please return the sequential display of the scripts in the script list - an ordinal list. Who even got the idea that tiles are good? We have to get rid of this "love of lamers" in Explorer, and now here too. Since you like tiles so much, make them as an option. It is inconvenient to search for the required script among several dozen.

tophf:
It depends on the width of the screen. On a typical 1920x1080 screen the single column mode wastes 90% of space, which is why the multi-column mode was introduced. It may be unusual for some people, of course, but there's nothing inherently wrong with multi-column display because it's being the default in many modern apps and sites.

You can restore a single column using this custom CSS:

```css
.scripts {
  display: block !important;
}
.scripts .script {
  width: auto !important;
}
```

wvxwxvw:
Thanks.

Issue #1210, opened 2021-02-24
[feature request] Table view in Installed Scripts


darkred:
My suggestion is to be able to view the installed scripts as a *table* .

&nbsp;
Indeed, VM, compared to TM, currently displays more scripts in a screenful at max *(see screens)* :
VM --> 36 scripts (3 columns x 12 rows)
TM --> 29 scripts (as table list),
but, I find that: 
- VM's view is not compact enough , as each entry box still has some margin from its next, and 
- most importantly, if you sort the scripts alphabetically,
with TM the resulted list can be read as any table, i.e. each script entry after another, 
A
B
C
while with VM each displayed row has 3 sorted script entries, e.g.
A B C (1st row)
D E F (2st row)
etc.
thus making it not very easy to read.

To clarify, making VM display the entries sorted vertically, e.g.
A D
B D
C F
wouldn't help anything, because if you have a lot of scripts, the sorted "column" A>B>C would continue below the viewport.

I personally prefer table view in general, because I find it more compact, and easier to read.

&nbsp;
Here is a screenshot comparison: 
<sup>*(using Chrome 88 and latest VM/TM betas)*</sup>

<details>
<summary>VM (alphabet. sort)</summary>

![](https://user-images.githubusercontent.com/723651/109031504-76e4a000-76cd-11eb-9e85-c587399f44bd.jpg)
</details>

<details>
<summary>TM (alphabet. sort)</summary>

![](https://user-images.githubusercontent.com/723651/109031904-d3e05600-76cd-11eb-9a6b-8a58a2d24fa1.jpg)
</details>

<details>
<summary>TM (default sort)</summary>

![1](https://user-images.githubusercontent.com/723651/109032167-130ea700-76ce-11eb-8988-2acc6986709a.jpg)
</details>

Thank you

tophf:
We will rethink the UI and ways to organize scripts in #287. Until then you can write your own CSS and put it in Violentmonkey's advanced settings in "Custom CSS" section, you can also publish it as a repo or a gist on github. If there will be many such themes we may even compile a list of them in Wiki here and link to it in VM settings.

darkred:
Thanks for your response.

darkred:
Thank you for implementing this, @tophf !

Issue #1135, opened 2020-12-16
(Suggestion) In the dashboard, move the `Sort by` dropdown menu outside Filters, and then remove `Filters` itself


darkred:
My suggestion is In the dashboard, to move the `Sort by` dropdown menu outside 'Filters', and then remove `Filters` itself.

The reason is that the word 'Filters' implies using criteria to look for matching subgroups in the installed scripts, while it actually only contains the irrelevant "Sort by" functionality.

Even if you plan to add more functionality inside Filters, I believe that "Sort by" is better separated.

Thank you

tophf:
You can adjust it via CSS, here's mine:

![image](https://user-images.githubusercontent.com/1310400/102399567-d6e60880-3ff1-11eb-8cbf-229b8a7c3ef3.png)

I doubt this option is used a lot so it's better to keep it hidden. We could rename the tooltip to `Filters and sort` but somehow this doesn't seem worth the effort.

darkred:
Could you please share your CSS?

tophf:
It's in https://gist.github.com/tophf/ba26f98902ec54ffde4968c9c8b01eb2. There's a lot more than filters so delete the rest.

darkred:
Thank you!

tophf:
I expect this part of the UI will be [completely] reworked when we implement #287. Until then CustomCSS is an acceptable compromise (unless we see more reports about this particular problem of course).

PR #1222, opened 2021-03-08
feat: options for table view and single-column mode

tophf:
Addresses #1135 #1210 #1221.

I didn't want to implement it initially because it seemed like a 100% CSS customization but now I kinda do.

* The table layout needs a clickable name because a) this is how such layout usually works and b) the edit button is too far from the script name. Without supporting this event internally the CSS has to stretch the edit button over the name button which is a dirty hack.
*  VM isn't Stylus so it doesn't have a quick way to toggle the style.
* A user might want to switch modes for a quick overview of all the scripts at once, then switch back.

### 1-column table

image

### 2-columns table

image

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -465,6 +465,12 @@ labelTranslator:
 labelUpdateURL:
   description: Label of script @updateURL in custom meta data.
   message: 'Update URL:'
+labelViewSingleColumn:
+  description: Label for option in dashboard script list to show the scripts in single column.
+  message: Single column
+labelViewTable:
+  description: Label for option in dashboard script list to show the scripts as a table.
+  message: Table view
 lastSync:
   description: Label for last sync timestamp.
   message: Last sync at $1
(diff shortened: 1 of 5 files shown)


gera2ld (approved):
Looks great!

tophf:
@gera2ld which mode do you think should by the default one - table or cards? I wanted to keep cards by default but then accidentally switched it to table so maybe I need to flip it...

gera2ld:
I guess it's better to keep cards by default as before, less surprise to users.

tophf:
Done: 44d5290f

T1mL3arn:
I personally find 1300px as first breakpoint is too big. For me 2-column mode on 1280px-wide monitor is good.

Here is the css to change first breakpoint

```css
:root {
  --columns-cards: 1280,1900,2500 !important;
}
```

tophf:
Too cramped.

tophf:
Although it depends on how long are the names. For short names we could probably show two columns. Actually it makes sense to simply allow the user to configure the minimum width of a column.

PR #1230, opened 2021-03-16
fix: group multiple update notifications in one

### LINKED ISSUE #1228: [UX/UI] Too many notifications when there are a lot of errors updating scripts
### What is the problem?

Sometimes (I have no idea why and when, probably my ISP is the reason) VM fails to update all the installed script. And with this it shows a lot of notifications like this:

image

These notifications occupy the whole vertical space (cause there are many of them), and if some links to the script are long - they occupy a lot of horizontal space too, making around of 40% browser window hidden.

### How to reproduce it?

I guess you need to call a lot of notifications https://github.com/violentmonkey/violentmonkey/blob/e8177755f1af05a2d47ac5e1eb8a51f6c66be0ad/src/background/utils/message.js#L5 at once (let's say 30) 

The notifications are shown with delay, one by one (or small groups 2-3 message each), not blowing the screen. Or maybe VM should not update (when it does it automatically) all the scripts at once.

Didn't test

### Environment
- Browser: Firefox 86.0.1 (64-bit)
- Violentmonkey version: v2.12.11
- OS: Win 8.1 x64

tophf:
Fixes #1228.

### DIFF
--- a/src/background/utils/notifications.js
+++ b/src/background/utils/notifications.js
@@ -1,20 +1,19 @@
-import { defaultImage, sendTabCmd } from '#/common';
+import { i18n, defaultImage, sendTabCmd } from '#/common';
+import ua from '#/common/ua';
 import { commands } from './message';
 
 const openers = {};
 
 Object.assign(commands, {
   /** @return {Promise<string>} */
-  async Notification(data, src, bgExtras = {}) {
-    const { items, onClick, type } = bgExtras;
+  async Notification(data, src, bgExtras) {
     const notificationId = await browser.notifications.create({
-      type: type || 'basic',
-      title: data.title || '',
+      type: 'basic',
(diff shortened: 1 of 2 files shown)


src/background/utils/notifications.js
```diff
@@ -1,18 +1,20 @@
...
 Object.assign(commands, {
   /** @return {Promise<string>} */
-  async Notification(data, src, bgExtras) {
+  async Notification(data, src, bgExtras = {}) {
+    const { items, onClick, type } = bgExtras;
     const notificationId = await browser.notifications.create({
-      type: 'basic',
-      title: data.title || i18n('extName'),
+      type: type || 'basic',
+      title: data.title || '',
```
tophf, line 12: I removed the default title because Chrome/FF already displays "Violentmonkey" in the non-customizable header of the notification.

gera2ld (approved):

PR #1232, opened 2021-03-20
fix: use "notify updates" option for errors too

tophf:


gera2ld:
That's quick.
I am thinking about a new option for error notifications, in case a user disables update notices but expects error prompts.

tophf:
Well, VM never showed them in the past so I thought it's not that useful or important for a separate option...

gera2ld (approved):
OK

PR #1234, opened 2021-03-22
fix: send chunks exactly once

### LINKED ISSUE #1229: large json parse failed
### What is the problem?
when xhr a large json file with `responseType:'json'` and `onprogress` or other, get response `SyntaxError: Unexpected end of JSON input`

### How to reproduce it?
1. install this script
```js
// ==UserScript==
// @name  New script - github.com
// @match       https://www.google.com/*
// ==/UserScript==

 GM_xmlhttpRequest({
  // url:'https://graph.facebook.com/github', // => res.response succeed
  url:'https://api.github.com/gists/public', // => SyntaxError
  method: 'GET',
  responseType: 'json',
  onabort (res) {
    // debugger;
  },
  onerror (res) {
    // debugger;
  },
  onload (res) {
    debugger;
    console.log(JSON.parse(res.responseText))
    console.log(res.response)
  },
  onprogress (res) {
    // debugger;
  },
  onreadystatechange (res) {
    // debugger;
  },
  ontimeout (res) {
    // debugger;
  }
});
```
2. xhr small json succeed
3. or remove other listener except onload succeed

parse json succeed

get SyntaxError

all below test in latest version v2.12.11

### Environment
- Browser: Chrome
- Browser version: 86.0.4240.198
- Violentmonkey version: 2.12.11
- OS: win10 x64

gera2ld:
Fix #1229 indeed.

In #1068 the `xhr.response` is sent multiple times when `progress` event is listened, and a partial response is cached in `raw`, leading to incomplete `req.response` and parse error for `json` type.
In this PR the response is ensured to be sent exactly once and concatenated as `req.raw` for all content types except `text`, including `document` and `json`.
I think `req.raw` is not necessary for `text`, `json` and `document` types as they are all regarded as `text` in the background, but we can make use of it to allow data to be handled by the same function `parseData`. Maybe we should handle all data as blob or arraybuffer in the background to support super large responses.

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -159,7 +159,7 @@ function xhrCallbackWrapper(req) {
   let response;
   let responseText;
   let responseHeaders;
-  const sent = new Set([undefined]);
+  let sent = false;
   const { id, chunkType, xhr } = req;
   // Chrome encodes messages to UTF8 so they can grow up to 4x but 64MB is the message size limit
   const chunkSize = 64e6 / 4;
@@ -201,7 +201,8 @@ function xhrCallbackWrapper(req) {
       }
     }
     const shouldNotify = req.eventsToNotify.includes(type);
-    const shouldSendResponse = shouldNotify && !sent.has(response);
+    // only send response when XHR is complete
+    const shouldSendResponse = xhr.readyState === 4 && shouldNotify && !sent;
     chainedCallback({
       contentType,
       id,
@@ -220,12 +221,12 @@ function xhrCallbackWrapper(req) {
       },
     });
     if (shouldSendResponse) {
-      sent.add(response);
+      sent = true;
       for (let i = 1; i < numChunks; i += 1) {
         chainedCallback({
           id,
+          chunkIndex: i,
           chunk: getChunk(i),
-          isLastChunk: i === numChunks - 1,
         });
       }
     }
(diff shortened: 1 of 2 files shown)


tophf (approved):
>we can make use of it to allow data to be handled by the same function parseData. 

Yes, the purpose of `raw` was to perform lazy parsing.

>handle all data as blob or arraybuffer in the background to support super large responses.

It would needlessly stress the CPU for responses smaller the threshold.

gera2ld:
> It would needlessly stress the CPU for responses smaller the threshold.

So we don't support super large text response now?

I did some more test and found that 18a32fe might also work, just that the `responseText` needs to be parsed. In that case the `response` is buggy but we are using `responseText` instead so it does not really matter.

tophf:
> we don't support super large text response now?

We do. My objection was about using this mode for small responses.

tophf:
Er, wait, actually I never tried a super large text mode response so I guess it doesn't work. So the chunked mode may be necessary if responseText.length > chunkSize.

gera2ld:
`numChunks` seems to be set only for non-text types so I guess it will not work for text responses.

https://github.com/violentmonkey/violentmonkey/blob/caa4e144a14064dc0fb8919a08df1d4cb22a9c71/src/background/utils/requests.js#L199-L201

PR #1235, opened 2021-03-23
fix: abort GM_xmlhttpRequest on page redirect

gera2ld:
When tab closed or page reloaded, the xhr won't stop. This could be waste of time and bandwidth when requesting large data.

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -9,17 +9,28 @@ import { commands } from './message';
 const VM_VERIFY = 'VM-Verify';
 const requests = {};
 const verify = {};
+const tabRequests = {};
 
 Object.assign(commands, {
   ConfirmInstall: confirmInstall,
   /** @return {string} */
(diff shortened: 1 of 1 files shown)


tophf:
`Local YouTube Downloader` opens a new window where it tracks the download so it's quite possible some users may close the original tab while waiting but this PR will make it impossible. I guess, to track the opened tabs you would have to intercept `TabOpen` command, `window.open` in the wrapper, `window.open` in the page context because userscripts may inject the code into the page too...

gera2ld:
How can a page opened via `window.open` receive GM_xmlhttpRequest callbacks? If the page is closed and callbacks are discarded, I guess the XHR becomes useless too.

I tested the script and it seems not to be the case. The new windows is just UI while the logic stays in the original page. If the original page is closed, it stops downloading anyway.

tophf:
Indeed, my bad.
LGTM.

PR #1243, opened 2021-04-05
feat: add keyboard support

### LINKED ISSUE #1238: [Feature] Keyboard navigation for Popup menu
### What is the problem?
It is possible to assign keyboard shortcut for opening the popup menu, but there is no way to further interact with elements in there.

### How to reproduce it?
0. (Assumed browser: Firefox, used in Dev edition.)
1. Visit `about:extensions` > ⚙ Manage Extensions Shortcuts > Violentmonkey > Activate toolbar button: (set shortcut).
2. Invoke shortcut set in step 1.
3. Try to navigate trough popup's elements using keyboard (intuitively <kbd>Tab</kbd>, <kbd>Shift</kbd><kbd>Tab</kbd> or alternatively arrow keys) and trigger primary action of any element in there, e.g. Disable Scripts toggle that seem to be the first in assumed tab order.

Disable Scripts toggle is highlighted after pressing <kbd>Tab</kbd> and invoked with <kbd>Enter</kbd> or <kbd>Spacebar</kbd>, like native checkbox.

Neither <kbd>Tab</kbd>/ arrow keys, <kbd>Enter</kbd> nor <kbd>Spacebar</kbd> keys do anything.  It is only possible to close the popup with <kbd>Escape</kbd> and with the assigned shortcut.

### Environment
- Browser: Firefox
- Browser version: 88.0b3 20210325185929
- Violentmonkey version: 2.12.12
- OS: Windows 10

### Notes
(I'm not fluent in Vue, so apologies for potential misinterpretation. Plus I'm quite sure this insight will be unnecessary for most devs in here, but I feel it should be told:) Looking into https://github.com/violentmonkey/violentmonkey/blob/master/src/popup/views/app.vue it seems most interactive elements seemingly functioning like button or checkbox are actually implemented as `<div>`s or `<tooltip>`s (from "vueleton") with attached `@click` handlers.  Using "divs instead of buttons" is commonly considered a very bad approach by those with at least marginal interest in accessibility and good coding practices in general.  I don't know whether used libraries or Vue itself does a bad job in re-using listeners for keyboard-initiated events or does not even try (I'd not be surprised; recreating accessible interactive element from scratch is usually lots of work).
Normally using native HTML `<button>` or `<input type="checkbox">`  produces focusable elements that fires click handlers on spacebar or enter key press.  Such approach is AFAIK generally accepted as a best practice.

Looking at Stylus extension, it uses this approach (https://github.com/openstyles/stylus/blob/master/popup.html), so the effect is you don't necessarily need a mouse for using it at all; I'd be happy to see same approach in Violentmonkey. In fact I'm quite surprised to see such differences considering both extensions share core dev(s) (?))

BTW Dashboard page could use some focus styles for those elements that are already "native HTML", and minor accessibility overhaul, but at least one can navigate among installed scripts with arrow keys in there, but it's probably topic for other issue.

gera2ld:
## Features

Now we can do almost everything without a mouse.

Closes #1238 

### Options page

- Vim-like shortcuts
  - [x] <kbd>h</kbd>/<kbd>j</kbd>/<kbd>k</kbd>/<kbd>l</kbd> move around
  - [x] <kbd>g</kbd> <kbd>g</kbd> jump to the first script
  - [x] <kbd>G</kbd> jump to the last script
  - [x] <kbd>/</kbd> search scripts
- also shortcuts with modifiers
  - [x] <kbd>cmd</kbd>/<kbd>ctrl</kbd> + <kbd>f</kbd> search scripts
  - [x] <kbd>cmd</kbd>/<kbd>ctrl</kbd> + <kbd>up</kbd>/<kbd>down</kbd>/<kbd>left</kbd>/<kbd>right</kbd> move around
  - [x] <kbd>alt</kbd>+<kbd>pgup</kbd>/<kbd>pgdn</kbd> switch to settings / about tabs
- handle focused script
   - [x] <kbd>?</kbd> show/hide hints on focused script
   - [x] <kbd>e</kbd> edit script
   - [x] <kbd>space</kbd> toggle script
   - [x] [<kbd>shift</kbd>+]<kbd>tab</kbd> jump between actions

Kapture 2021-04-05 at 23 19 20

### Popup page

- navigate with [<kbd>shift</kbd>+]<kbd>tab</kbd> / <kbd>↑</kbd> / <kbd>↓</kbd> / <kbd>←</kbd> / <kbd>→</kbd>
- press <kbd>enter</kbd> to click

## Implementation

A keyboard service inspired by VSCode is introduced to handle shortcuts based on realms and conditions.
Each shortcut can be bound to a condition, and only takes effect when the condition is true. Only the callback of the first shortcut matched will be called.


tophf:
* What about `Tab` and `Shift-Tab` to access all controls in the active script's card? Or maybe add more keys to invoke them directly?
* Switching to `Settings` and `About` - what about using the same key from the editor `Alt-PgUp` and `Alt-PgDn`?

gera2ld:
>     * What about `Tab` and `Shift-Tab` to access all controls in the active script's card?

Done. But there is [a terrible known issue](https://stackoverflow.com/a/11713537/4238335) in Firefox on macOS.

>     * Or maybe add more keys to invoke them directly?

This is considered but not implemented until the tooltips can handle keyboard events properly.

src/common/ui/smooth-scroll.js
```diff
@@ -0,0 +1,17 @@
+export function createSmoothScroll(el) {
```
tophf, line 1: What about the [built-in smooth scroll](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll)?

```js
element.scroll({
  top: 100,
  left: 100,
  behavior: 'smooth'
});
```

gera2ld: That's much better 😅

tophf:
There was a merge conflict so I've merged the master branch but maybe it'd be better if you rebase on the new master yourself...

PR #1247, opened 2021-04-08
Don’t disable Server URL for anonymous logins

da2x:
Only disable the username and password fields.

### DIFF
--- a/src/options/views/tab-settings/vm-sync.vue
+++ b/src/options/views/tab-settings/vm-sync.vue
@@ -22,9 +22,9 @@
       <label class="sync-server-url">
         <span v-text="i18n('labelSyncServerUrl')"></span>
         <input
-          type="text"
+          type="url"
           v-model="state.userConfig.serverUrl"
-          :disabled="!state.canAuthorize || state.userConfig.anonymous"
+          :disabled="!state.canAuthorize"
         />
       </label>
       <div class="mr-2c">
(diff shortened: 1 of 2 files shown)


src/options/views/tab-settings/vm-sync.vue
```diff
@@ -22,9 +22,9 @@
       <label class="sync-server-url">
         <span v-text="i18n('labelSyncServerUrl')"></span>
         <input
-          type="text"
+          type="url"
```
tophf, line 25: You also need to add `input[type=url]` in our [global css](https://github.com/violentmonkey/violentmonkey/blob/master/src/common/ui/style/style.css) in three places where `input[type=text]` is currently used.

tophf (changes requested):

src/options/views/tab-settings/vm-sync.vue
da2x: Done.

tophf (approved):

tophf:
Regarding `type=url`: actually, I'm not sure we want this. The documentation says it performs validation, which can be easily seen if we add something like `input[type=url]:invalid { border-color: red; }` but the input still allows nonsensical things like `foo: bar`.

da2x:
Benefits of using the url input type:

* Browsers auto-complete knows to suggest URLs (not sure if any still suggest from browsing-history, but they at least suggest the right stuff from input-history). Specifically, this means you get the same partial auto-fill you get in the address field. Example: users can start typing "mailbo" and browsers will suggest https://webdav.mailbox.org/etc" if that’s in their history.
* Implies inputmode=url (touch-screen keyboards adds buttons for choosing TLDs and includes slashes, dots, and colons.)
* Implies spellcheck=false.

PR #1248, opened 2021-04-08
fix: update zip.js

gera2ld:
AMO requires third-party libraries to be at a published release instead of a copy of code ([ref](https://extensionworkshop.com/documentation/publish/third-party-library-usage/)).

So I'm going to drop the copy of `zip.js` in this repo and use an alternative from NPM. Finally I decide to use JSZip (which is more popular) instead of zip.js because of the friendly APIs and documentation.


tophf:
Isn't it too big? ~120kB minified vs the old ~51kB. I also don't like that it adds 10ms to page load time because it loads in the main chunk whereas previously it was loaded dynamically.

tophf:
>published release instead of a copy of code

How does this relate to VM?
Here's the published release: https://www.npmjs.com/package/@zip.js/zip.js

tophf:
Also, JSZip is several times slower...

tophf:
Ah, no, the speed is the same, the db was different.

gera2ld:
JSZip uses pako under the hood so it's bigger in size.
The reason I don't want to stay with zip.js is the lack of documentation and comparation of different implementations... Anyway I'll try since it's much smaller and may be better in performance.

src/common/zip.js
```diff
@@ -1,18 +1,8 @@
...
-    script.onload = () => resolve();
-    script.onerror = () => reject();
-    document.body.append(script);
-  });
-}
-
 const loadZip = memoize(async () => {
-  await loadScript('/public/lib/zip.js/zip.js');
-  return window.zip;
+  const zip = await import(/* webpackChunkName: 'zip' */ '@zip.js/zip.js/dist/zip');
```
tophf, line 4: Neat! But it creates `vendors~zip.js` in the root directory so maybe add `...splitVendor('@zip'),` in plaid.conf.js?

gera2ld: Done!

tophf:
The library is 34kB bigger now... but the real problem is that it's much slower: the old zip.js exported 6MB of scripts into 1.4MB zip in 2 seconds, while the new one takes 6 seconds. In devtools Sources` panel I see a lot of workers are created for a fraction of a second and then destroyed. In the old zip.js there was just one worker.

gera2ld:
I guess it's due to the consecutive `writer.add`s. Maybe I should replace them with concurrent ones.

src/options/views/tab-settings/vm-export.vue
```diff
@@ -137,60 +114,57 @@ function normalizeFilename(name) {
...
+  });
+  files.push({
+    name: 'violentmonkey',
+    content: JSON.stringify(vm),
+  });
+  const zip = await loadZip();
+  const blobWriter = new zip.BlobWriter('application/zip');
+  const writer = new zip.ZipWriter(blobWriter);
+  await Promise.all(files.map(file => writer.add(file.name, new zip.TextReader(file.content), {
+    lastModDate: file.lastModDate,
```
tophf, line 164: Yeah it's fast now but I wonder why [the documentation](https://gildas-lormeau.github.io/zip.js/core-api.html#zip-writing) suggests `bufferedWrite: true`? The output is identical in both cases.

tophf (approved):

src/options/views/tab-settings/vm-export.vue
gera2ld: The author has given a detailed clarification [here](https://github.com/gildas-lormeau/zip.js/issues/236).
I think it does not matter much in our case, but may make a difference when files are large.

PR #1256, opened 2021-04-13
fix: clear previous requests when GetInjected is fired

### LINKED ISSUE #1255: [Firefox] GM_xmlhttpRequest/GM.xmlHttpRequest regression in v2.12.12+
### What is the problem?

`GM_xmlhttpRequest` and `GM.xmlHttpRequest` don't work in some cases in Violentmonkey >= v2.12.12, but work in v2.12.11 and v2.12.10.

### How to reproduce it?

1. Install the script below
2. Run it on an IMDb movie page ([e.g.](https://www.imdb.com/title/tt0033467/)) with no other scripts running on the page
3. On the legacy/current site, an alert should appear showing details of the response
4. On the [new](https://i.imgur.com/iswQfqw.png) (React) site, the request is not executed and the alert is not shown

```javascript
alert("response: 200 OK (1256 bytes)")
```

No alert. The request is not initiated; no progress, state-change, error or load events are fired; no error is logged.

#### Works in

- Violentmonkey v2.12.11
- Tampermonkey 4.12.6132
- Greasemonkey 4.11 (just `GM.xmlHttpRequest`)

#### Doesn't work in

- Violentmonkey 2.12.12 - 2.12.14

### Script

```javascript
// ==UserScript==
// @name          XHR Test
// @description   Test XHR regression in Violentmonkey on the new IMDb site
// @version       0.0.1
// @namespace     test
// @include       https://www.imdb.tld/title/*
// @grant         GM.xmlHttpRequest
// @noframes
// ==/UserScript==

const request = {
    method: 'GET',
    url: 'https://www.example.com/',
    onerror: console.error,
    onload (res) {
        const { length } = res.responseText
        alert(`response: ${res.status} ${res.statusText} (${length} bytes)`)
    },
}

GM.xmlHttpRequest(request)
```
### Environment

- Browser: Firefox 87.0
- Violentmonkey: 2.12.14
- OS: Linux (Arch)

gera2ld:
fix #1255

### DIFF
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -13,10 +13,10 @@ import { getOption, hookOptions } from './utils/options';
 import { getInjectedScripts } from './utils/preinject';
 import { SCRIPT_TEMPLATE, resetScriptTemplate } from './utils/template-hook';
 import { resetValueOpener, addValueOpener } from './utils/values';
+import { clearRequestsByTabId } from './utils/requests';
 import './utils/clipboard';
 import './utils/hotkeys';
 import './utils/notifications';
-import './utils/requests';
 import './utils/script';
 import './utils/tabs';
 import './utils/tester';
@@ -62,7 +62,10 @@ Object.assign(commands, {
   /** @return {Promise<Object>} */
   async GetInjected(_, src) {
     const { frameId, tab, url } = src;
-    if (!frameId) resetValueOpener(tab.id);
+    if (!frameId) {
+      resetValueOpener(tab.id);
+      clearRequestsByTabId(tab.id);
+    }
     const res = {
       expose: !frameId && url.startsWith('https://') && expose[url.split('/', 3)[2]],
     };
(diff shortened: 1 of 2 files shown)


tophf:
To clear on navigation e.g. in `GetInjected` command because content scripts run only when a real navigation is performed to a new page.

gera2ld:
The script injection seems to start immediately if `window.vmData` exists while `GetInjected` is sent asynchronously. So I'm not sure if it's safe to clear requests in the callback of `GetInjected`.

tophf:
GetInjected should arrive first because it's sent first, before any script from vmData runs. Arguably.

tophf:
I mean GetInjected will arrive before GM_xmlhttpRequest sends its request to the background script.

gera2ld:
That's true. I'll make an update.

tophf:
Almost forgot: this should run only for frameId `0`.

tophf (approved):

Issue #1253, opened 2021-04-12
Import / Export does not work (v2.12.13)


wvxwxvw:
Firefox ESR, VM v2.12.13
Import / Export does not work. I also checked on a clean profile.

gera2ld:
zip.js web worker seems to be blocked by Firefox.

tophf:
The new zip.js creates workers from Blob, apparently. We should use a separate file like we did in the old version. I see such files in [dist](https://github.com/gildas-lormeau/zip.js/tree/master/dist) so we can either to get them from node_modules or build the same ourselves. Will you look into this, @gera2ld?

gera2ld:
I will look into this later.



Sent from my Mi phone
Apr 12, 2021 18:48 wrote on tophf ***@***.***>:

The new zip.js creates workers from Blob, apparently. We should use a separate file like we did in the old version. I see such files in dist<https://github.com/gildas-lormeau/zip.js/tree/master/dist> so we can either to get them from node_modules or build the same ourselves. Will you look into this, @gera2ld<https://github.com/gera2ld>?

PR #1257, opened 2021-04-18
fix: add missing support for "readScriptOptions"

22-2:
Probably related to #1253

### DIFF
--- a/src/options/views/tab-settings/vm-import.vue
+++ b/src/options/views/tab-settings/vm-import.vue
@@ -127,8 +127,12 @@ async function importBackup(file) {
         },
         position: more.position,
         props: {
-          lastModified: more.lastModified || +entry.lastModDate,
-          lastUpdated: more.lastUpdated || +entry.lastModDate,
+          lastModified: more.lastModified
+            || more.props.lastModified // Import data from Tampermonkey
+            || +entry.lastModDate,
+          lastUpdated: more.lastUpdated
+            || more.props.lastUpdated // Import data from Tampermonkey
+            || +entry.lastModDate,
         },
       },
     };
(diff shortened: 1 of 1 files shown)


src/options/views/tab-settings/vm-import.vue
```diff
@@ -127,8 +127,12 @@ async function importBackup(file) {
         },
         position: more.position,
         props: {
-          lastModified: more.lastModified || +entry.lastModDate,
-          lastUpdated: more.lastUpdated || +entry.lastModDate,
+          lastModified: more.lastModified
+            || more.props.lastModified // Import data from Tampermonkey
```
gera2ld, line 131: `more.props` would not exist when importing from a Violentmonkey export. You can use optional chaining (`?.`) here.

gera2ld, line 131: `more.props` would not exist when importing from a Violentmonkey export. You can use optional chaining (`?.`) here.

gera2ld (changes requested):

src/options/views/tab-settings/vm-import.vue
```diff
@@ -127,8 +127,12 @@ async function importBackup(file) {
         },
         position: more.position,
         props: {
-          lastModified: more.lastModified || +entry.lastModDate,
-          lastUpdated: more.lastUpdated || +entry.lastModDate,
+          lastModified: more.lastModified
+            || more?.props.lastModified // Import data from Tampermonkey
```
gera2ld, line 131: It should be `more.props?.lastModified` since `more` always exists but `more.props` may be undefined.

gera2ld (changes requested):

gera2ld (approved):

22-2:
Thank you for reviewing and approving this!

gera2ld:
Thanks for the PR 😃

PR #1263, opened 2021-04-26
fix: show tooltips in the bottom for popup buttons

gera2ld:
...to make them more consistent and friendly to keyboard navigation


tophf (approved):
This PR revealed an earlier problem: when hovering a button, its highlight and tooltip aren't cleared after moving the mouse outside the popup or over a static text:



I guess it should be cleared in `mouseout`.

gera2ld:
That was by design because I thought it might be good to continue from the last focused item with keyboard navigation.
Apparently this is not a common behavior. The focus is cleared now on mouse leave.

tophf:
Oops it actually breaks the click on these buttons and shows an error in devtools for the popup: The .native modifier for v-on is only valid on components but it was used on <span>.

PR #1285, opened 2021-05-23
I18n patch

Mattes0h:


tophf:
If you want to help with translation, please use [transifex](https://violentmonkey.github.io/localization/).

PR #1300, opened 2021-06-29
fix: preserve current unsaved value in storage editor

### LINKED ISSUE #1299: Changes for the currently opened key in storage editor are lost if GM_setValue is used for an unrelated key
### What is the problem?
Unable to append to JSON values through editor normal fashion. Also if you press save when you additions are present and invalid,  the value itself gets deleted completely. 

### How to reproduce it?
1. Create a JSON value like below
Snap45

2. try to insert the number 3 between 2 and 4 
3. Then while number is present press save. If you are too slow the edits gets removed, if you manage to press save while edits are present and there is a small error the stored value is deleted.

Ability to amend and append to the JSON value in a normal fashion without timeout.

It just resets value unless you press save quickly and then if there is an error it removes stored value and everything is lost.

### Environment
- Browser: Firefox
- Browser version: 89.0.2 64bit
- Violentmonkey version: 2.13.0
- OS: Windows 10

tophf:
fixes #1299.

### DIFF
--- a/src/options/views/edit/values.vue
+++ b/src/options/views/edit/values.vue
@@ -241,13 +241,16 @@ export default {
     onStorageChanged(changes) {
       const data = changes[scriptStorageKey]?.newValue;
       if (data) {
-        this.setData(data);
         const { current } = this;
+        const oldText = current && this.getValue(current.key);
+        this.setData(data);
         if (current) {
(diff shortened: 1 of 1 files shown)


gera2ld (approved):
LGTM

Issue #1086, opened 2020-10-20
"incognito": "split"


AlttiRi:
Currently the extension has no `"incognito": "split"` in `manifest.json` so all connections go through the common window even a userscript is running in the incognito tab.

It can be a problem if an Opera user that uses "Opera VPN" in an incognito tab with a userscript that loads some data from this site due to Violentmonkey that will load the data through the common window (instead the incognito one) where all connection does not go through that HTTPS proxy. It will disclose the site visiting to IPS, and IP to visited site.

Another case, not Opera related:
You open a site in the incognito tab that uses DDoS protection (Cloudflare's, or Google's one), you pass a captcha, but the userscript probably will not work due to you will not have the passed captcha in the common window, and the downloading of a content will be interrupted with 503 HTTP code.

Also it will allow to edit scripts in the incognito window.

It can be fixed with one line: add`"incognito": "split"` to `manifest.json` . 
Note: [It works only in Chrome.](https://bugzilla.mozilla.org/show_bug.cgi?id=1380812)

By the way, delete from the Chrome version manifest the Firefox related code: `browser_specific_settings: ...`. It's annoying to see all time in the extension's error log: `Unrecognized manifest key 'browser_specific_settings'.`.

---

It works OK, but there is one problem with the synchronisation of the GM storage (`GM_getValue`, `GM_setValue`).
The changes made in the other mode tab is visible only after the page reloading. The tabs in one mode have no such problem: after using of `GM_getValue` in one tab, `GM_setValue` returns the updated value in the other tab.
But this problem happens only if you interact with site in both private tab and common mode tab at the one moment.

Here is code to test it (do not forget to add `"incognito": "split"` in the manifest):
```js
// ==UserScript==
// @name        splite mode - example
// @namespace   Violentmonkey Scripts
// @match       *://example.com/
// @grant       GM_getValue
// @grant       GM_setValue
// @grant       GM_deleteValue
// @grant       GM_listValues
// ==/UserScript==


document.querySelector("div").remove();
const btn = document.querySelector("body").insertAdjacentHTML("afterBegin", `
<div>
  <button id="clear-button">Clear</button>
  <button id="increment-button">Increment</button>
  <span id="value"></span>
</div>
`);

const incrementButton = document.querySelector("#increment-button");
const clearButton = document.querySelector("#clear-button");
const valueElem = document.querySelector("#value");

value.textContent = GM_getValue("value", 0);

incrementButton.addEventListener("click", _ => {
[...]

PR #1307, opened 2021-07-06
feat: separate cookies in gmXHR per container

### LINKED ISSUE #1167: Can leak incognito data
Now this only occurs with Super Preloader enabled, but I think a UserScripts manager should block this either way

Bug Report here: https://github.com/machsix/Super-preloader/issues/485

tophf:
Fixes #1167.

`Cookie` and `Set-Cookie` HTTP headers for incognito tabs in Chrome or all tabs in FF will use their own cookie store. Chrome has two (default and incognito), FF may have many additional containers.

Could you help test it, please, @AlttiRi, @brian6932? Here's a dev version: [dev.zip](https://github.com/violentmonkey/violentmonkey/files/6769437/dev.zip)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@gera2ld/plaid-vue": "~1.5.5",
     "@gera2ld/plaid-webpack": "~1.5.5",
     "@types/chrome": "0.0.101",
+    "@types/firefox-webext-browser": "82.0.0",
     "cross-env": "^7.0.2",
     "cross-spawn": "^7.0.1",
     "del": "^5.1.0",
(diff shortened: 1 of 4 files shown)


AlttiRi:
What's about my issue: https://github.com/violentmonkey/violentmonkey/issues/1086

Okay, it fixes the second case with you have passed a captcha in the incognito window (and you have not passed in the common window).

But anyway it's not an alternative for `"incognito": "split"` mode, when you use a proxy extension in the incognito window (and it's disabled for the common window). (the first case of my issue.)

tophf:
> What's about my issue: #1086

This PR doesn't try to imitate `incognito:split` mode.

tophf:
@gera2ld, this code seems to work correctly with my scripts over the past few weeks so maybe you could take a look at the code?

gera2ld (approved):
LGTM

brian6932:
Yea this solved the issue 👍

PR #1309, opened 2021-07-09
feat: indicate in confirm UI when reinstalling

### LINKED ISSUE #1259: When reinstalling a script, the installation page is the same as in the initial installation
### What is the problem?

When you re-install a script version, the installation page is the same as in the initial installation:
the first sentence is `Installing script`, and the button is named `Confirm installation`, 
i.e. it doesn't inform the user that the script is already installed.

### How to reproduce it?
1. Open e.g. [this script page](https://greasyfork.org/en/scripts/394820-mouseover-popup-image-viewer), click "Install this script", and in the next page click "Confirm installation" ([screenshot](https://user-images.githubusercontent.com/723651/115593945-d16d3780-a2dd-11eb-8549-a27fcbc9073c.jpg))
2. Reload the page, notice that the button now says "Reinstall version 1.2.5", click it.
3. The installation page is the same as in the initial installation (i.e. as in the screenshot in step 1):

The page should inform the user that the script is already installed:
the first sentence should be `Re-Installing script`, the button should be named `Confirm re-installation`.

<details>
<summary>Here is a latest TM beta screenshot for comparison:</summary>

</details>

The page is the same as in the initial installation:
the first sentence is `Installing script`, the button is named `Confirm installation`

I'm using the latest version, 2.12.14

### Environment
- Browser: Chrome 90, Firefox 88
- Violentmonkey version: 2.12.14
- OS: win10 x64

tophf:
Fixes #1259.

* ### Reinstalling (same code)

  image

* ### Installing

  image

### DIFF
--- a/src/confirm/views/app.vue
+++ b/src/confirm/views/app.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="page-confirm frame flex flex-col h-100">
+  <div class="page-confirm frame flex flex-col h-100" :class="{ reinstall }">
     <div class="frame-block">
       <div class="flex">
         <div class="image">
@@ -7,7 +7,10 @@
         </div>
         <div class="info">
           <h1>
-            <div v-text="heading"/>
+            <div>
+              <span v-text="heading"/>
+              <span v-text="i18n('msgSameCode')" style="font-weight:normal" v-if="sameCode"/>
+            </div>
             <div class="ellipsis" v-text="name"/>
           </h1>
           <a class="url ellipsis" v-text="decodedUrl"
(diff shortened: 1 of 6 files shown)


src/confirm/views/app.vue
```diff
@@ -223,14 +243,13 @@ export default {
...
         this.error = null;
-        this.dependencyOK = true;
         this.installable = true;
         this.message = null;
         return true;
       }
     },
     close() {
-      sendCmd('TabClose');
+      sendCmdDirectly('TabClose');
```
gera2ld, line 252: `TabClose` seems to work only with `sendCmd`, because the handler depends on `tabId` that is only sent with `sendCmd`.

tophf: Indeed! I've fixed sendCmdDirectly instead so it automatically switches to sendCmd. I've added a reminder in the function's description so hopefully this will make it less fragile when we add commands in the future.

gera2ld (approved):
LGTM

tophf:
FWIW I've slightly desaturated the green/blue background.

PR #1313, opened 2021-07-18
fix: use K_SAVE hotkey + enable it in non-code tabs

tophf:
Depends on https://github.com/violentmonkey/vm-shortcut/pull/4


gera2ld:
Could you update `@violentmonkey/shortcut` to `1.2.4` by the way?

tophf:
Yes, I've updated it in the master.
Everything works now properly, including the case of pressing Ctrl-S on an unmodified script in Settings tab, which should do nothing, but previously it opened browser's save-as dialog.

gera2ld (approved):

PR #1327, opened 2021-08-01
feat: show dependencies right inside installer

tophf:
The simple lists are no longer shown for dependencies.
See the new screenshot below.


gera2ld (approved):

tophf:
I switched from `<select>` to a normal `<div>` so it allowed me to set a bigger height on the main script's entry, customize background color, add a file size column, a small `↗` link to open the dependency in a separate tab.

I had to use `this.$set` explicitly - is it okay?

image

tophf:
Okay, I've removed `this.$set` so I'm merging this.

PR #1330, opened 2021-08-07
fix: encode webRequest's responseHeaders to utf8 for GMxhr

### LINKED ISSUE #1329: `responseHeaders` are not `ByteString`
VM returns headers as `String` while it should be [`ByteString`](https://developer.mozilla.org/en-US/docs/Web/API/ByteString).
In most cases it works transparent, but not for some "Content-Disposition" header (when `filename` has non-ASCII character).

Native `XMLHttpRequest`, `fetch` and `GM.xmlhttpRequest` of TM return the related header string as `ByteString`.

That enchantment (?) may break a code that expect _only_ `ByteString`.

My current workaround (specially only for VM) is to check is it really a `ByteString` object:
```js
function byteStringToString(byteString) {
  const chars = [...byteString];
  const isBinaryString = chars.every(ch => ch.charCodeAt(0) < 256);  
  if (isBinaryString) {
    console.log("isBinaryString", isBinaryString);
    const isASCII = chars.every(ch => ch.charCodeAt(0) < 128);
    if (isASCII) { // not required
      console.log("isASCII", isBinaryString);
      return byteString;
    }
    const bytes = new Uint8Array(chars.map(ch => ch.charCodeAt(0)));
    return new TextDecoder().decode(bytes);
  }
  return byteString;
}
```

I tested it 1-2 months ago, but I don't think that something changed now.

tophf:
Fixes #1329.

@AlttiRi, please test it as I don't remember sites with such headers.

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -10,6 +10,7 @@ const VM_VERIFY = 'VM-Verify';
 const requests = {};
 const verify = {};
 const tabRequests = {};
+let encoder;
 
 Object.assign(commands, {
   ConfirmInstall: confirmInstall,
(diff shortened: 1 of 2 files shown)


AlttiRi:
For example, in some forums' attachments.
`(await fetch("https://example.moe/forum/index.php?attachments/xxx")).headers.get("content-disposition")`

---

For: `—123—✔✅—.zip`

I can't correctly post here the result:
`attachment; filename="â123ââââ.zip"; filename*=UTF-8''%E2%80%94123%E2%80%94%E2%9C%94%E2%9C%85%E2%80%94.zip`
In console log:
image
When I copy and paste to console's input:
image

I will test it later.

tophf:
Thanks for the link, it helped me re-implement this PR correctly, at last. I hope :-)

Issue #1154, opened 2020-12-27
GM_xmlhttpRequest lost its response when type is text


dodying:
### What is the problem?
sometimes, resp.response removed

### How to reproduce it?
1. new script
```js
// ==UserScript==
// @name        test
// @include     *
// @version     1
// @run-at      document-end
// @grant       GM_xmlhttpRequest
// ==/UserScript==
/* global  */
/* global $ */
/* eslint-disable no-debugger  */
(function () {
  GM_xmlhttpRequest({
    method: 'GET',
    url: 'http://www.google.com',
    onload (resp) {
      console.log(resp);
    },
    onprogress (res) { // or onreadystatechange // i don't test others

    }
  });
})();
```
2. in console, you can find `resp.response` is `undefined`
3. but remove `onprogress`, it run

### What is the expected result?
get `resp.response` when set `onprogress` or others

### What is the actual result?
get `resp.response` `undefined`

### What is the result in the [upcoming release](https://github.com/violentmonkey/violentmonkey/releases)?
same as stable version

### Environment
- Browser: Chrome
- Browser version: 86.0.4240.198
- Violentmonkey version: 2.12.9
- OS: win

tophf:
v2.12.9.3 will have the fix when it's [released](https://github.com/violentmonkey/violentmonkey/releases).

PR #1332, opened 2021-08-08
speedup: GMxhr in FF + use FileReader for internal encoding

tophf:
* Firefox can send Blob/File/ArrayBuffer directly via extension messaging, also in incognito tabs. Now, downloading huge files (100MB or more) takes just a couple of seconds. Previously, we used chunks because FF could block fetching of blobs on a page with a strict CSP.
* FileReader is much faster than buffer2string: 2x in Chrome, 10x in Firefox
* Using base64 makes the internal message size 1.5x smaller at least in Chrome because Chrome encodes our buffer2string's result to UTF-8, which increases the size and is slow. Now with base64 Chrome doesn't re-encode.

I've also renamed `details` to `opts` to indicate these are user options of GMxhr's call. I regularly got confused about `details` as it can conceptually refer to a response or a request or anything for that matter.

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -297,13 +297,16 @@ async function httpRequest(details, src, cb) {
   if (!req || req.cb) return;
   req.cb = cb;
   req.anonymous = anonymous;
-  // Firefox applies page CSP even to content script fetches of own blobs https://bugzil.la/1294996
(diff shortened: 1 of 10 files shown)


tophf:
* Added all supported body types like ArrayBuffer, typed arrays, URLSearchParams, ReadableStream
* And even fixed a bug: #1154 regressed again, apparently in 9903ca9e

PR #1334, opened 2021-08-09
fix: load icons via blob to avoid "tainted canvas"

tophf:
There is no icon in the confirm UI when installing [this script](https://github.com/tophf/mpiv/raw/master/script.user.js) in Chrome and there was an error in the background script about "tainted canvas". It could be a bug in Chrome, but it usually takes years before an extension-related bug is fixed. This PR implements a workaround.

P.S. Weirdly I can't reproduce the problem in a simplified extension that contains only loadImageData.


tophf:
Can't reproduce the bug anymore.

PR #1336, opened 2021-08-11
feat: implement GM_addElement + GM.addElement

### LINKED ISSUE #1335: `GM_addElement` missing
### What is the problem?

Scripts that use `GM_addElement` are completely broken :(

### How to reproduce it?
1. Install https://raw.githubusercontent.com/SNDST00M/material-dynmap/main/src/user.js
2. Open https://mc.westeroscraft.com
3. Watch the console get lit like a Christmas tree

https://www.tampermonkey.net/documentation.php#GM_addElement

Doesn't work and borks the whole script.

Identical result, there's been no work in this avenue.

### Environment
- Browser: Edge.
- Browser version: 92.0
- Violentmonkey version: 2.13.0
- OS: Windows 10 Home 20H2

tophf:
https://www.tampermonkey.net/documentation.php#GM_addElement

Fixes #1335.

### DIFF
--- a/src/injected/web/gm-api.js
+++ b/src/injected/web/gm-api.js
@@ -28,7 +28,7 @@ const {
   TextDecoder: { prototype: { decode: tdDecode } },
   URL: { createObjectURL, revokeObjectURL },
 } = global;
-const { removeAttribute } = Element.prototype;
+const { remove, removeAttribute } = Element.prototype;
 
 const vmOwnFuncToString = () => '[Violentmonkey property]';
 export const vmOwnFunc = (func, toString) => {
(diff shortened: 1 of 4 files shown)


gera2ld (approved):
LGTM

src/injected/web/gm-api.js
```diff
@@ -152,20 +154,21 @@ export function makeGmApi() {
...
+     * @param {Object} [attributes]
+     * @returns {HTMLElement}
+     */
+    GM_addElement: (parent, tag, attributes) => (
+      typeof parent === 'string'
+        ? addElement(undefined, parent, tag)
+        : addElement(parent, tag, attributes)
+    ),
+    GM_addStyle: css => (
+      addElement(undefined, 'style', { textContent: css, id: getUniqId('VMst') })
```
tophf, line 170: @gera2ld, now that AddElement is using parentIndex, I wonder if we can remove `id: getUniqId('VMst')` ?

gera2ld: Yeah, I think it's useless now.

tophf: After a lot of attempts to handle ShadowDOM (all were too complicated and wouldn't work if `parent` is a detached node not in DOM), I reverted to VM's old approach with a temporary `id` so the style's id will stay to avoid breaking something.

PR #1339, opened 2021-08-14
fix: accept only JS to disable prefetch and CSP errors

tophf:
How to test: click the install button in https://greasyfork.org/scripts/430515 and when VM's installer shows up, open devtools console.

@JasonBarnabe, do you think this solution (sending an `Accept` header only with JS types) is universal? Otherwise maybe you could return empty content with 404 status if we add some special header to the request?

@gera2ld, I'm not sure this PR is even necessary because the only bad thing is the spam in console, and it happens only when a dependency on greasyfork was deleted so it responds with a `404` html page.

P.S. I have another solution: 50 dense lines of stripping the header via webRequest API, but it's probably an overkill.

### DIFF
--- a/src/common/util.js
+++ b/src/common/util.js
@@ -225,6 +225,24 @@ export async function requestLocalFile(url, options = {}) {
   });
 }
 
+/**
+ * Excludes `text/html` to avoid LINK header that Chrome uses to prefetch js and css,
+ * because GreasyFork's 404 error response causes CSP violations in console of our page.
+ */
+const FORCED_ACCEPT = [
+  'application/*',
+  'text/javascript',
+  'text/ecmascript',
+  'text/plain',
+  'text/css',
+].join(',');
+/**
+ * WARNING! Rework this if a newly added host offers content not covered by FORCED_ACCEPT,
+ * e.g. if github starts to add LINK, userscripts may use it to load `image/*` and so on.
+ */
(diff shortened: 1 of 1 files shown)


JasonBarnabe:
I see a few errors in devtools, I'm not sure which this is supposed to solve.

Greasy Fork's behaviour when a deleted script is requested is defined [here](https://github.com/JasonBarnabe/greasyfork/blob/c038a15123a189c7a9e1fd4d437f8fe01363747e/app/controllers/application_controller.rb#L28-L37); if the client requested text/html, then a full HTML 404 page, otherwise a 404 with no body.

tophf:
@JasonBarnabe, thanks! I tested in Chrome Canary, which supports the LINK header. There's a bunch of CSP-related errors about https://greasyfork.org/packs/js/application-12ee71cff1375d2ade9a.js which is loaded via this header. Our requests previously used the default `accept: */*` which is probably `format.all` in GF's `render_error` so maybe ruby thinks `format.html` is included in `format.all` and runs the first branch?

JasonBarnabe:
I'd expect that when you say `Accept: */*` then both `format.html` and `format.all` would match. The behaviour is that it'll use the first match, so you're getting the HTML 404 page.

From my perspective, sending `Accept: text/javascript` would be the proper thing to do here.

tophf:
Okay, so now I've made it simple and reliable enough to be merged.

PR #1343, opened 2021-08-15
fix: forbid {...objSpread} in injected/* via eslint

tophf:
Our .browserslistrc targets old browsers so the compiled code for {...objSpread} uses babel's polyfill that calls methods like `Object.assign` instead of our safe `assign`

Ideally, `eslint-plugin-compat` should be used but I couldn't make it work.


tophf:
The change is quite trivial so I'm merging it myself. If a better solution will be found it can be implemented afterwards.

PR #1345, opened 2021-08-18
refactor: use Proxy for `browser`, embed it as module

tophf:
1. The startup time is faster for ~5ms because Proxy works on demand.

   The overall time spent inside the module has marginally decreased (from 7ms to 5ms) judging by performance profiler while opening VM's dashboard page.

   Now there's no need to manually add new methods because the meta object lists just the special cases - non-asynchronous and overridden methods.

2. browser.js is embedded as a module because executing each separate content script has internal overhead. The overall file size decreased a little because the actual contents of browser.js is just 0.9k, the rest 1.2k was webpack's bootstrap. Now browser.js is embedded into two files: injected.js and common.js.

   Embedding it allowed some deduplicating of webpack/plaid configs.


tophf:
I've tested it for a few days and made some changes, everything seems to work so I'm merging it to release a new beta.

PR #1351, opened 2021-08-24
feat: option to group disabled scripts in popup

tophf:
image

image

* The iframe scripts aren't affected, they're still listed in one combined menu.
* The `>` icon is moved to the left since all persistent icons are shown there
* The background color of the active menu is one notch stronger
* The last menu has a dashed border in collapsed state to separate it from the footer


tophf:
Seems fine. The only thing I decided to change is the order of items in the option: show > group > hide.

PR #1352, opened 2021-08-25
feat: customizable @noframes in editor settings

tophf:
The label is `Run in frames` which is the opposite of `noframes` because it seems easier to understand than Tampermonkey's "Run only in top frame". People normally say "frame" to refer to an `<iframe>` or `<frame>` element, we don't normally call the main page "top frame", it's an internal term from the extension API or from other low-level internals/specifications.

image


tophf:
Seems fine.

PR #1354, opened 2021-08-25
feat: separately inject start/body scripts, cache for 1hr

### LINKED ISSUE #1353: chrome.storage is super slow in Chrome
`chrome.storage.local` has always been insanely slow in Chrome, it can easily take more than a second to read a multimegabyte script with `@require` like [this one](https://greasyfork.org/scripts/24508). For 10 years it was this slow and apparently it will take at least another 10 years to fix it.

Such a huge delay means all scripts for this tab, even those with `document-start`, will run after the site has already loaded.

The simplest solution is to keep the code in the same `cache` we use for matching. Memory usage won't increase by much as the most of it is spent on JS engine itself and our other variables.

A more convoluted solution is to mirror the data in IndexedDB, which is 10 times faster, but still 1000 times slower than memory, and it will waste the space on the disk for mirroring.

tophf:
Arguably this should fix #1353 in 99% of cases.  Memory usage won't increase by much as the most of it is spent on JS engine itself and our other variables.

Edit:

Turns out the longer caching doesn't help in case of huge `cache` or `values` because just sending them via a message takes an insane amount of time (at least in Chrome) so I've implemented separate loading of `document-start` and -body scripts/data from storage and separate injection. The rest (document-end and -idle) will be sent in the next message.

### DIFF
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -48,14 +48,14 @@ Object.assign(commands, {
       clearRequestsByTabId(tabId);
     }
     const res = await getInjectedScripts(url, tabId, frameId);
-    const { feedback, gmVal } = res._tmp;
+    const { feedback, valOpIds } = res._tmp;
     res.isPopupShown = popupTabs[tabId];
     // Injecting known content scripts without waiting for InjectionFeedback message.
     // Running in a separate task because it may take a long time to serialize data.
     if (feedback.length) {
       setTimeout(commands.InjectionFeedback, 0, { feedback }, src);
     }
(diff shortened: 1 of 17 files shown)


tophf:
Another speedup, file size optimization, and an improvement for step-by-step debugging was getting rid of webpack's internal redirections used by `::safe()` method imported from `helpers.js`. Now these variables and methods are auto-created by `webpack.conf.js` and checked in a custom eslint `no-restricted-syntax` rule.

tophf:
No problems so far for the past week that I've been using it so I guess I'll merge it and release a beta.

PR #1358, opened 2021-09-02
ui: show "Not configured yet" instead of sync error

tophf:
* show "Not configured yet" instead of sync error
* show the status in the same row to make it more noticeable

   image

* hide "[x] Sync script status" when `None` is selected 

   image


gera2ld (approved):

tophf:
Maybe `Not authorized yet` would be better?

tophf:
Yeah, I guess.

PR #1359, opened 2021-09-03
speedup: define safe globals via header injection

### LINKED ISSUE #1373: replacing Function.prototype.call can break violentmonkey
### What is the problem?
replacing `Function.prototype.call` can break violentmonkey if there is `// @run-at document-end` or `// @run-at document-idle` userscript that runs on the webpage

### How to reproduce it?
1. install userscript like:
```javascript
// ==UserScript==
// @name        useless script
// @namespace   http://bzzzzdzzzz.blogspot.com/
// @description does nothing
// @author      BZZZZ
// @include     *
// @version     0.1
// @grant       none
// @run-at      document-end
// ==/UserScript==

undefined;
```
2. save this as html file:
```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="UTF-8"/>
  <title>violentmonkey bug demo</title>
  <script type="text/javascript">
    Function.prototype.call=function(cthis,...cargs){
      console.log('Function.prototype.call:\n\tfunction:%O\n\tthis:%O\n\targuments:%O\n\t%O',this,cthis,cargs,Error('dummy error to show call stack'));
      document.write('<h1>open javascript console</h1>');
      return Reflect.apply(this,cthis,cargs);
    };
  </script>
</head>
<body>
</body>
</html>
```
and open with browser with violentmonkey (works on firefox and brave)
3. if webpage shows "**open javascript console**" you reproduced the bug
4. open javascript console
you will see `Function.prototype.call` logs but nothing in webpage and userscript calls ` Function.prototype.call`

for developer: to fix this do
```javascript
const fcall=Function.prototype.call.bind(Function.prototype.call);
```
before webpage script can replace any native functions and use
```javascript
fcall(func,thisArg,...args);
```
instead of
```javascript
func.call(thisArg,...args);
```
in `injected-web.js`

### Environment
- Violentmonkey version: Violentmonkey v2.13.0

tophf:
Profiler shows injected.js loads faster: 13ms previously, 10ms now. This is because webpack was calling its internal machinery each time the safe method was defined and used.

Production files also became ~1kB smaller.

It's hackish, but hopefully still acceptable. The only small problem I don't know how to solve is making VSCode jump to definition of such globals as `forEach`.

Edit: also fixes #1373.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -10,6 +10,6 @@ module.exports = {
     }],
   ],
   plugins: [
-    '@babel/plugin-proposal-function-bind',
+    './scripts/babel-plugin-safe-bind.js',
   ],
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,12 +1,21 @@
+const acorn = require('acorn');
 const unsafeEnvironment = [
   'src/injected/**/*.js',
-  // these are used by `injected`
+];
+// some functions are used by `injected`
+const unsafeSharedEnvironment = [
   'src/common/browser.js',
   'src/common/consts.js',
   'src/common/index.js',
   'src/common/object.js',
   'src/common/util.js',
 ];
+const commonGlobals = getGlobals('src/common/safe-globals.js');
+const injectedGlobals = {
+  ...commonGlobals,
+  ...getGlobals('src/injected/safe-injected-globals.js'),
+};
+
 module.exports = {
   root: true,
   extends: [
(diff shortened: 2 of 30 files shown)


tophf:
I guess a better solution might be to write a plugin for webpack that intercepts the imports from `safe-globals.js` and replaces the calls with a dummy constant. It will cause the import to be dropped as unused. The constant will be added to `header` string, which in the end will be written to the output file. Not sure I can implement it though :-)

tophf:
@gera2ld, I'm about to merge it because it also fixes a fundamental vulnerability :-)

PR #1381, opened 2021-09-30
feat: customize export file name

### LINKED ISSUE #1378: Exported ZIP file name should contain `violentmonkey` word
VM exports the settings to a ZIP file with name like this: `scripts_2021-09-29_12.40.33.zip`.

The `scripts` word is too common. It's makes the zip difficult to find with a search. 
Also when you found in your folder such file by accident you will not understand that it is, what the scripts are in the zip.

It should be prefixed with `violentmonkey` word.

---
_BTW, (off topic) may be interested: https://github.com/AlttiRi/twitter-click-and-save#filename-format_

_The most of people do not sort files by directories, they just put them in a bulk folder, so the files should have such name to make the search of them simple._

tophf:
Fixes #1378 using a subset of [moment.js tokens](https://momentjs.com/docs/#/displaying/format/).

image

The tooltip:

image

@gera2ld, @AlttiRi, PTAL.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -562,6 +562,12 @@ menuScriptEnabled:
 msgCheckingForUpdate:
   description: Message shown when a script is being checked for updates by version numbers.
   message: Checking for updates...
+msgDateFormatInfo:
+  description: Help text of the info icon in VM settings for the export file name.
+  message: >-
+    Click to open MomentJS documentation.
+    Allowed tokens: $1.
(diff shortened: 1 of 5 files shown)


AlttiRi:
Well, for me it would be enough just `violentmonkey-`, or `violentmonkey-backup-chrome-` (as it does TM) prefix.
_BTW, it it support of using `[`, `]` characters?_

tophf:
Of course, see the screenshots.

PR #1398, opened 2021-10-15
Set CI to upload artifacts

sn-o-w:
It would be nice to have automated builds every time there is a new commit or PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,16 +1,24 @@
 name: CI
 
-on:
-  - pull_request
+on: [push, pull_request]
 
 jobs:
   ci:
     runs-on: ubuntu-latest
-    steps:
+
+    steps: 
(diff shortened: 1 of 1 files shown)


tophf:
1. The PR should probably disable AMO signing in our `.github/workflows/release.yml` because currently it's completely broken so it would just wait for the timeout and waste time (currently 30 minutes, by default 15).
2. By default it uses `retention-days: 90` which seems wasteful so I'd use `30` or less or maybe automatically delete all artifacts when a new beta release is made?
3. The current location of artifacts seems hard to discover, AFAICT, so you should probably add a link in `README.md`

@gera2ld, WDYT? The idea here is apparently to attract more people to use our Nightlies and report bugs earlier.

tophf:
Duh, github intentionally doesn't allow non-logged-in users to download the artifacts.

PR #1431, opened 2021-11-27
better regexp fix

xBZZZZ:
comments in `src/injected/web/util-web.js` explain it all

also is there better way to suggest change than copying everything?

edit: I forgot to tell that if `global` is accessed on `null` prototype regexp and it doesn't exist it will behave like `false` (replacing will behave weird)

### DIFF
--- a/src/injected/web/requests.js
+++ b/src/injected/web/requests.js
@@ -1,7 +1,8 @@
 import bridge from './bridge';
 
 const idMap = createNullObj();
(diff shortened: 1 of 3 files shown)


tophf:
Calling setPrototypeOf even once one a page will put it into a slow or otherwise bad mode so this is not acceptable.

tophf:
I've had enough with this moronic language feature so we'll just remove all use of regexp replace in our web part.

PR #1438, opened 2021-12-10
feat: add `isBlobUrl` parameter to GM_getResourceURL

tophf:
```js
GM_getResourceURL(name) // blob: URL
GM_getResourceURL(name, true) // blob: URL
GM_getResourceURL(name, false) // data: URL
```


tophf:
I've simplified the PR: the second parameter is a boolean now, `true` (default) is a blob URL, otherwise it's a data URL.

gera2ld (approved):
LGTM

lisonge:
isBlobUrl not work
image

tophf:
@lisonge, it works since Violentmonkey v2.13.0.15, so you're probably using an old version.

PR #1440, opened 2021-12-22
more null objects to not get messed up by setters and getters in `Object.prototype`, `JSON` functions are stored separately in vault

xBZZZZ:
post title explains it all


tophf:
The GM_info object is exposed to userscripts so it should have a standard object shape because userscripts may use standard object methods like hasOwnProperty directly.

JSON.parse and JSON.stringify are own keys from a different JS realm so it cannot be spoofed/broken by the current realm.

I don't see the need for this PR.

xBZZZZ:
> The GM_info object is exposed to userscripts so it should have a standard object shape because userscripts may use standard object methods like hasOwnProperty directly.

but `GM_info.script` is null prototype object and `GM_info.script.hasOwnProperty` will not work (inside userscript)
is that intended?

tophf:
It's an unintentional consequence of using a null prototype in metaCopy for convenience, fixed in 5755efd0, thanks.

tophf:
Hmm, if we use standard object prototype our code can be broken by a hostile page that throws inside the setter so I'll probably rework it...

xBZZZZ:
> JSON.parse and JSON.stringify are own keys from a different JS realm so it cannot be spoofed/broken by the current realm.

note: if web page can get object (array, function, error, object, regexp) from vault's realm it can mess up `JSON`: `object_from_vaults_realm.hasOwnProperty.constructor("JSON.parse='asd';")();`

tophf:
> if web page can get object (array, function, error, object, regexp) from vault's realm

AFAIK it can't. If there is such a way it would make the vault completely meaningless and useless.

tophf:
On the other hand, theoretically the page can do it by using a vulnerability in some popular userscript.
I guess I'll make SafeJSON a prototypeless copy just in case...

tophf:
Or maybe you can reduce this PR to JSON-related changes and I'll merge it.

xBZZZZ:
> Or maybe you can reduce this PR to JSON-related changes and I'll merge it.

I don't think I can do that (I deleted the fork).

xBZZZZ:
> AFAIK it can't. If there is such a way it would make the vault completely meaningless and useless.

only if vault iframe `contentWindow` gets modified before functions are stored in array

PR #1444, opened 2021-12-31
chore: publish release on tag push

gera2ld:
Publish to CWS if a release tag is created.


happymx (approved):

tophf:
@gera2ld, are you merging this PR? I want to publish a new release now.

tophf:
I meant a beta release, so maybe this PR is not necessary?

gera2ld:
it's not necessary

gera2ld:
It seems that self-distributed version on AMO also [requires source code review](https://extensionworkshop.com/documentation/publish/signing-and-distribution-overview/#distributing-your-addon) now, so the release process on AMO cannot be automatic any more. I think we can drop support for beta version for Firefox.

tophf:
It doesn't say the submission is blocked until the check is performed.

gera2ld:
But it's actually blocked.

image

tophf:
Do you know if it's possible to include the source code when submitting automatically?

gera2ld:
There is [a new API](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-sources) for this but no support in sign-addon yet. I'm trying to achieve this by using cURL directly.

tophf:
Our version numbers are too long e.g. `2.13.10.1`. What about dropping one component? Next release: 2.14. Beta: 2.14.1.

gera2ld:
> What about dropping one component?

Yep I agree.

Issue #1432, opened 2021-11-29
/src/injected/web/requests.js `async function encodeBody(body)` problems


xBZZZZ:
https://github.com/violentmonkey/violentmonkey/blob/7c565c03a252594b00880add2740b1c2431568ca/src/injected/web/requests.js#L160<ul>
<li>web page can<pre lang="javascript">Blob.prototype.__defineGetter__(Symbol.toStringTag,console.trace);</pre></li>
<li>web page can<pre lang="javascript">Promise.prototype.then=console.trace;</pre>to break <code>async function</code>s that return promise (or any object that has <code>then</code> in prototype chain)<hr>try running<pre lang="javascript">Promise.prototype.then=console.trace;
(async function(){
  return new Promise(()=>{});
})();</pre>(firefox recommended because <a href="https://www.reddit.com/r/chrome/comments/qo3435/devtools_console_bug/">it breaks chromium devtools</a>)</li>
</ul>

tophf:
I'm aware of this but I don't see what we can do about it because breaking a Promise is completely nonsensical. Only a specifically crafted hostile web page made by "m@d skillz haxx0rz" would do it so I don't see the point of fighting it.

xBZZZZ:
note: this is a problem only if web page can access vault iframe's `Promise.prototype`

PR #1453, opened 2022-01-21
make SafePromise.prototype.then unreplaceable

xBZZZZ:
https://github.com/violentmonkey/violentmonkey/issues/1432


tophf:
I don't understand why changing a function reference affects the prototype, which we don't use. Can you demonstrate the bug?

tophf:
Ah, I guess it's the same eval thing.

xBZZZZ:
> I don't understand why changing a function reference affects the prototype, which we don't use. Can you demonstrate the bug?

`Object.freeze(obj)` freezes `obj` and returns `obj`
---
```javascript
Object.freeze(Promise.prototype).then;
```
does same thing as
```javascript
(function(){
  var pp=Promise.prototype;
  Object.freeze(pp);
  return pp.then;
})();
```

tophf:
Yeah, I know. I just misread it the first time and thought you were freezing the function.

PR #1491, opened 2022-04-13
feat: update isRemote

lelinhtinh:
Add a simple test for private address ([RFC 1918](https://datatracker.ietf.org/doc/html/rfc1918), [RFC 4193](https://datatracker.ietf.org/doc/html/rfc4193)) and reserved domains ([RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606)).

### DIFF
--- a/src/common/util.js
+++ b/src/common/util.js
@@ -252,8 +252,19 @@ const FORCED_ACCEPT = {
   'greasyfork.org': 'application/javascript, text/plain, text/css',
 };
 
-export const isRemote = url => url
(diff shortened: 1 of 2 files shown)


tophf:
Could you convert the entire thing to regexp? In modern browsers it works faster than JS checks.

lelinhtinh:
You right 👍 
https://jsben.ch/f1uzE

[Built-in Method vs RegExp](https://jsben.ch/f1uzE)

PR #1521, opened 2022-07-03
fix: always show buttons in popup on touchscreen

tophf:


gera2ld (approved):
LGTM. BTW it's quick! 🚀

PR #1524, opened 2022-07-12
fix: show buttons if popup was open as unfocused

### LINKED ISSUE #1520: [BUG] cant see edit button when hovering over scripts on expanded toolbar icon
- [ x] I have searched for existing issues that already reported this problem and found none

**Describe the bug**
click on violentmonkey
see the list of installed scripts 
cant see edit script or hamburger menu of script unless user knows where it is and click on it blindly. then they appear.

**Expected behavior**
they should be seen without user click on them first.

**Environment:**
- OS: linux
- Browser: firefox 102
- Violentmonkey Version: 2.13.0

tophf:
Fixes #1520 hopefully.

### DIFF
--- a/src/popup/views/app.vue
+++ b/src/popup/views/app.vue
@@ -103,7 +103,8 @@
                  @contextmenu.exact.stop="onEditScript(item)"
                  @mousedown.middle.exact.stop="onEditScript(item)" />
           </div>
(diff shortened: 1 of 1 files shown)


tophf:
I've pushed bd68576e separately.

PR #1532, opened 2022-07-20
feat: faster wrapper prop resolution

### LINKED ISSUE #1531: Math.xxx >50x slower in userscripts
# Issue
calling `Math.abs`, `Math.random`, etc. is >50x slower in userscripts

## test
to test just copy the following into a userscript open the site it is injected into, and paste the same into the console:
```js
let acc = 0;
console.time("Math.randomx1000000");
for(let i=0; i<1000000; i++) acc += Math.random();
console.timeEnd("Math.randomx1000000");
console.time("Math.absx1000000");
for(let i=0; i<1000000; i++) acc += Math.abs(acc);
console.timeEnd("Math.absx1000000");

// result in userscript:
//   Math.randomx1000000: 4118.552978515625 ms
//   Math.absx1000000: 4082.463134765625 ms
// result in console:
//   Math.randomx1000000: 135.682861328125 ms
//   Math.absx1000000: 55.900634765625 ms
```
## temporary fix
rebinding the object like this fixes the problem:
```js
const Math = window.Math;
```

despite the simple fix, this caused me major issues (freezes when doing image manipulation) and headaches (took me way to long to debug, because the profiler just shows 40s spent with `Math.abs` being the only thing on that line)
## speculation on the origin
I assume the property access somehow has overhead because of sand-boxing or something, ~~but the Math object is indistinguishable from the normal one (no proxy, no funky getters / setters), so I have no clue what is going on~~.

tophf:
Fixes #1531.

* Uses a second `with` to let JS engine resolve the globals we already resolved instead of calling our Proxy's `get` trap that replicated the same behavior but much slower.
* ~Uses local variables for 16 frequently used globals like `Array` or `Math`. The only possible problem is that a userscript cannot alter the value of these globals via the standard approach like `window.Math = polyfill` as this will not change the variable.~
* Removes `with` from grant-none scripts to avoid the drop in performance

### DIFF
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -38,23 +38,9 @@ const KEY_XHR_INJECT = 'xhrInject';
 const BAD_URL_CHAR = IS_FIREFOX
   ? /[#&',/:;?=+]/g // FF shows `@` fine as ASCII but mangles it as full-width
   : /[#&',/:;?=+@]/g;
-const BIGINT = 'BigInt';
 const GRANT_NONE_VARS = `{GM,GM_info,unsafeWindow${
   IS_FIREFOX ? '' : ',cloneInto,createObjectIn,exportFunction'
 }}`;
-/** Userscript globals that are likely to be used hundreds of times per second */
-const INLINED_GLOBALS = [
-  BIGINT in global && BIGINT,
-  'Boolean',
-  'Date',
-  'Math',
-  'Node',
-  'Number',
-  'Object',
-  'parseFloat',
-  'parseInt',
-  'performance',
-]::trueJoin(',');
 const expose = {};
 let isApplied;
 let injectInto;
@@ -306,8 +292,7 @@ function prepareScript(script) {
       // using a shadowed name to avoid scope pollution
       grantNone ? GRANT_NONE_VARS : 'GM'}${
       IS_FIREFOX ? `,${dataKey}){try{` : '){'}${
-      // inlining is inside a separate block scope to be fast
-      grantNone ? '' : `with(this)with(c){delete c;{let{${INLINED_GLOBALS}}=this;`
+      grantNone ? '' : 'with(this)with(c)delete c,'
     // hiding module interface from @require'd scripts so they don't mistakenly use it
     }((${grantNone ? 'define,module,exports' : ''})=>{`,
     ...reqsSlices,
@@ -317,7 +302,7 @@ function prepareScript(script) {
     // adding a new line in case the code ends with a line comment
     !code.endsWith('\n') && '\n',
     hasReqs && wrap && '})()',
-    wrap && `})()${grantNone ? '' : '}}'}${IS_FIREFOX ? `}catch(e){${dataKey}(e)}` : ''}}`,
+    wrap && `})()${IS_FIREFOX ? `}catch(e){${dataKey}(e)}` : ''}}`,
     // 0 at the end to suppress errors about non-cloneable result of executeScript in FF
     IS_FIREFOX && ';0',
     // Firefox lists .user.js among our own content scripts so a space at start will group them
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -38,6 +38,25 @@ const KEY_XHR_INJECT = 'xhrInject';
 const BAD_URL_CHAR = IS_FIREFOX
   ? /[#&',/:;?=+]/g // FF shows `@` fine as ASCII but mangles it as full-width
   : /[#&',/:;?=+@]/g;
+/** Userscript globals that are likely to be used hundreds of times per second */
+const INLINED_GLOBALS = [
+  'Array',
+  'Date',
+  'Boolean',
+  'Math',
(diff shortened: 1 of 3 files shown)


tophf:
I should probably exclude Promise, setTimeout and clearTimeout from the list because some libraries used by userscripts may want to polyfill these things...

RubenArtmann:
> The only possible problem is that a userscript cannot alter the value of these globals via the standard approach like `window.Math = polyfill` as this will not change the variable.

You can make this work, because `window` is a `Proxy`, so you can just propagate the change it the setter, but `Math = polyfill` wont work, because that is `with(userscript_scope)`. `unsafeWindow.Math = polyfill` will work only for the global one and not change for the userscript itself (unless you make that one a proxy too).

src/background/utils/preinject.js
```diff
@@ -282,9 +301,17 @@ function prepareScript(script) {
...
   const wrap = !meta.unwrap;
+  const { grant } = meta;
+  const grantNone = grant.length === 1 && grant[0] === 'none';
   const injectedCode = [
+    wrap && `window.${dataKey}=function(scope,${dataKey}){
+      try{
+        with(scope)
+        ${grantNone ? '' : `with(${dataKey})`}
+        ((define,module,exports)=>{
+          ${grantNone ? '' : `delete(${dataKey}); var{${INLINED_GLOBALS}}=this;`}
```
RubenArtmann, line 312: I dont understand this. Wouldnt this set `var Math = this`???
how do you make `window.Math` reflect this var?
also how do you set this var if `window.Math = polyfill` is executed?

RubenArtmann:
I propose a test like this:
```js
// userscript
let start = performance.now();
for(let i=0; i<1000000; i++) Math.abs(i);
// with fastprop: ~10ms
// without: ~5000ms
let time = performance.now() - start;
assert(time < 100);

// you can polyfill or whatever
Math = "a";
assert(Math===window.Math);
window.Math = "b";
assert(Math==window.Math);
```
and
```js
let Math = "shadowing works";
```

tophf:
> You can make window.Math = polyfill work

No, the change won't be reflected in the variable. Anyway, it's a moot point as I don't envision anyone polyfilling Math.

> BigInt, JSON

I didn't see it being used in such loops.

RubenArtmann:
> > BigInt, JSON
> 
> I didn't see it being used in such loops.

BigInt is needed for 64bit integers, and as far as i know `JSON.parse(JSON.stringify(value))` is used as a deep clone as its optimized by the engines. Your decision though.

tophf:
The point is I didn't see it being used in **these** loops i.e. inside `for(let i=0; i<1000000; i++)`

RubenArtmann:
> The point is I didn't see it being used in **these** loops i.e. inside `for(let i=0; i<1000000; i++)`

Did you expect me to run *any* loops with >10mil iterations in a userscript?

tophf:
Hundreds/thousands aren't uncommon, I think.

RubenArtmann:
> Hundreds/thousands aren't uncommon, I think.

what you expected: 100/1000 iterations => 5ms overhead (nobody cares)
what eventually happened: 10 000 000 iterations => 50s overhead (freezes)

My point being: 
At least BigInt should be included, because just like you didnt expect me to use `10mil x Math.abs`, even if you dont expect someone to use `10mil x BigInt`, someone eventually will. So if you just add it now, you can spare them the surprise of finding out that BigInt costs an Proxy lookup and thus causes freezes. 
Or don't. I just wanted to make sure you have at least the information i have, so you can make an good decision.

tophf:
I think, BigInt is a recent addition to the web platform so it's likely that userscripts may use a polyfill, which is a strong point against. BigInt calculus is also slow anyway so the lookup won't be so pronounced.

tophf:
OTOH, if they use a polyfill it should skip if the type is already present, so I can add the inlined var conditionally, which is doable.

tophf:
I've removed inlining because on 1k reps there's no speedup at all, 10k reps gets 3x. I'm not convinced we should micro-optimize for the super rare case of 1kk reps that uses a global like Math.

src/background/utils/preinject.js
```diff
@@ -282,17 +285,25 @@ function prepareScript(script) {
...
+  const grantNone = !numGrants || numGrants === 1 && grant[0] === 'none';
   const injectedCode = [
+    wrap && `window.${dataKey}=function(${
+      // using a shadowed name to avoid scope pollution
+      grantNone ? GRANT_NONE_VARS : 'GM'}${
+      IS_FIREFOX ? `,${dataKey}){try{` : '){'}${
+      grantNone ? '' : 'with(this)with(c)delete c,'
     // hiding module interface from @require'd scripts so they don't mistakenly use it
-    wrap && `window.${dataKey}=function(module,${dataKey}){try{with(module)((define,module,exports)=>{`,
+    }((${grantNone ? 'define,module,exports' : ''})=>{`,
```
gera2ld, line 298: This line is breaking UMD requirements in pages using RequireJS when `@grant` is not none.

PR #1539, opened 2022-07-30
refactor: split bg requests.js

tophf:
@gera2ld, requests.js seems overly "packed" to me so I've split it into 3 files. WDYT?


gera2ld:
LGTM

PR #1540, opened 2022-07-31
feat: enable modern JS syntax in vue templates

tophf:


tophf:
I also wonder if we could start migrating to a more modern framework like Vue3 or svelte?

gera2ld:
> I also wonder if we could start migrating to a more modern framework like Vue3 or svelte?

Good idea. I guess the effort for migration to Vue3 is acceptable.

gera2ld (approved):

PR #1544, opened 2022-08-07
Migrate to Vue 3

gera2ld:
## Notable changes

- Vue 2 -> Vue 3
- Webpack 4 -> Webpack 5
- PostCSS 7 -> PostCSS 8
  - Dropped preCSS because it has no PostCSS 8 support and seems to be deprecated
  - TailwindCSS can be enabled at any time
- TypeScript supported but not required
- Tape -> Jest

## Remaining Issues

- ~~`wrapper-webpack-plugin` [does not work well with Webpack 5](https://github.com/webpack/webpack-sources/issues/92#issuecomment-709601572), maybe should be dropped~~
- code need to be changed accordingly


tophf:
wrapper-webpack-plugin 2.2.2 seems to work.
I'll try rewiring WebpackProtectBootstrapPlugin now.
The options.html page opens but doesn't progress after the first batch of canRender :-)

gera2ld:
> The options.html page opens but doesn't progress after the first batch of canRender :-)

I'm looking into it, might be due to some objects not `reactive`.

src/options/views/tab-installed.vue
```diff
@@ -47,34 +49,36 @@
...
-            <setting-check name="filters.showOrder" :label="i18n('labelShowOrder')" />
-          </div>
-          <div class="mr-2c">
-            <setting-check name="filters.viewTable" :label="i18n('labelViewTable')" />
-            <setting-check name="filters.viewSingleColumn" :label="i18n('labelViewSingleColumn')" />
-          </div>
+          <template #content>
+            <div>
+              <locale-group i18n-key="labelFilterSort">
+                <select :modelValue="filters.sort.value" @update:modelValue="onOrderChange">
```
tophf, line 60: Changing this option doesn't work now...

tophf:
There's still another problem with the sort option: it always shows "execution order" on page load/reload even though the actual order is correct.

tophf:
There are also multiple warnings in the editor and Ctrl-F doesn't work:

```
[Vue warn]: Property "value" was accessed during render but is not defined on instance. 
  at <ToggleButton modelValue=false onUpdate:modelValue=fn > 
  at <VlTooltip content="Use Regex" > 
  at <Code class="abs-full" value="" ref="code"  ... > 
  at <Index key=1 initial= {config: {…}, custom: {…}, meta: {…}, props: {…}, $cache: {…}, …} onClose=fn > 
  at <TabInstalled class="tab flex-auto" > 
  at <KeepAlive> 
  at <App>
```

package.json
```diff
@@ -26,36 +26,31 @@
...
     "gulp-plumber": "^1.1.0",
-    "husky": "^4.2.3",
+    "husky": "^8.0.1",
     "js-yaml": "^4.1.0",
     "jsdom": "^20.0.0",
-    "npm-run-all": "^4.1.5",
     "plugin-error": "^2.0.0",
+    "postcss-scss": "^4.0.4",
+    "postcss-simple-vars": "^6.0.3",
     "sharp": "0.30.1",
```
gera2ld, line 50: Just curious, why is the version of `sharp` locked?

tophf: Because transparency+resize is bugged in all subsequent versions. I've reported the bug upstream.

tophf:
`confirm` page doesn't show the code...

tophf:
Trying to debug it, I see that sourcemaps are completely wrong: it counts from the first line of the .vue file instead of the actual line 38.

gera2ld:
> I see that sourcemaps are completely wrong: it counts from the first line of the .vue file instead of the actual line 38.

Indeed. I found a workaround by shifting `<script>` before `<template>`.

tophf:
Shifting script tag still seems to be off-by-1... I guess it's a bug in Vue3?

gera2ld:
I think it's related to our configuration. Only files under `src/common/ui` are affected. The issue is gone if I copy the same file to another directory, e.g. copy `src/common/ui/code.vue` to `src/options/views` and import it, the line numbers will be as expected.

tophf:
Looks like a bug in Vue3 after all because its loader plugin doesn't pass `pad: true` option.
I've tried forcing it in webpack.conf.js:

```js
require('@vue/compiler-sfc').parse = (orig => function (source, opts, ...rest) {
  if (opts && opts.pad == null && opts.sourceMap) {
    opts.pad = true;
  }
  return orig.call(this, source, opts, ...rest);
})(require('@vue/compiler-sfc').parse);
```

Unfortunately, it exposed another bug inside Vue's padContent function, which breaks the output code...

tophf:
Catching up with master is painful due to the changed URL prefix from `#` to `@`. It seems like the only problem here is that sourcemaps in SFC mode are completely messed up so debugging is essentially impossible unless I disable sourcemaps altogether, which is a no-go as it's counter-productive. Moving `<script setup>` to the top doesn't help. I couldn't understand what's wrong in Vue last time I tried. Maybe you can fix it? Otherwise I think we'll have to put a lid on the SFC syntax, and undo it in externals.vue and toggle-button.vue.

gera2ld:
I think it has nothing to do with `<script setup>`. All components in `src/common/ui` have incorrect source maps. It works if I move the file out of that directory without touching the content. I'm not sure what is different in that directory though. Still checking.

tophf:
One difference is that it goes into a separate chunk `common-ui`. Maybe WrapperWebpackPlugin somehow mess it up?

gera2ld:
It seems to be `FakeDepLoader` that messes the sourcemap up.

tophf:
ShowMessage is broken with the new vueleton, which doesn't seem to have `show` method now:
TypeError: vueleton_lib_modal__WEBPACK_IMPORTED_MODULE_2__.default.show is not a function

tophf:
Tests failed here because MessagePort is not defined. I don't understand how it's even possible for a built-in type in nodejs, moreover the same test runs fine in the master branch. Maybe there's some option that disabled parts of nodejs?

gera2ld:
The tests failed because of jest, let me fix it.

jsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
+    "jsx": "preserve",
     "paths": {
-      "#/*": ["./src/*"]
+      "@/*": ["./src/*"]
```
tophf, line 6: Why is it necessary? It changes so many files, which is painful when merging from master.

gera2ld: Because there are webpack loaders only supporting `@` but not `#`. Some of them were confirmed as bugs and fixed but I've encountered several times. So I want to follow the popular tools like vite, to avoid such cases for good.

gera2ld: I think we can make a release this weekend to fix the remaining bugs, and then merge this MR to get rid of the alias issue.

tophf: I wonder if we can change to @ in master before merging so that we don't feel pressured to merge?

tophf: I mean "pressured to merge this PR earlier" :-)

tophf:
Another moderately annoying problem when debugging in devtools:
1. pause on a breakpoint
2. hover variables in the scope that are still alive at this point

Vue3|master
-|-
There's no tooltip in this branch:<br><br> chrome_2022-10-01_18-56-36 | but it's shown in the main branch:<br><br> chrome_2022-10-01_18-57-51

The same happens in older versions of Chrome.

tophf:
I've bisected chromium builds and found [this commit](https://chromium.googlesource.com/chromium/src/+/b3c794969b499b588fea13020080ab47f86fc576%5E%21/#F0) from Chrome 79 that added support for treating js files as typescript/jsx, so here's my thoughts on the culprit:
1. a different bug in devtools support for typescript
2. a bug in this PR's config that enables typescript - maybe it should be enabled only in ts files? or it should be limited by regex in plaid.conf/webpack.conf for files where we actually use ts?
3. a bug in Vue3

gera2ld:
I think this could be a Chrome issue, it happens on latest Chrome (106) and Edge (Chrome 105), but not on Firefox and Vivaldi (Chrome 104). Not sure what is different from Vue 2 / Webpack 4 though.
FYI TypeScript is not enabled now since `@babel/preset-typescript` is not included in .babelrc.

tophf:
It doesn't work in Chrome 108.0.5329.0 for me. Chrome 78 and earlier works. While it indeed looks like a bug in Chrome but it's triggered by something in this PR.

tophf:
Lol it works in 108.0.5330.0...

tophf (approved):
I guess it's fine because I've been using it all this time.

tophf:
Yay!!! I found a bug though: this PR no longer runs in old browsers due to `{...obj}`. Looks like our `browserlist` and/or `babelrc` is not applied. I didn't investigate it, so my guess would a bug in some of the upstream tools like plaid.

gera2ld:
It seems to be vueleton compiled targeting newer browsers. Let me fix it.

tophf:
I guess we should apply our own settings to everything, maybe even zipjs: #1586.

gera2ld:
https://github.com/violentmonkey/violentmonkey/blob/d1a821c283ea7083a3c3d3affb7c2b5b8b9b9dc8/scripts/webpack.conf.js#L108-L112

You can add any package that is supposed to be compiled here.
But I'm against compiling everything because it's slow and sometimes we get unexpected errors when trying to compile ES5 code with Babel.

PR #1546, opened 2022-08-09
fix: set define to undefined even when grant is not none

gera2ld:
Fix a regression introduced by #1532 : UMD requirements are broken in pages using RequireJS when `@grant` is not none.

## Reproduction

- Page: <https://www.skype.com/en/get-skype/>
- Script:

```js
// ==UserScript==
// @name        Test UMD require
// @match       https://www.skype.com/en/get-skype/
// @grant       GM_addStyle
// @version     1.0
// @require     https://cdn.jsdelivr.net/combine/npm/@violentmonkey/dom@^2,npm/@violentmonkey/ui@^0.7
// @author      -
// @description 8/9/2022, 6:17:32 PM
// ==/UserScript==

console.log('ok');
```

### DIFF
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -291,7 +291,7 @@ function prepareScript(script) {
       IS_FIREFOX ? `,${dataKey}){try{` : '){'}${
       grantNone ? '' : 'with(this)with(c)delete c,'
     // hiding module interface from @require'd scripts so they don't mistakenly use it
-    }((${grantNone ? 'define,module,exports' : ''})=>{`,
+    }((define,module,exports)=>{`,
     ...reqsSlices,
     // adding a nested IIFE to support 'use strict' in the code when there are @requires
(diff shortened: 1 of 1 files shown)


tophf (approved):
Yeah I thought about it and made the wrong decision although your demo script prints ok for me without errors. The reason I removed the dummy variables was that since our resolveProp uses a list of globalKeys that existed at document_start, there'll be no define/module/exports. On the other hand, I forgot that in Firefox the content scripts are injected on extension install/update so these keys will be present in our globalKeys.

gera2ld:
Script updated, there was a cache problem.

tophf:
Now I understand my mistake: userscripts have direct access to `this`, so they can modify it directly, and we can't prevent that.

tophf:
On the other hand, we can try to serve a patched Function.prototype as well as all the other means to evaluate a js string and execute it as `wrapper::func`.

PR #1582, opened 2022-09-22
Fix script list reorder

gera2ld:
The event listener is not called.


tophf:
The bug was added in 28d79430, where I switched to a single listener but forgot about this relay event. It's no longer needed now because we can just call `onDrop();` directly.

tophf (approved):

PR #1584, opened 2022-09-24
Move recycle bin to a new tab

### LINKED ISSUE #1583: Improve Recycle bin navigation and discoverability
- [x] I have searched for existing issues that already reported this problem and found none

**Describe the bug**
Viewing the garbage prevents the user from viewing the installed scripts. The `scripts` url hash must be removed to get back to installed scripts

**To Reproduce**
1. Go to garbage
2. Click on installed scripts
3. Observe how it does not go back

**Expected behavior**
Go back to installed scripts

**Actual behavior**
Does not go back to installed scripts

**What is the result in the [upcoming release](https://github.com/violentmonkey/violentmonkey/releases)?**
The same

**Environment:**
- OS: macOS
- Browser: chrome
- Browser Version: 105.0.5195.125
- Violentmonkey Version: v2.13.1

gera2ld:
close #1583

~~blocked by #1544~~


tophf:
Instead of reusing the standard and obvious navigation that we already have, this will keep the standard navigation broken will make the UI even more cryptic due to a third method of navigating, the presence of which is completely inexplicable to the user because the recycle bin page looks the same as all the others, not a modal dialog.

gera2ld:
Yes, reasonable. It was there because I took it as a special label to filter out scripts marked as removed. Apparently it was a bad design.

tophf:
Ah, I see, it was a great idea, FWIW, and we even have a feature request for tags, so maybe a better solution than a new tab would be converting the recycle bin to such a tag?

tophf:
Ah, I see, it was a great idea, FWIW, and we even have a feature request for tags, so maybe a better solution than a new tab would be converting the recycle bin to such a tag? But it should be done after Vue3 is merged, probably.

I don't know how it would look though. Maybe like github issue filtering?
Currently active tags usually are displayed immediately above the container and they have an 'x' to dismiss.

gera2ld:
I think removed scripts should be treated differently. They should not occupy the namespace/name identifier. And we won't show installed and removed scripts together.
So maybe a new tab is better for the removed scripts.

tophf:
Indeed, filtering via labels/tags assumes the thing will be shown in the main list when no label is applied. On the other hand some users may like it... LOL I just don't know.

tophf:
Wow, we should probably merge Vue3 already :-)

The discoverability of this feature for new users should be improved somehow, especially if they maximized the browser on a big display where the small label is hard to notice. Maybe use standard color and size on the tab title when the trashcan is filled. Maybe keep the old button in the toolbar so that the existing users can find it, otherwise they are likely to report it as a bug. Maybe animate the label somehow e.g. let the script element fly over there :-)

The underscore in the `recycle_bin` route is unusual, how about `#removed`?

Clicking the trash icon in the script card does nothing, I have to reload on chrome://extensions page to see the change, the console is empty - same for the "restore" icon inside the recyclebin tab.

The `removed` parameter is `getScripts` may need additional handling when a new script is installed with the same name/namespace as one in the trash, otherwise when the user restores the old script it will be in conflict.

gera2ld:
Maybe we should put the tab name right below "Installed Scripts"?

 The buttons do not work yet, still wip.

gera2ld:
My expected flow is:

- namespace conflicts are allowed in the trash, so we can create a new script and remove it, and then create another new script with the same namespace/name and remove it
- when restoring a script, it's similar to create a new script and save, we check namespace by then and ask the user to choose to overwrite or use a different name

tophf:
> Maybe we should put the tab name right below "Installed Scripts"?

That was my initial idea, but maybe it'll be too much exposure :-) I don't know now. Either way we can change it later based on users' reaction and our own experience.

> we check namespace by then and ask the user

Yeah, it's good, as long as it indeed works like that, I just wasn't sure it does.

tophf:
Oh, `recycle_bin` -> `recycleBin` maybe?

gera2ld:
`recycleBin` vs `trash`, which do you prefer? @tophf

tophf:
I think it should match the label.

tophf:
Wow, that's an impressive rewrite.

I wonder if this scenario can be improved: when we click a removed script inside the recycle bin, nothing happens. A new user might be confused about the need to click the "restore" icon. Maybe show a confirmation to restore the script? Ideally though I would like to open some kind of small readonly preview in CodeMirror with a button to restore and maybe "delete forever".

gera2ld:
Good idea, we can trigger a full installation on restore. Maybe it's better to add a checkbox to suppress the confirmation in case there are a lot of scripts to restore (e.g. due to unexpected sync removal).

- [ ] trigger a full installation with a confirmation on restore
- [ ] add a checkbox to suppress the confirmation
- [ ] show a readonly edit window when clicking a removed script

src/options/views/script-item.vue
```diff
@@ -281,6 +285,8 @@ $removedItemHeight: calc(
 
 .script {
   position: relative;
+  display: grid;
+  grid-template-columns: $iconSize 1fr auto;
```
tophf, line 289: It requires Chrome 57, so minimum_chrome_version and browserlist should be changed accordingly.

gera2ld: Ah yes, I'll do that since users are less likely to stick to an old version of Chrome.

tophf: BTW, last time I checked AMO statistics, FF 52 was still used by a few hundred of users, probably because it's the last version for Windows XP. As for Chrome, its last XP-compatible version is 49, so we can raise our minimum to something more sensible, for example 86 used by the current release of Centbrowser.

tophf: Chrome 80 implements a [lot of good stuff](https://caniuse.com/?compare=chrome+55,chrome+80&compareCats=all#:~:text=prefers%2Dcolor%2Dscheme) natively (optional chaining, spreading in objects, flat, flatMap, SVG favicon, `prefers-color-scheme`, `#rrggbbaa` to name a few), so that would make sense as a minimum if we bump FF minimum to 68, which was a ESR in 2019-2020.

gera2ld: I tested on Firefox 52 and it worked well. I think we'd better keep supporting Firefox 52 to make things simple (without the need to create a separate version for FF52).
As for the new features, most of them are supported through Babel and we don't feel it so I guess we can keep it this way.

tophf: What about `minimum_chrome_version` in manifest.json? It's necessary to forbid installing the extension in old Chrome otherwise it would show the script list incorrectly.

tophf: 1. Hmm, just as I feared `grid` is slower than `flex`. The first frame is now shown in ~380ms on average whereas previously it was in ~345ms. There's always now a blank paint frame when opening the dashboard (or reloading this tab). It's too much of a regression for such an insignificant convenience for us when writing CSS, so unless this can be fixed I'm against grid.
2. The edit button `</>` is now shown in table view mode, but it shouldn't be.
3. The script names in the table view mode are blue now, should be standard color.

gera2ld: > There's always now a blank paint frame

I don't really see this, even without `batchRender.limit` I got everything out in the first paint frame. (~80 scripts)
image

I think it's totally OK for the dashboard to have such a delay if there is no flashing.

tophf: There's no empty frame anymore but the performance is still considerably worsened (the time to paint the first meaning frame) from ~345ms to ~380ms.

tophf: Re-tested in my main browser with less scripts (60) than in Canary (135) and the first paint frame time is the same as `master`, no regression.

tophf:
Bug: space for action icons is not reserved in table view mode so everything shifts inside due to lazy render.

tophf:
Fixing the table view has broken the alignment of size and trashcan icon in the default view (rows 1 and 3):
image

gera2ld:
I'm thinking about removing `batchRender.limit` as it makes no sense to me and adds unnecessary complexity to the code. Everything can be rendered quickly in the first paint for me in different browsers (Chrome, Firefox, Chromium families, etc.).

tophf:
You don't see any difference only because you don't have a lot of scripts or your computer is exceptionally fast. I see it even with 60 scripts on i7 4GHz. Some people have much more than a hundred of scripts and they have much slower notebooks.

gera2ld:
Shall we merge this first since the logic is the same as the existing one? I'll leave the confirmation to the future.

src/background/utils/db.js
```diff
@@ -208,20 +210,20 @@ export function getScriptById(id) {
...
-    script = store.scripts.find(({ props }) => uri === props.uri);
+    script = store[removed ? 'removedScripts' : 'scripts'].find(({ props }) => uri === props.uri);
   }
   return script;
 }
 
 /** @return {VMScript[]} */
 export function getScripts() {
-  return store.scripts.filter(script => !script.config.removed);
+  return store.scripts;
```
tophf, line 226: Previously it returned a copy, so I wonder if we should use `[...store.scripts]` here (seems wasteful) or maybe `[...getScripts()]` in plugin/index.js?

src/background/utils/db.js
```diff
@@ -379,7 +380,7 @@ function reportBadScripts(ids) {
  * @return {Promise<{ scripts: VMScript[], cache: Object }>}
  */
 export async function getData({ ids, sizes }) {
-  const scripts = ids ? ids.map(getScriptById) : store.scripts;
+  const scripts = ids?.map(getScriptById) ?? [...store.scripts, ...store.removedScripts];
```
tophf, line 383: `[...store.scripts, ...store.removedScripts]` is duplicated in 4 places, so I'd use a function like this:

```js
export function getScriptsByIdsOrAll(ids) {
  return ids
    ? ids.map(id => scriptMap[id])
    : [...dbScripts, ...dbRemovedScripts];
}
```

I can open a new PR with similar code cosmetics after merging this one.

tophf (commented):
Works fine, but I didn't test it a lot. I think we should release a new stable version first, then merge this and release a beta.

PR #1594, opened 2022-10-03
fix: HttpRequest msg.data context

dmarto:
_Disclaimer: I am not well versed in webexts, so I am not sure if this is the proper fix._

After update to `v2.13.1`, I noticed that some functionality with `GM_xmlhttpRequest` was broken.
(I did test latest master commit and latest beta release, both had the issue as well.)

In particular, this simplified excerpt stopped working:

```js
var url = "...";
var formData = new FormData();

formData.append("..", var1);
formData.append("..", var2);
formData.append("..", var3);
formData.append("..", var4);

GM_xmlhttpRequest({
    method: "POST",
    url: url,
    data: formData,
    onload: function(r) {
        ...
    }
});
```

leading to the following error in the browser console:

```
Error: Not allowed to define cross-origin object as property on [Object] or [Array] XrayWrapper
```
at the following line of code https://github.com/violentmonkey/violentmonkey/blob/7f2a6dff07c45fccc27bb4b0ad51f983781d0b80/src/injected/content/requests.js#L38

https://github.com/violentmonkey/violentmonkey/commit/86887363ae5cdc4e27769984d01640f556f2e6d1 is the first bad commit (in which the example stopped working), if we are to trust git bisect 

---

I am using Firefox 105.0.1 (64-bit), have not tested this under any other browser or version, but this change fixes my scripts. 

P.S I really like the newest version :)

### DIFF
--- a/src/injected/content/requests.js
+++ b/src/injected/content/requests.js
@@ -2,7 +2,6 @@ import bridge from './bridge';
 import { getFullUrl, makeElem, sendCmd } from './util';
 
 const {
-  cloneInto,
   fetch: safeFetch,
   FileReader: SafeFileReader,
   FormData: SafeFormData,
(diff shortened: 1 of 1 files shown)


tophf (approved):
Thanks for a good catch and indeed it's the correct solution for Firefox, but Chrome doesn't have/need cloneInto and we already used it in similar cases, so I've generalized our code and adapted the PR accordingly.

PR #1601, opened 2022-10-09
Separate "Update" into two strings + rename "buttonUpdate" to "buttonCheckForUpdates"

sn-o-w:
`Update` is used in two cases:
- As label of the update section in settings
- As command in popup "..." menu for a script (first introduced in BETA v2.13.1.2)

In Romanian it would be as a noun and a verb, so we need two different translations.
And I'm pretty sure other languages might need this as well, for instance, French.

Hope everything's okay with this PR. 😊

### DIFF
--- a/src/popup/views/app.vue
+++ b/src/popup/views/app.vue
@@ -180,7 +180,7 @@
            tabindex="0"
            @click="onRemoveScript"/>
       <div v-if="!activeExtras.data.config.removed && canUpdate(activeExtras.data)"
-           v-text="i18n('titleScriptUpdated')"
+           v-text="i18n('buttonUpdate')"
            tabindex="0"
            @click="onUpdateScript"/>
     </div>
--- a/src/_locales/cs/messages.yml
+++ b/src/_locales/cs/messages.yml
@@ -1,6 +1,9 @@
 buttonCancel:
   description: Cancel button on dialog.
   message: Cancel
+buttonCheckForUpdates:
+  description: Button to check a script for updates.
+  message: Zkontrolovat aktualizace
 buttonClose:
   description: Button to close window.
   message: Zavřít
@@ -90,8 +93,8 @@ buttonUndo:
   message: ''
   touched: false
 buttonUpdate:
-  description: Check a script for updates.
-  message: Zkontrolovat aktualizace
+  description: Button to update a script.
+  message: ''
 buttonUpdateAll:
   description: Check all scripts for updates.
   message: Zkontrolovat aktualizace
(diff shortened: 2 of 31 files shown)


src/popup/views/app.vue
```diff
@@ -180,7 +180,7 @@
            tabindex="0"
            @click="onRemoveScript"/>
       <div v-if="!activeExtras.data.config.removed && canUpdate(activeExtras.data)"
-           v-text="i18n('titleScriptUpdated')"
+           v-text="i18n('buttonUpdate')"
```
gera2ld, line 183: Should this be `buttonCheckForUpdates`?

gera2ld (changes requested):
Thanks for your effort. Can you explain why `buttonUpdate` is renamed to `buttonCheckForUpdates`?

gera2ld:
`titleScriptUpdated` is still used in https://github.com/violentmonkey/violentmonkey/blob/1c3db7beb97249603f996c3f1bcd1f76a674cda5/src/background/utils/update.js#L137-L143
I guess it need to be updated there too.

src/popup/views/app.vue
sn-o-w: Nope, because is this button used as direct action.

r235234t

So technically, all languages except English should add new translations for this action.

Given that the former `buttonUpdate` is now renamed as `buttonCheckForUpdates`, we should be able to submit new translations for this new `buttonUpdate` via Transifex, right? 🤔

sn-o-w:
> `titleScriptUpdated` is still used in
> 
> https://github.com/violentmonkey/violentmonkey/blob/1c3db7beb97249603f996c3f1bcd1f76a674cda5/src/background/utils/update.js#L137-L143
> 
> 
> I guess it need to be updated there too.

Whoops, I could have sworn I was going to miss something. 😆

Should be okay now, I also modified the description for `optionUpdate` -> **Label of the update section in settings and notification title for script updates.**

src/popup/views/app.vue
sn-o-w: @gera2ld I mean the button always showed "Update" since BETA v2.13.1.2, not "Check for Updates", so I wanted to let it as is, but to be able to add a new translation for it.

gera2ld: Ah I see what you mean. Could you run `yarn i18n` to normalize the translation files?

sn-o-w: Done. 👍

gera2ld (approved):
Thank you!

PR #1604, opened 2022-10-09
fix: more reliable commandsForSelf check

tophf:
Instead of manually maintaining the list in `commandsForSelf`, which is error-prone and easy to forget, introducing two methods  addOwnCommands and addPublicCommands, which explicitly indicate the security scope.


gera2ld (approved):
LGTM

PR #1623, opened 2022-10-17
fix: OUJS update check

### LINKED ISSUE #1622: OUJS Update Issue
Hello, for some time I've been experiencing an issue with updating scripts through OUJS. An admin there is pretty confident it is a VM issue and encouraged me to post here to have you look into it. Our discussion is here: https://openuserjs.org/discuss/Scripts_not_properly_installing

tophf:
fixes #1622

### DIFF
--- a/src/background/utils/update.js
+++ b/src/background/utils/update.js
@@ -2,6 +2,7 @@ import {
   compareVersion, getScriptName, getScriptUpdateUrl,
   i18n, sendCmd, trueJoin,
 } from '@/common';
+import { METABLOCK_RE } from '@/common/consts';
 import { fetchResources, getScriptById, getScripts, parseScript } from './db';
 import { parseMeta } from './script';
 import { getOption, setOption } from './options';
@@ -91,6 +92,7 @@ async function downloadUpdate(script, urls) {
   try {
     const { data } = await requestNewer(updateURL, {
       cache: 'no-cache',
+      // Smart servers like OUJS send a subset of the metablock without code
       headers: { Accept: 'text/x-userscript-meta,*/*' },
     }) || {};
     const { version } = data ? parseMeta(data) : {};
@@ -98,7 +100,8 @@ async function downloadUpdate(script, urls) {
       announce(i18n('msgNoUpdate'), { checking: false });
     } else if (!downloadURL) {
       announce(i18n('msgNewVersion'), { checking: false });
-    } else if (downloadURL === updateURL) {
+    } else if (downloadURL === updateURL && data?.replace(METABLOCK_RE, '').trim()) {
+      // Code is present, so this is not a smart server, hence the response is the entire script
       announce(i18n('msgUpdated'));
       return data;
     } else {


Martii:
Thanks @tophf . 

While it's not usually done this working example *(and others similar to it in the wild)* might break the logic too *(depends on regular expression context of that identifier which I haven't looked up yet)*:

``` javascript
// ==UserScript==

alert('Hello, World from a valid .user.js snippet!');

// @name          RFC 2606§3 - Hello, World!
// @namespace     http://localhost.localdomain
// @description   JavaScript alert box saying Hello, World!
// @copyright     2007+, Marti Martz (http://userscripts.org/users/37004)
// @license       CC-BY-NC-SA-4.0; https://creativecommons.org/licenses/by-nc-sa/4.0/legalcode
// @license       GPL-3.0-or-later; http://www.gnu.org/licenses/gpl-3.0.txt
// @version       2022.02.14.0
// @icon          https://raw.githubusercontent.com/Martii/UserScripts/master/res/world.png
// @author        Marti Martz <somewhere@example.com> (https://openuserjs.org/users/Marti)

// @include   http://www.example.com/*
// @include   http://www.example.net/*
// @include   http://www.example.org/*

// @updateURL https://openuserjs.org/install/Marti/RFC_2606§3_-_Hello,_World!.user.js
// @downloadURL https://openuserjs.org/install/Marti/RFC_2606§3_-_Hello,_World!.user.js

// ==/UserScript==
```

See also:
* [Knowing Your Own Metadata](https://sourceforge.net/p/greasemonkey/wiki/Metadata_Block/#knowing-your-own-metadata) with any of the available restructuring methods.

tophf:
Yes, but have you seen such scripts?

tophf:
Anyway, not a problem even if such a script exists, because when this `if` fails, the script will be simply re-downloaded fully.

Martii:
Yes... myself, I think sizzle has done it in the past, and currently [jesus2099](https://openuserjs.org/users/jesus2099) does it for sure on a few of his. Can't recall all instances but the wild is big on the web. :smile_cat: 

I can understand if VM doesn't want to support the structure fully however because the issue will resurface under these use cases that makes it a vector again. Even our OUJS libraries can be multiplexed like this between the different types and optionally with Code.

This is an improvement for the "status quo" though.

tophf:
While it's not a problem in this check (the worst outcome is still a correct re-download), I wonder how we should deal with intentionally "spoofed" meta like this:

```js
const foo = `
// ==UserScript==
// @name        not-foo
// @grant       none
// ==/UserScript==
`;
// ==UserScript==
// @name        foo
// @grant       none
// ==/UserScript==
```

or 

```js
// ==UserScript==
const foo = `
// @name        foo
`;
// @grant       none
// ==/UserScript==
```

Martii:
> I wonder how we should deal with intentionally "spoofed" meta like this

Well the regular expression *(re)* portion of `\n)\s*` will just snip out the first metadata block contents and return all the remainder of the Code. Then the conditional return will be the remainder showing that there is "some Code" *(whether valid or not)* and redownload it.

As far as:

``` js
// ==UserScript==
const foo = `
// @name        foo
`;
// @grant       none
// ==/UserScript==
```

... goes that's technically not spoofed i.e. it's a valid representation of the metadata block and how "Knowing your own metadata" works and how it "could" work in a .user.js *(or JavaScript in general)*.

The issue I'm mentioning is that since the Code is multiplexed *within* the metadata block and none outside of the metadata block *(for any reason)* then it will return a false positive with the optional chaining and the vector comes back into play.

tophf:
> The issue I'm mentioning

Yeah, and I mean it's not an issue, because the worst thing that happens is such a script is re-downloaded correctly.

Martii:
Ahh I see your point now... `true` && `false` *(technically empty string)*. Thanks. Been a while since I've exercised these neurons. :smiley: \*creaks some more\*

tophf:
Cherry-picked as 0118be18.

PR #1633, opened 2022-10-24
fix: calc number of columns resiliently

### LINKED ISSUE #1632: Multiple columns aren't enabled in Centbrowser
in catxp browser，it shows two in one line。but in cent browser，it shows one in one line。and i did not find options in setting。how i became two in one line
QQ截图20221024131538
QQ截图20221024131603

tophf:
Fixes #1632 and adds a bonus for non-bugged browsers: no style recalc due to a getComputedStyle call.

This PR conflicts with the recycle bin PR, so which one will we merge first, @gera2ld?

### DIFF
--- a/scripts/plaid.conf.js
+++ b/scripts/plaid.conf.js
@@ -18,6 +18,7 @@ exports.pages = [
       ...options,
       title: 'Violentmonkey',
       injectTo: item => ((item.attributes.src || '').endsWith('/index.js') ? 'body' : 'head'),
+      scriptLoading: 'blocking', // we don't need `defer` and it breaks in some browsers, see #1632
     })),
   },
 }), {});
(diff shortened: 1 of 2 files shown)


tophf:
Is there something I need to change to make merging easier?

tophf:
Hey, I found a simpler fix: all we need is to remove `defer` from our index.html. We don't need it anyway.

gera2ld:
Just out of curiosity, why don't we implement the columns with pure CSS? e.g. using `grid`.

tophf:
1. Grids have changed behavior several times in the browser version range we support we would likely to need complicated workarounds in older browsers. 
2. AFAIK there's no easy solution for grids with variable number of columns that depend on screen width and should be easily customizable via customCSS.
3. AFAIK grids are [much] slower for large number of items, which is why there's a hardcoded number of rows in Chrome. Even flex is slower than `block` layout, I've seen some cases (might be bugs in browsers) where it was like 10-1000 times.

PR #1636, opened 2022-10-25
feat: infer `@supportURL` from homepage

tophf:
...and show it under a new name `Feedback` in the popup extra menu **instead** of homepage to reduce the amount of stuff there, also to match Tampermonkey which only shows "Report bug".

### DIFF
--- a/src/background/utils/script.js
+++ b/src/background/utils/script.js
@@ -4,7 +4,7 @@ import {
 import {
   HOMEPAGE_URL, INFERRED, METABLOCK_RE, SUPPORT_URL, USERSCRIPT_META_INTRO,
 } from '@/common/consts';
-import { mapEntry, nest } from '@/common/object';
+import { mapEntry } from '@/common/object';
 import { addOwnCommands } from './message';
 import { getOption } from './options';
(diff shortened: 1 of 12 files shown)


src/background/utils/script.js
```diff
@@ -121,3 +121,60 @@ export function getNameURI(script) {
...
+      u = u.join('/');
+    }
+  }
+  if (!u) {
+    u = script.meta.namespace;
+    u = /^https?:\/\/(?!tampermonkey\.net\/)/.test(u)
+      && getFullUrl(u).replace(/^https?(:\/\/userscripts)(\.org\/users\/\w)/, 'https$1-mirror$2');
+  }
+  if (u) {
+    script.props[HOMEPAGE_URL] = u;
```
gera2ld, line 160: Can we infer it on demand during the render? I guess we don't need to store the inferred URLs.

tophf: I thought it'd be wasteful to run those regexps on all scripts every time. I guess I can calculate them in GetData when opening the popup, it'll be rare enough.

gera2ld (approved):

PR #1647, opened 2022-11-06
fix: safe-guard window.vmXXX initializers + speedup

tophf:
While fixing the bug described in the title I did a few more fixes/speedups and while doing that I refactored injection-related code considerably for which I'd like at least a cursory review from you, @gera2ld.

* preinject.js now reuses results of each individual `preparedScript` if it's still cached
* all injection-related consts are extracted to safe-globals-shared.js because importing a lot of these consts in like every file has been an increasing PITA for a while.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,9 +11,12 @@ const FILES_WEB = [`src/injected/web/**/*.js`];
 const FILES_SHARED = [
   'src/common/browser.js',
   'src/common/consts.js',
+  'src/common/safe-globals-shared.js',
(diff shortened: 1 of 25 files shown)


tophf:
Since it's mostly mechanical changes maybe I'll just merge it and let the beta testers find the bugs?

gera2ld:
Sure, let's merge it first.

tophf:
I've released a beta yesterday so that I have more time to reflect on this PR in background of my mind :-)

PR #1730, opened 2023-01-21
fix: let body overflow to hide toolbar in Firefox Android

### LINKED ISSUE #1729: Firefox installed script interface on Android side shows exception[BUG]
The last script in the single column view is not displayed fully when the script reaches a certain amount, because Firefox does not automatically hide the address bar in this interface

gera2ld:
fix #1729

Changes:

- The script list has a max-height of `100vh` and is still scrollable. So on small screens, the total height will be `height of header + 100vh`, so we can trigger a body scroll event and hide the toolbar.
- When a script editor is open on a small screen, the script list will be hidden and the editor will have `position: static` to avoid scroll glitches. In the meantime, the height of the editor is set to `100vh + 1px` which is a bit larger than the screen height to trigger a body scroll event.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "codemirror": "^5.65.10",
     "tldjs": "^2.3.1",
     "vue": "^3.2.45",
-    "vueleton": "^2.0.0-5"
+    "vueleton": "^2.0.1"
   },
   "engines": {
     "node": ">=16"
--- a/src/common/ui/style/style.css
+++ b/src/common/ui/style/style.css
@@ -55,7 +55,7 @@
 }
 
 body {
-  overflow: hidden; // avoid flash of viewport caused by tooltip
+  // overflow: hidden; // avoid flash of viewport caused by tooltip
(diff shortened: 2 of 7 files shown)


tophf:
Using a normal window on the desktop it's fine but in a small window there are 4 scrollbars instead of 1. People occasionally tile the windows on the desktop so this situation is not impossible although rare.

image

gera2ld:
In mobile devices the scroll bars should be hidden by default, right? (assuming all screens smaller than 768px are from mobile)

I removed the inner scroll bar for the script list, the drawback is that the user has to scroll to the beginning of the list to see the `aside` menus. So there are 3 scroll bars left and 2 of them are expected. I'll investigate the small one in `aside` later.

FYI the scroll bar on `body` is a must if we want to let Firefox hide its toolbar. However there are situations we cannot remove the inner scroll bar, such as the code editor.

gera2ld:
The vertical scrollbar in nav menu is removed.

PR #1737, opened 2023-01-27
fix: set popup height to 100% in mobile browsers

### LINKED ISSUE #1736: Kiwi Browser options page adaptation[Feature]
**Motivation**

When viewing a running script in the Kiwi Browser browser, it does not fit the screen adequately.
Screenshot_2023-01-27-18-06-28-735-edit_com kiwibrowser browser

**Proposed Solution**

The ability to display more running scripts (although sliding) looks and feels better.
**Use Cases**

tophf:
fixes #1736

Works in my emulator, but maybe it doesn't in a real phone? @gera2ld, please test.
I tested only in Firefox because Kiwi doesn't start in the emulator.

### DIFF
--- a/src/popup/style.css
+++ b/src/popup/style.css
@@ -212,7 +212,6 @@ footer {
 .submenu {
   display: none;
   min-height: 2rem;
-  max-height: 30rem;
   overflow-y: auto;
   background: var(--bg);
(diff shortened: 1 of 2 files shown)


gera2ld (approved):
LGTM

PR #1740, opened 2023-02-01
fix: speed up imports by throttling render

tophf:
Importing a few megabyte zip that has many differences to the current database was very slow if the script list was already shown. It took several seconds when going from 60 to 130 scripts on my fast desktop computer, which means it would be a minute on a mobile phone. With throttling this operation takes just ~1 second as tested on the same old database in a clean install.

P.S. When adding `importing` to `store` I've also moved 3 needlessly(?) separated props from index.js.

### DIFF
--- a/src/options/index.js
+++ b/src/options/index.js
@@ -1,5 +1,7 @@
 import '@/common/browser';
-import { formatByteLength, getLocaleString, i18n, sendCmdDirectly, trueJoin } from '@/common';
+import {
+  formatByteLength, getLocaleString, i18n, makePause, sendCmdDirectly, trueJoin,
+} from '@/common';
 import handlers from '@/common/handlers';
 import { loadScriptIcon } from '@/common/load-script-icon';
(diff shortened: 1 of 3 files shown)


gera2ld:
Maybe we should switch to a virtual list in the future to improve performance and keep the code relatively simple.
It seems it takes quite a long time for Vue to just create v-dom for components.

tophf:
Virtual lists only help if we artificially limit the amount of scripts on-screen to some low number and show just one column with huge margins in the center. This is what hip sites and apps are doing, which makes it practically acceptable, but I don't like it on a conceptual level, and it also means we would abandon users who want to see all their 100 or 200 scripts at once if they have a non-tiny display.

gera2ld (approved):

tophf:
On the other hand the virtual list idea may work if we add a new side-by-side mode, the list on the left of the script editor, which is used in several other userscript managers:

pic

PR #1760, opened 2023-02-19
feat: customize @inject-into in script settings

tophf:
image


tophf:
I've shortened the label to `Injection mode`, also in advanced settings, because it was super long in some languages and the word "Default" didn't add a lot of meaning to it.

tophf:
I've removed the icon and link because other settings don't have any.

image

PR #1775, opened 2023-02-27
chore: strip "default" from labelInjectionMode

tophf:
Weirdly most languages retained the same translation with the word "Default" in #1774 even though they all should have been cleared when I updated the original labelInjectionMode. Either most translators have reused the old message by mistake or it's a bug in transifex.


tophf:
@gera2ld, could you please check if the upload indeed has failed as the [job log says](https://github.com/violentmonkey/violentmonkey/actions/runs/4278453330/jobs/7448193224)?

gera2ld:
Indeed, something went wrong, the updated keys now have empty values. I'll investigate it later.

Issue #1783, opened 2023-03-07
[BUG] Violentmonkey doesn't work with GNU IceCat due to hard coded cookie store ids.


jovaska1337:
### Sequence of actions:
1. Install extension in GNU IceCat
2. Check about:debugging -> Violentmonkey -> Inspect -> Console
3. See error about invalid store id
4. Extension doesn't work at all

### Problem:
Cookie store ids are hard coded in `src/background/utils/tabs.js` and `src/background/utils/requests.js`. makeicecat causes these to be renamed `icecat-default` and `icecat-container-*` due to batch rebranding. This could be fixed by getting the store prefix with `browser.cookie.getAllCookieStores()` when the extension loads and using that instead of hard coding `firefox`. I fixed this for myself by just replacing them manually.

### Expected result:
Extension works without issue.

#### Devtools console contents:
Uncaught (in promise) Error: Illegal cookieStoreId: firefox-default [Violentmonkey]

#### Environment:
- OS: Linux
- Browser: GNU IceCat 102.8 ESR
- Violentmonkey Version: 2.14.0

PR #1784, opened 2023-03-07
Determine cookie store ID prefix dynamically.

jovaska1337:
Determine the cookie store ID prefix on startup dynamically instead of hard coding it as `firefox`. Allows the extension to work on GNU IceCat and fixes https://github.com/violentmonkey/violentmonkey/issues/1783.


tophf:
Thanks, but I already did it in 3509f2a0.

PR #1786, opened 2023-03-15
chore: use numeric version for AMO beta

tophf:
Fixes AMO warning:

> The version string should be simplified because it won't be compatible with Manifest Version 3 and above. The version should be a string with 1 to 4 numbers separated with dots. Each number should have up to 9 digits and leading zeros will no longer be allowed. Letters will no longer be allowed either.

Not sure I did it correctly, please review, @gera2ld.

### DIFF
--- a/scripts/version-helper.js
+++ b/scripts/version-helper.js
@@ -6,11 +6,7 @@ const pkg = require('../package.json');
  * > manifest.version = `${pkg.version}.${pkg.beta}`
  */
 function getVersion() {
-  return `${pkg.version.match(/\d+\.\d+/)[0]}.${pkg.beta || 0}${
-  // Create a beta release with the same code as in stable release.
-  // Used in unlisted version.
-    process.env.BETA ? 'b' : ''
-  }`;
+  return `${pkg.version.match(/\d+\.\d+/)[0]}.${pkg.beta || 0}`;
 }
 
 function isBeta() {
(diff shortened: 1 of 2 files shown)


scripts/version-helper.js
```diff
@@ -6,11 +6,7 @@ const pkg = require('../package.json');
  * > manifest.version = `${pkg.version}.${pkg.beta}`
  */
 function getVersion() {
-  return `${pkg.version.match(/\d+\.\d+/)[0]}.${pkg.beta || 0}${
-  // Create a beta release with the same code as in stable release.
-  // Used in unlisted version.
-    process.env.BETA ? 'b' : ''
-  }`;
+  return `${pkg.version.match(/\d+\.\d+/)[0]}.${pkg.beta || 0}`;
```
gera2ld, line 9: IIRC there is a problem here, if we release `2.14.0` as a listed version, we cannot release `2.14.0` also as a BETA version. That's why I added a suffix.

Unless we don't release the same code to BETA channel?

tophf: I never released the same version in beta and AFAIK it's impossible now because releases intentionally end on `.0` while beta ends on a non-zero number. Even if the js code is the same between channels, manifest.json will be different anyway due to different version.

gera2ld: Oh I already removed the additional BETA release in 8b332ce557a0313701e08396bc23a7e6e82af72a
Then no concern from my side.

PR #1808, opened 2023-04-25
Added an optional queue update script to avoid when there are many scripts, the update domain names of multiple scripts are the same, thus being denied access by the target site

XiaoYingYo:
Added an optional queue update script to avoid when there are many scripts, the update domain names of multiple scripts are the same, thus being denied access by the target site

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
   ],
   plugins: ['jest'],
   rules: {
+    "no-debugger": "off",
     'prettier/prettier': 'off',
     'no-use-before-define': ['error', {
       'functions': false,
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ node_modules/
 /dist-assets/
 .eslintcache
 .env
+/.idea
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -38,6 +38,7 @@ async function jsProd() {
   return require('@gera2ld/plaid-webpack/bin/build')({
     api: true,
     keep: true,
+    minify: true,
   });
 }
 
--- a/src/background/utils/update.js
+++ b/src/background/utils/update.js
@@ -17,13 +17,27 @@ addOwnCommands({
    */
   async CheckUpdate(id) {
     const scripts = id ? [getScriptById(id)] : getScripts();
-    const jobs = scripts.map(script => {
+    let queueUpdates = getOption('queueUpdates');
+    let results = [];
+    let jobs = [];
+    for (let script of scripts) {
       const curId = script.props.id;
       const urls = getScriptUpdateUrl(script, true);
-      return urls && (id || script.config.enabled || !getOption('updateEnabledScriptsOnly'))
-        && (processes[curId] || (processes[curId] = doCheckUpdate(script, urls)));
-    }).filter(Boolean);
-    const results = await Promise.all(jobs);
+      if (urls && (id || script.config.enabled || !getOption('updateEnabledScriptsOnly'))) {
+        if (!processes[curId]) {
+          processes[curId] = doCheckUpdate(script, urls);
+        }
+        if (!queueUpdates) {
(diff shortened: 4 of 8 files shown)


XiaoYingYo (commented):
image

It seems to work very well, for me, it's very satisfying to use this height, because it allows me to read more content at once and find scripts faster

.eslintrc.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
   ],
   plugins: ['jest'],
   rules: {
+    "no-debugger": "off",
```
tophf, line 12: Remove this please.

.gitignore
```diff
@@ -7,3 +7,4 @@ node_modules/
 /dist-assets/
 .eslintcache
 .env
+/.idea
```
tophf, line 10: Remove this please. We don't list artifacts of various IDE. You should do it in your local git config so it applies to all git projects you author.

gulpfile.js
```diff
@@ -38,6 +38,7 @@ async function jsProd() {
   return require('@gera2ld/plaid-webpack/bin/build')({
     api: true,
     keep: true,
+    minify: true,
```
tophf, line 41: Remove this please.

src/background/utils/update.js
```diff
@@ -65,7 +85,7 @@ async function doCheckUpdate(script, urls) {
     if (canNotify(script) && (msgOk || msgErr)) {
       res = {
         script,
-        text: [msgOk, msgErr]::trueJoin('\n'),
+        text: [msgOk, msgErr]:: trueJoin('\n'),
```
tophf, line 88: Remove the space.

src/popup/views/app.vue
```diff
@@ -521,6 +521,13 @@ export default {
     keyboardService.enable();
     // innerHeight may be bigger than 600px in a mobile browser which displays the popup as a fullscreen page
     this.$el.style.maxHeight = Math.min(Math.max(600, innerHeight), screen.availHeight - window.screenY - 8) + 'px';
+    if (chrome && typeof (chrome["settingsPrivate"]) !== "undefined") {
```
tophf, line 524: Remove this.

tophf (changes requested):
1. It should be one setting where 0 disables the limit and uses the default browser behavior.
2. It might make sense to parallelize per hostname not globally

tophf:
You didn't fix it. Open https://github.com/violentmonkey/violentmonkey/pull/1808/files and remove all unrelated changes.
Also what about this:
>It should be one setting where 0 disables the limit and uses the default browser behavior.
It might make sense to parallelize per hostname not globally

XiaoYingYo:
> You didn't fix it. Open https://github.com/violentmonkey/violentmonkey/pull/1808/files and remove all unrelated changes. Also what about this:
> 
> > It should be one setting where 0 disables the limit and uses the default browser behavior.
> > It might make sense to parallelize per hostname not globally

`It should be one setting where 0 disables the limit and uses the default browser behavior.`
mentioned above,Are you referring to parallel downloads?

What should I do if it is 0? What should I do if it is 1? Can you explain in more detail, because I am using the translation function to understand your description, and there may be situations where I have not fully understood

@tophf

tophf:
Currently you have two settings: queueUpdates and queueUpdatesParallel. I suggest using one setting (maybe name it `updateParallel`) which can be 0. Zero means the option is disabled and the update behaves as it did previously, no limit.

XiaoYingYo:
> Currently you have two settings: queueUpdates and queueUpdatesParallel. I suggest using one setting (maybe name it `updateParallel`) which can be 0. Zero means the option is disabled and the update behaves as it did previously, no limit.

OK

tophf:
There are 20 or more unrelated changes, please remove or undo them. You should add `.idea` and `.vscode` in your local git config so it applies to all your projects, see https://www.google.com/search?q=git+global+exclude.

PR #1809, opened 2023-04-26
Add an option to update scripts in batches

XiaoYingYo:
Added an optional queue update script to avoid when there are many scripts, the update domain names of multiple scripts are the same, thus being denied access by the target site


tophf (changes requested):
Please remove all unrelated changes for popup.

XiaoYingYo:
> Please remove all unrelated changes for popup.

About adapting to third-party browsers to lift the limit height of pop-up windows, can I open another pull request and submit it? Excuse me your willing to accept it?

src/options/index.js
```diff
@@ -133,5 +133,8 @@ function initMain() {
     RemoveScripts(ids) {
       store.removedScripts = store.removedScripts.filter(script => !ids.includes(script.props.id));
     },
+    log(obj) {
```
tophf, line 136: Please remove this.

.gitignore
```diff
@@ -6,4 +6,4 @@ node_modules/
 /tmp/
 /dist-assets/
 .eslintcache
-.env
+.env
```
tophf, line 9: Please remove this change by restoring the exact file or by adding a newline at the end.

tophf:
> About adapting to third-party browsers to lift the limit height of pop-up windows

Only if you can do it without adding settingsPrivate permission. Otherwise, no.

XiaoYingYo:
> > About adapting to third-party browsers to lift the limit height of pop-up windows
> 
> Only if you can do it without adding settingsPrivate permission. Otherwise, no.

Ok, I didn't add any new permissions, but instead provided a new option in the settings page, is this allowed?, and if so, can I commit to this pull request in one go?
@tophf

tophf:
> can I commit to this pull request in one go?

No, it's a separate problem, so it should be a separate pull request.

XiaoYingYo:
> > can I commit to this pull request in one go?
> 
> No, it's a separate problem, so it should be a separate pull request.

OK

.gitignore
XiaoYingYo: Hello, this is actually the original file, but I submitted it once in order to ignore a certain file. The last time I submitted this file, it was already the original file in the warehouse
In general, please ignore the modification of this file, it does not actually make any changes, it is just for correction, I submitted this file by mistake earlier

XiaoYingYo: If you still think it has to be uncommittable, then I re-fork the repository and reapply a new pull request

@tophf

src/background/utils/update.js
```diff
@@ -17,13 +17,55 @@ addOwnCommands({
...
+    if (parallel !== 0) {
+      let result = [];
+      for (let host in tasks) {
+        jobs.push((async () => {
+          let Run = [];
+          let hostTasks = tasks[host];
+          for (let i = 0; i < hostTasks.length; i++) {
+            let { curId, script, urls } = hostTasks[i];
+            if (i % parallel === 0) {
+              result.push(await Promise.all(Run));
```
gera2ld, line 51: It seems an empty list will always be added when `i === 0`?

src/background/utils/update.js
```diff
@@ -17,13 +17,55 @@ addOwnCommands({
...
+            }
+            if (!processes[curId]) {
+              processes[curId] = doCheckUpdate(script, urls);
+            }
+            Run.push(processes[curId]);
+          }
+          if (Run.length != 0) {
+            result.push(await Promise.all(Run));
+          }
+          return new Promise((resolve) => {
```
gera2ld, line 62: No need to wrap a Promise here.

src/_locales/zh_CN/messages.yml
```diff
@@ -418,6 +418,9 @@ labelNotifyUpdates:
 labelNotifyUpdatesGlobal:
   description: Option to prioritize global notification option over script's setting.
   message: 忽略每个脚本的提示（编辑器中的“设置”页）
+labelUpdatesParallel:
+  description: 当多个脚本的更新地址为同一个域时,会根据设置的并发数进行分批次更新
```
gera2ld, line 422: `,` -> `，`

gera2ld (commented):
Just curious, what is the benefit of batches over concurrency limit?

src/background/utils/update.js
XiaoYingYo: `i % parallel === 0`
The code here indicates that if the number of tasks currently being updated by this domain name reaches the concurrency limit, wait for the tasks being updated to be updated before continuing

@gera2ld

src/background/utils/update.js
XiaoYingYo: I changed to `return Promise.resolve(result);`

@gera2ld

src/_locales/zh_CN/messages.yml
gera2ld: We only use full-width punctuation for Chinese, please don't change those in English.

XiaoYingYo:
> Just curious, what is the benefit of batches over concurrency limit?

Sorry, I may not be able to understand the specific meaning of your description according to the translation. Let me explain roughly:
If the `parallel` set by the user is not 0, then when updating, a thread pool will be created to queue the update script according to each host according to the value of `parallel` set, which can avoid being a user When there are too many scripts, multiple scripts will appear from the same update URL. If `parallel` is 0, that is, this new function is not enabled, then it is possible that when more than one update URL is requested at the same time, the target site will be regarded as attacker, thus denying access

@gera2ld

src/_locales/zh_CN/messages.yml
XiaoYingYo: OK

gera2ld:
**Batches:**
Separate all scripts into batches, say 10 per batch, wait until all the 10 updates are done, then start the next 10.

**Concurrency limit:**
Start 10 updates and wait, whenever one is done, start a new one.
So the processing scripts will always be 10 and we don't need to wait for all of them to be done to start the next one.

BTW I'm OK with batches here.

src/_locales/zh_CN/messages.yml
```diff
@@ -418,6 +418,15 @@ labelNotifyUpdates:
 labelNotifyUpdatesGlobal:
   description: Option to prioritize global notification option over script's setting.
   message: 忽略每个脚本的提示（编辑器中的“设置”页）
+labelUpdatesParallel:
+  description: 当多个脚本的更新地址为同一个域时，会根据设置的并发数进行分批次更新
+  message: 每个域名并发 $1 个进行更新脚本
+labelAutoPopupHeight:
```
gera2ld, line 424: Please don't change the keys that are not related to this PR.

The `description` doesn't need to be translated. You can run `yarn i18n` to normalize the translations before pushing.

XiaoYingYo:
> **Batches:** Separate all scripts into batches, say 10 per batch, wait until all the 10 updates are done, then start the next 10.
> 
> **Concurrency limit:** Start 10 updates and wait, whenever one is done, start a new one. So the processing scripts will always be 10 and we don't need to wait for all of them to be done to start the next one.
> 
> BTW I'm OK with batches here.

My initial submission was to set batches for the script alone, instead of batching according to the specific host of the script. It was proposed by @tophf before. It may be more meaningful to batch according to the host, so I made improvements, and finally it was this submission

@gera2ld

src/_locales/zh_CN/messages.yml
XiaoYingYo: I'm sorry, I haven't done international project development yet, I don't understand this aspect, can you do it for me directly?

@gera2ld

XiaoYingYo: Did you mean `labelAutoPopupHeight`?
This is another function that I submitted earlier. After consideration, I may need to open another pull request and submit again, so this problem can be ignored

@gera2ld

tophf: You should have removed labelAutoPopupHeight and labelAutoPopupHeightHint from this pull request. I just did it for you, so please synchronize your local repository.

XiaoYingYo: ok thank you, i never understood the real meaning, sorry

@gera2ld

src/_locales/zh_CN/messages.yml
```diff
@@ -418,6 +418,9 @@ labelNotifyUpdates:
 labelNotifyUpdatesGlobal:
   description: Option to prioritize global notification option over script's setting.
   message: 忽略每个脚本的提示（编辑器中的“设置”页）
+labelUpdatesParallel:
+  description: 当多个脚本的更新地址为同一个域时，会根据设置的并发数进行分批次更新
```
gera2ld, line 422: Please run `yarn i18n` to normalize the i18n files before pushing

src/background/utils/update.js
```diff
@@ -17,13 +17,53 @@ addOwnCommands({
...
+            }
+            if (!processes[curId]) {
+              processes[curId] = doCheckUpdate(script, urls);
+            }
+            Run.push(processes[curId]);
+          }
+          if (Run.length != 0) {
+            result.push(await Promise.all(Run));
+          }
+          return Promise.resolve(result);
```
gera2ld, line 62: Why not just `return result`?

gera2ld (changes requested):

src/_locales/zh_CN/messages.yml
gera2ld: FYI the `description` is for the translator (who understands English) to understand what this key is for, so it doesn't need to be translated. Only `message` should be translated.

src/background/utils/update.js
XiaoYingYo: Because Promise has an execution priority relationship, so keep it like this, I think it is OK

@gera2ld

src/_locales/zh_CN/messages.yml
XiaoYingYo: understand, thank you for your description

src/background/utils/update.js
tophf: There's no difference because Promise.resolve doesn't create a new microtick.

XiaoYingYo: OK

@tophf

gera2ld (approved):
LGTM. Thank you for your time!

tophf:
@gera2ld, how about we make this behavior the default using 2 as the limit and remove the option?

tophf:
Well then 1 would work for everything and the code can be simplified greatly. I don't like adding an option for such a rare use case, especially if it's separate from the main update section in settings.

gera2ld:
Indeed. `1` sounds better and no need for an option.

tophf:
@XiaoYingYo, do you agree? And if yes, could you do it in this pull request?

XiaoYingYo:
> @XiaoYingYo, do you agree? And if yes, could you do it in this pull request?

@tophf 

May I ask whether to update the scripts in batches according to the download URL to batches according to the number of scripts?

tophf:
I didn't understand your question :-( please rephrase.

XiaoYingYo:
Should it be changed that if `parallel` is not 0, then the script will be updated according to the number of scripts according to the value of `parallel`, without the need to judge any url?
@tophf

tophf:
There will be no `parallel`, because it will be always 1.

XiaoYingYo:
@tophf 
I'm sorry, but I'm not sure what the final effect should be. Can you give a general overview? What should the specific steps of this update script look like? Are you updating one by one?

tophf:
Simply replace `parallel` with `1`, remove `labelUpdatesParallel` and the option, remove the parts of the code that will never run (parallel === 0).

XiaoYingYo:
@tophf 
I have processed it, please check it

tophf:
Why `5`? Do you think there are no servers that have a lower restriction (like `1`)?

XiaoYingYo:
If it is 1, I think it is too slow, I have a lot of `violent monkey script`, I actually run it once by `1`, I think it takes too much time

@tophf

XiaoYingYo:
@tophf
I have another 210 violent monkey scripts

tophf:
Please tell how long does the update lasts for you with `1`, `2`, `3`, `4`, `5` so we can choose a balanced number.

XiaoYingYo:
@tophf 
This is different from person to person, I think my standard here should not be able to make a real decision, because I am in mainland China, and the network policy or firewall in mainland China will block some foreign websites, so sometimes we will Rely on VPN tools to surf the Internet, in fact, I still think it is best to let users freely decide this value

tophf:
I don't like adding an option if we can choose a good default value. I thought `2` is fine because it's sufficiently fast.

XiaoYingYo:
done
@tophf

tophf:
Cool! BTW I remember now I always wanted to implement this limit at a low number because Firefox locks up the extension when all requests are performed at once.

tophf:
It just occurred to me that `CheckUpdate` is the wrong place for this, because we should also limit requests for `@require` and `@resource`. The correct place to do it is inside [request()](https://github.com/violentmonkey/violentmonkey/blob/master/src/common/util.js#L296) if `isRemote(url)`. And judging from the fact that the goal is to avoid VPN congestion, there's no need to sort by `host`. Instead there should be a global queue for requests with several concurrent slots (2).

gera2ld:
Reasonable. Actually we already have an implementation of concurrency limit here:

https://github.com/violentmonkey/violentmonkey/blob/5ea67ea1b5fc940c675a8bb820e246482ef03621/scripts/transifex.mjs#LL38C1-L38C1

cyfung1031:
> Sounds like your network may be limiting network connections issued in quick succession, which happens when updating all scripts, which will be eventually fixed in https://github.com/violentmonkey/violentmonkey/pull/1809.

@tophf I think it is because the receiver server (greaskfork.org) has some server connection protection to avoid multiple network requests emitted by a single user at the same time. If the network requests send one by one, wait the previous request get responded first, then request the next one, the problem should solve.

cyfung1031:
@XiaoYingYo  did you really test your code change?
obviously `const host = urls && new URL(urls[0]).host;` is wrong. host will be always `true`

Anyway, I would like to contribute this issue as well. I will try to follow your mindset and update the PR.

tophf:
I was waiting for the author to implement the overall limit as suggested in [the last comment](https://github.com/violentmonkey/violentmonkey/pull/1809#issuecomment-1528708049) above, which didn't happen, so I'm closing this PR in favor of #1856.

PR #1847, opened 2023-07-07
add "?filter_locale=0" to show results for all languages

fligger:
Greasyfork.org will by default only show results for your location (in my case germany). 
This is bad as most scripts you want to use are not translated to german and do not show up. 
We always have to click on show all languages to get good results (this annoyes me enough to make this pr). 

This pr adds "?filter_locale=0" to the url and will automatically show all results (you can go back to your locations result by clicking an an link at the top of the page).

### DIFF
--- a/src/popup/views/app.vue
+++ b/src/popup/views/app.vue
@@ -334,7 +334,7 @@ export default {
     findUrls() {
       const query = encodeURIComponent(store.domain);
       return {
-        [`${i18n('menuFindScripts')} (GF)`]: `https://greasyfork.org/scripts/by-site/${query}`,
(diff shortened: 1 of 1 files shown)


tophf:
I don't think we should do it. The filtering was added by Greasyfork in response to a multitude of complaints about thousands of Chinese-named scripts in the listings. You can suggest the author of the site to implement a better search e.g. list the unfiltered results right afterwards.

@gera2ld?

gera2ld:
Maybe we can make it configurable instead of changing the url globally?

tophf:
This is already configurable in greasyfork account settings...

PR #1851, opened 2023-07-17
fix: clarify 'Track local file' checkbox label

SimonAlling:
Before I learned from experience how the checkbox worked, I interpreted the label as if ticking the checkbox and then closing the tab would make Violentmonkey track the file.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -245,7 +245,7 @@ installOptionClose:
   message: Close after installation
 installOptionTrack:
   description: Option to track the loading local file until page is closed.
-  message: Track changes in local file while this page is open
+  message: Track edits in local file while this page is open
 installOptionTrackTooltip:
   description: >-
     Tooltip in Firefox 68+ for the option to track the loading local file before
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -244,8 +244,8 @@ installOptionClose:
   description: Option to close confirm window after installation.
   message: Close after installation
 installOptionTrack:
-  description: Option to track the loading local file until window is closed.
-  message: Track local file until this window is closed
+  description: Option to track the loading local file until page is closed.
+  message: Track changes in local file while this page is open
 installOptionTrackTooltip:
   description: >-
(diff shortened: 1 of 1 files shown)


gera2ld (approved):

tophf:
My problem with `local file` and the current phrasing is that someone who never used this feature wouldn't be able to guess what this option does just by its name, which is why I want to use the word `editor` that immediately describes the concept.

SimonAlling:
I feel like the change I suggested is desired regardless of where the "local file" vs "external editor" discussion might end up, so maybe any such clarification deserves its own PR?

tophf:
I'd like to decide now to avoid creating unnecessary work for volunteers who translate our language files every time we change them. The improvement in this PR is not crucial, the phrasing is still awkward and the choice of terms like "track" or "window" is confusing.

tophf:
What about `Track your edits in the local file while this page is open`, @gera2ld?

gera2ld:
I'm okay with that

SimonAlling:
> Track your edits in the local file while this page is open

Changed it to `Track changes in local file while this page is open`. I used "changes" instead of "edits" because the latter seems to imply the file being manually edited by a human, but the `.user.js` file may in fact just as well be an artifact from some build setup. For example, many of my userscripts are built from TypeScript via webpack etc. Also, I omitted "your" and "the" to keep the label short and succinct.

WDYT?

tophf:
It became more abstract and less self-evident.

tophf:
Even if something may be produced via a build script it is still usually in response to an edit, which is the whole point of this feature.

SimonAlling:
> Even if something may be produced via a build script it is still usually in response to an edit, which is the whole point of this feature.

Would you be OK with `Track edits in local file while this page is open`?

tophf:
Yeah, if it feels natural to you. You're the only native speaker of the language around I guess :-)

SimonAlling:
I'm not a native speaker! :innocent: I'd personally have gone with "changes" for the reason mentioned earlier, but I think "edits" works too. Your call! :slightly_smiling_face:

PR #1868, opened 2023-08-10
feat: reload tab without scripts + icon context menu

### LINKED ISSUE #1796: [Feature] An option to disable Violentmonkey on current tab
**Motivation**
When comparing a website content with and without Violentmonkey enabled, you would need to manually switch it on/off globally. It would be a great solution to have an option that allows you to  switch Violentmonkey on/off per tab.

**Proposed Solution**
A second button next to the global switch on/off button. Maybe not by default, since it would confuse many users as there aren't tooltip info when you hover over the current buttons.

tophf:
Fixes #1796.

image

image

image

The default icon is gray now.
The color icon is displayed when scripts are actually injected.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -808,6 +808,15 @@ sideMenuInstalled:
 sideMenuSettings:
   description: 'Side menu: Settings'
   message: Settings
+skipScriptsCmd:
+  description: Title of the hotkey in the browser UI for customizing the hotkeys.
+  message: Reload page without userscripts
+skipScriptsHint:
+  description: Tooltip shown over the global enabled/disabled switch in popup.
+  message: (right-click to reload the page without userscripts)
+skipScriptsMsg:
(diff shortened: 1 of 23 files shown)


tophf:
Ah, I've also replaced "Visit website" in the popup with homepage_url from manifest.json:
image

gera2ld:
How about adding a `...` icon at the top next to the plus icon and put the extra menu items in a dropdown menu? perhaps hide the cogs in the extra menu too.

My concern for the current solution is that it's hard for the user to know that we can get different options by right-clicking some buttons. TBH I won't read the tooltips carefully for every button. I'd expect a more intuitive UI.

tophf:
Here we go:

image

The 3 old buttons should be visible I think as they all seem equally useful. The extra menu will contain rare commands. I'll add "Blacklist current domain" there later. Maybe also "Update matched scripts".

tophf:
Added `Settings` and `Update scripts (#)` where # is the number of updatable scripts shown in the popup. I think it's ready for a review.

image

PR #1870, opened 2023-08-11
feat: add IDE autocomplete hints for i18n id

tophf:
image


.husky/pre-commit
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+SRC=src/_locales/en/messages.yml
+DST=src/types.generated.d.ts
+
+if git diff --cached -- "$SRC"; then
+  . "$(dirname -- "$0")/_/husky.sh"
+  node scripts/generate-types.js "$SRC" "$DST"
```
gera2ld, line 7: How about running this in npm scripts `prepare` and `i18n` and ignore `types.generated.d.ts` in the repo?

tophf: Why? The point of this PR is that changing `messages.yml` also changes `types.generated.d.ts` in the same commit.

tophf: I also found a problem: it should run in `post-rewrite` and `post-merge`, but in these cases it should probably create a new commit instead of amending the merge...

gera2ld: Since it can be generated on demand, why do we keep it in the repo and maintain such a file manually?

tophf: How else can we make IDE see our message ids?

gera2ld: Here is what I suggest:

- Add `src/types.generated.d.ts` to `.gitignore` so it won't be kept in the repo, but the file can still be there
- Instead of running `scripts/generate-types.js` on commit, I suggest running in the following cases:
  - `yarn prepare` - when we install/update deps
  - `yarn i18n` - when we add new i18n keys
  - `post-checkout` - when we switch to a new branch
- For VSCode we can even [add a hook](https://www.roboleary.net/vscode/2020/10/19/vscode-task-onstartup.html) to always recreate the `.d.ts` file on load

BTW I'm a bit confused why we need the hints since we usually add `i18n('xxx')` first and then generate its translation with `yarn i18n`. Is it for reusing existing keys?

tophf: 1. Can you add a locally generated file to jsconfig.json so that the IDE sees it?
2. What about updating the types to match messages.yml while working on a single branch?
3. I've never used `yarn i18n`, do I have to do it each time I edit `en/messages.yml` and if yes, then why don't we auto-run this task via hooks?

>Is it for reusing existing keys?

Yes. Without autocomplete I have to open the file and search inside and that's highly inconvenient due to the presence of descriptions and the structural keys (description and message) which produce false matches while using incremental search or filter. The lack of such autocomplete might be the reason why there are many duplicates in our messages.

gera2ld: > 1\. Can you add a locally generated file to jsconfig.json so that the IDE sees it?

Yes, as long as the file exists.

> 2\. What about updating the types to match messages.yml while working on a single branch?

How about running it in `yarn i18n` and adding `yarn i18n` to pre-commit hook?

> 3\. I've never used `yarn i18n`, do I have to do it each time I edit `en/messages.yml`

I think so, actually `en/messages.yml` is generated/updated by `yarn i18n` automatically, adding new keys and removing unused keys (only in prod mode).

tophf: So we need to commit an empty `types.generated.d.ts` in the repo in order to make jsconfig.json correct? Anyway, the solution that relies on an ignored file seems inherently fragile.

As for `yarn i18n` **current** behavior, why do I need to run it if I've just added a new entry to messages.yml and used it in the code? As for making it run automatically, pre-commit hook is probably insufficient, we'd also need post-merge and post-rebase.

gera2ld: > So we need to commit an empty `types.generated.d.ts` in the repo in order to make jsconfig.json correct?

No, it is ignored if the file doesn't exist, and picked by tsserver if the file is present. It's just a match pattern instead of an exact file to be imported.

> Anyway, the solution that relies on an ignored file seems inherently fragile.

Check `node_modules/.vite` for Vite, and `.svelte-kit` for Svelte, they are exactly for the same purpose, which contain types generated on demand, and ignored in git. I think this is quite a common use case for TS/JS.

> why do I need to run it if I've just added a new entry to messages.yml and used it in the code

Firstly, `yarn i18n` will add the keys automatically which might be helpful if you create a new view with many new keys. Secondly, it verifies and formats the message.yml files just like `lint`.

> we'd also need post-merge and post-rebase.

I think we can just hook `prepare` and `post-checkout`, and let the developer deal with the edge cases.

tophf: > picked by tsserver if the file is present

I see.

> I think we can just hook prepare and post-checkout, and let the developer deal with the edge cases.

Which edge cases? Merging and rebasing is not an edge case. The point of my PR is to make it as reliable as possible. Ideally it should work on-the-fly, but that's only possible via an IDE plugin, so at the least it should regenerate the types when the change is committed in git.

tophf:
Here's another idea - we won't need any type generation and we'll even see the text in autocomplete!

Instead of messages.yml make `src/common/locale-en.js`:

```
export const i18n = {
  msgFoo: [
    'message',
    'description',
  ],
  ........
};
```

Then use `i18n.msgFoo` everywhere. In the `compilation` hook for webpack we'll replace it with a call for `i18nImpl` function (previously i18n in src/common/util.js) and generate messages.json in the output.

.husky/pre-commit
tophf: BTW I've just tried `gulp i18n` and I really don't like how it modifies 30 files by adding meaningless empty translations for the id I've just added. It bloats my simple commit into a monster in which I can't even see what's changed at a glance.

gera2ld: We can change to only modify `en`, and leave others to the transifex script.

gera2ld:
> Instead of messages.yml make `src/common/locale-en.js`

It seems not easy to find unused keys in this way. And if we generate `messages.json` from this file, it's not easy to tree-shake the unused keys either.

tophf:
> It seems not easy to find unused keys in this way. And if we generate messages.json from this file, it's not easy to tree-shake the unused keys either.

Why? There's no difference because messages.json will be generated from this file. Everything should be exactly the same as it is.

gera2ld:
- Before:
  - All keys are directly used in the source code, and `messages.yml` is maintained by `gulp i18n` which automatically adds new keys and mark unused keys. When adding a new key, we only type it once.
- After:
  - We maintain the keys in `src/common/locale-en.js` and we cannot tell if a key is being used from it.
  - Each key has to be declared before we can use it, i.e. we need to modify two places when we add a key.
  - Since it's a JavaScript file, it's not easy for a script to modify it and read descriptions from the comments.
  - Of course we can do string substitutions but it's not reliable and requires more conventions on how to code in that file, which doesn't seem a good idea to me.

tophf:
The only thing that changes is that en/messages.yml will be auto-generated from locale-en.js during build, so literally everything will be the same in the end.

> We maintain the keys in src/common/locale-en.js and we cannot tell if a key is being used from it.

We can. Both in the IDE via usages and via simple regex like `\bi18n\.\w+` similar to what we use currently.

> Each key has to be declared before we can use it, i.e. we need to modify two places when we add a key.

I don't understand. Nothing will change in this regard. Instead of adding a key in en/messages.yml we will add it in locale-en.js, just one time.

> Since it's a JavaScript file, it's not easy for a script to modify it and read descriptions from the comments.

It's trivial using acorn parser and I do the same for safe-globals.js.

> Of course we can do string substitutions but it's not reliable and requires more conventions on how to code in that file,

I don't understand. If the substitutions for a description's comment then what's so hard? Any comment(s) before a message are combined into a description when uploading to transifex. Substitutions for `i18n.msgFoo` in the code can be implemented reliably via a compilation hook e.g. the way we use babel-plugin-safe-bind.js.

gera2ld:
> We can. Both in the IDE via usages and via simple regex like `\bi18n\.\w+` similar to what we use currently.

I mean, how do we know if a key is unused and we can remove it from `src/common/locale-en.js`? It's not automatic any more.

I don't see much benefit because I don't add keys to `en/messages.yml` manually, and it looks a bit overcomplicated to me.

Anyway I'm not against the idea.

tophf:
1. Most importantly, the build detects the used/unused keys in exactly the same or similar fashion, so the output will be automatically correct.
2. Pruning the unused keys in locale-en.js (and other files) can be done automatically when uploading to transifex.

> I don't add keys to en/messages.yml manually

How else can I specify the text of a new locale message?

> I don't see much benefit

And to me the current system is problematic and inconvenient.

tophf:
> I mean, how do we know if a key is unused and we can remove it from src/common/locale-en.js? It's not automatic any more.

I don't understand what's changed. It can be conceptually the same as the current regex matching in i18n.js but instead of `'\\b(?:i18n\\(\'|i18n-key=")(\\w+)[\'"]'` there'll be `\bi18n\.`. It can be also the much more correct detection via acorn parser.

gera2ld:
For example, if a component is removed and several keys are not used any more, `yarn i18n` will mark the unused keys as `touched: false` or remove them directly. But if the same happens to `src/common/locale-en.js`, do we modify this JS file by string substitution? I'm not talking about the generated `message.json`, instead I'm talking about the source code.

The biggest difference to me is, `messages.yml` is maintained automatically, while the proposed `src/common/locale-en.js` is not.

Regarding the duplicate messages, sometimes it's hard to avoid because in one language a phrase can be reused doesn't mean it can be reused in another language too.

> it was necessary earlier when translation was done in github

That's exactly why they exist. I also mentioned earlier we can skip languages other than `en`.

tophf:
Okay, yarn i18n will add/remove keys in locale-en.js via acorn parser or we can mandate a simple structure of the file, I don't see it as a problem either.

tophf:
Duplicates: "All", "Code", "Feedback", "Homepage", "Recycle Bin", "Settings", "Update". AFAICS only `Update` has two meanings (action/noun), the rest are 100% duplicates in any language.

tophf:
That'll be a different PR though.

PR #1883, opened 2023-09-03
refactor: replace `tldjs` w/ `tldts`

SukkaW:
The PR replaces `tldjs` with `tldts`.

`tldjs` is no longer actively maintained. The latest commit of the `tldjs` repository occurred 5 years ago and the prior npm publish of the `tldjs` package was 6 years ago, while the `tldts` remains actively maintained.

Also, `tldts` is about 2x faster than `tldjs`, see benchmark suite: https://replit.com/@isukkaw/tldjs-vs-tldts

### DIFF
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -2,7 +2,7 @@ import '@/common/browser';
 import { makePause, sendCmd } from '@/common';
 import { TIMEOUT_24HOURS, TIMEOUT_MAX } from '@/common/consts';
 import { deepCopy } from '@/common/object';
-import { getDomain } from 'tldts';
+import { getDomain } from '@/common/tld';
 import * as sync from './sync';
 import { addOwnCommands, addPublicCommands, commands } from './utils';
 import { getData, getSizes, checkRemove } from './utils/db';
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -41,7 +41,7 @@ addOwnCommands({
     const host = url && new URL(url).hostname;
     return {
       host,
-      domain: host && getDomain(host, { allowPrivateDomains: true }) || host,
+      domain: host && getDomain(host) || host,
     };
   },
 });
--- a/src/background/index.js
+++ b/src/background/index.js
@@ -2,7 +2,7 @@ import '@/common/browser';
 import { makePause, sendCmd } from '@/common';
 import { TIMEOUT_24HOURS, TIMEOUT_MAX } from '@/common/consts';
 import { deepCopy } from '@/common/object';
-import { getDomain } from 'tldjs/tld';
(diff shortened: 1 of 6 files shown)


src/background/index.js
```diff
@@ -41,7 +41,7 @@ addOwnCommands({
     const host = url && new URL(url).hostname;
     return {
       host,
-      domain: host && getDomain(host) || host,
+      domain: host && getDomain(host, { allowPrivateDomains: true }) || host,
```
SukkaW, line 44: `tldts` by default does not respect the public suffix list and requires `{ allowPrivateDomains: true }` for manual opt-in.

tophf:
Great! But I don't like the need to remember this new option, which we can easily forget in the future, so I suggest adding `src/common/tld.js` like this:

```js
import {
  getDomain as getDomain_,
  getPublicSuffix as getPublicSuffix_,
} from 'tldts';

const opts = { allowPrivateDomains: true };
export const getDomain = url => getDomain_(url, opts);
export const getPublicSuffix = url => getPublicSuffix_(url, opts);
```

tophf:
I mean I can add that myself, just wondered if you have a better solution.

SukkaW:
> I mean I can add that myself, just wondered if you have a better solution.

Sorry, I went out for dinner after the submission of the PR. I will apply the suggestions now.

src/common/tld.js
```diff
@@ -0,0 +1,10 @@
+import { getDomain as getDomain_, getPublicSuffix as getPublicSuffix_ } from 'tldts';
+
+/**
+ * tldts does not respect the public suffix list by default, but can be opt in manually
+ * with the option `allowPrivateDomains`. Hoist the `sharedOpts` can also help avoid
+ * re-creating the object every time.
+ */
+const sharedOpts = { allowPrivateDomains: true };
```
remusao, line 8: Nit. I noticed that you only seem to call `getDomain` and `getPublicSuffix` on inputs which are already valid hostnames (result of `new URL(url).hostname`). In such cases, you can reduce the amount of work tldts performs by default with some extra options (by default tldts will validate hostnames but also check if it's necessary to extract the hostname from input in case it's a URL). This can be disabled with:

```js
const sharedOpts = {
  allowPrivateDomains: true,
  extractHostname: false, // inputs are already hostnames
  validateHostname: false, // inputs are already valid, no need to perform extra validation
  detectIp: false, // in case you know you don't get IPs as inputs you can also get extra speed by disabling this (not sure if that's something you need for your use-case)
};
```

tophf: Thanks!

SukkaW: Wow! I didn't know that before.

PR #1896, opened 2023-09-17
fix #1894: add `tag` to GM_notification

### LINKED ISSUE #1894: [Feature] Add a `tag` option for GM_notification
**Motivation**

```javascript
// ==UserScript==
// @name        New script - github.com
// @match       https://github.com/*
// @grant       GM_notification
// @grant       GM_openInTab
// ==/UserScript==
'use strict';

(() => {
    // codes...
    const result = { // Simulate the result.
        id: 255,
        text: "New release has been found!",
        ver: "2.5.4",
        title: "Electronic-RRMusic",
        uri: "https://example.org/electronic-rrmusic/releases/v2.5.4"
    }
    GM_notification({
        onclick: () => GM_openInTab(result.uri),
        tag: result.id, // want this
        ...result
    });
})();
```
I want to use the `tag` option, [just like the legacy API](https://developer.mozilla.org/docs/Web/API/Notification/Notification#tag)

**Proposed Solution**

I don't want to use the Legacy API, because the Notification will turn into a "ghost" notification after user close the tab. (`onclick` function was destroyed, so nothing will happen when you click it)

**Use Cases**

To update the notification content.

tophf:


tophf:
Seems trivial enough for me to go ahead and merge it.

PR #1903, opened 2023-09-22
chore: upgrade to node@18

gera2ld:


cyfung1031:
@gera2ld specifying the node version is not a feature in v2. You have to upgrade them to v3.

Currently there are still depreciation messages in every CI build.
Screen Shot 2023-10-14 at 21 09 17

Also, [`set-output` shall be changed to `GITHUB_ENV/GITHUB_OUTPUT/GITHUB_STATE` as well](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

gera2ld:
@cyfung1031 thanks for the information.

PR #1906, opened 2023-09-24
feat: show option to allow edits in updated scripts

### LINKED ISSUE #1901: [Question] Regarding the "Allow Updates", how to make it like the previous version?
Just a newbie to use Violentmonkey BETA. (v2.15.4)  

In previous version there is no such warning.

I want to edit the script downloaded from internet, but when there is any update, Violentmonkey can check for me and ask me should update it or not.

Screen Shot 2023-09-22 at 10 31 34

I think this behavior shall be still "allow update"? How to set it properly to make that allow me editing while notify me for the update where it is auto checked.

^^^^ this should be the default behavior as the same as Tampermonkey.

## Remarks

_Temporary Solution: https://github.com/violentmonkey/violentmonkey/issues/1901#issuecomment-1732253676_

tophf:
Fixes #1901.

Stage | pic
-|-
Editor right after opening is gray, there's a red hint, and no buttons to save | image
Editor after trying to type or clicking the red label (it's clickable only for non-removed scripts) | image
When enabled the option becomes red and the note stays on screen until the script is saved | image
The option in the settings tab is also red when enabled | image

* I've extracted the update option into a separate component, hopefully correctly - it seems to work correctly at least.
* To make everything reactive I got rid of `settings` and just used a copy of the original script.
* Added `<keep-alive>` for the tabs in the editor and switched to `v-if`

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -807,7 +807,7 @@ readonly:
   message: Read-only
 readonlyNote:
   description: Warning when trying to type in the editor.
-  message: Auto-updated script is read-only until you disable updates or allow editing.
+  message: Auto-updated script is read-only unless you disable updates or allow editing.
 readonlyOpt:
   description: Option label in the editor, displayed along with readonlyOptWarn.
   message: Allow edits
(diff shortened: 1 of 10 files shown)


tophf:
I want to release this in a new beta to see the reception, so we'll deal with possible conceptual problems in my code later.

PR #1936, opened 2023-10-16
perf: improve `tldts.getDomain` speed

SukkaW:
Address suggestions from @remusao (https://github.com/violentmonkey/violentmonkey/pull/1883#discussion_r1359305249).

Update `getDomainSharedOpts` for `getDomain` as inputs are known to be valid hostnames.

The `getPublicSuffix` still uses the old options, as `getPublicSuffix` is only used in unit test cases and inputs can be invalid.

### DIFF
--- a/src/common/tld.js
+++ b/src/common/tld.js
@@ -4,7 +4,19 @@ import { getDomain as getDomain_, getPublicSuffix as getPublicSuffix_ } from 'tl
  * tldts does not respect the public suffix list by default, but can be opt in manually
  * with the option `allowPrivateDomains`. Hoist the `sharedOpts` can also help avoid
  * re-creating the object every time.
+ *
+ * Note `extractHostname` and `validateHostname` are set to false because the inputs are
+ * from `new URL(url).hostname` and are known to be valid
  */
-const sharedOpts = { allowPrivateDomains: true };
-export const getDomain = url => getDomain_(url, sharedOpts);
-export const getPublicSuffix = url => getPublicSuffix_(url, sharedOpts);
+const getDomainSharedOpts = {
+  allowPrivateDomains: true,
+  extractHostname: false, // inputs are already hostnames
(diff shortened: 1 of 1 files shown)


src/common/tld.js
```diff
@@ -4,7 +4,19 @@ import { getDomain as getDomain_, getPublicSuffix as getPublicSuffix_ } from 'tl
...
+  extractHostname: false, // inputs are already hostnames
+  validateHostname: false, // inputs are already valid, no need to perform extra validation
+}
+
+const getPublicSuffixSharedOpts = {
+  allowPrivateDomains: true
+};
+
+export const getDomain = url => getDomain_(url, getDomainSharedOpts);
+export const getPublicSuffix = url => getPublicSuffix_(url, getPublicSuffixSharedOpts);
```
tophf, line 22: AFAICT the same options can be used for both functions.

SukkaW: I am worrying about the `getPublicSuffix` case here:

image

image

The input of `getPublicSuffix` can be invalid IMHO.

tophf: AFAICT it can't be invalid because it's extracted as dot-separated valid sequence via `((?:\.[-\w]+)+)/` or `(|(?:\.[-\w]+)+)`

SukkaW: @tophf Exactly! Using regex can not ensure whether the input is a valid domain or not, thus extra check from `tldts` is required!

Also, the option will affect how `tldts` handles the invalid TLD:

https://runkit.com/sukkaw/652d2ea9557abc00083e0209

image

tophf: No, my point is that our regex is guaranteed to extract a valid sequence.

tophf: Also note that optimizing getDomain is pointless performance-wise as we use it only when showing the popup and when the user creates a new script for the tab.

PR #1945, opened 2023-10-28
feat: add zombieUrl to GM_notification

### LINKED ISSUE #1943: [Feature]  Add `url` property for `GM_notification` to open a new tab on click and make it more persistant.
**Motivation**

```javascript
// ==UserScript==
// @name        New script - github.com
// @match       https://github.com/*
// @grant        GM_notification
// @author      x
// @description 2023/10/28 09:25:14
// ==/UserScript==
'use strict';
let API_URL = "https://example.org/api234";
const cache = {};
setInterval(async () => {
    const item = await fetch(API_URL);
    if (!(item.name in cache) || item.version > cache[item.name]) {
        cache[item.name] = item.version;
        // Hope --> GM_notification(item.name, {url: item.url});
       const ct = GM_notification(item.name, {onclick: () => {window.open(item.url); ct.remove();}}) // Actually
    }
}, 5000);
```
It just push an notification that refer to an URL, it shouldn't be destroyed after the tab close (if using `onclick` event, it will do so).
Using `zombieTimeout` cannot solve this problem, because the function onclick refers to has been destroyed.

**Proposed Solution**

Hope to suppport `url` property. Just like Tampermonkey 5.0.6189 does: https://www.tampermonkey.net/documentation.php#api:GM_notification

**Use Cases**

push news.

我不太会说英语，但是十分感谢开发者们！

tophf:
fixes #1943.

### DIFF
--- a/src/background/utils/notifications.js
+++ b/src/background/utils/notifications.js
@@ -1,13 +1,24 @@
-import { i18n, defaultImage, sendTabCmd, trueJoin } from '@/common';
-import { addPublicCommands } from './init';
+import { i18n, defaultImage, sendTabCmd, trueJoin, getFullUrl } from '@/common';
+import { addPublicCommands, commands } from './init';
 import { CHROME } from './ua';
 
+/** @type {{ [nid: string]: browser.runtime.MessageSender | function }} */
 const openers = {};
-const removeNotification = id => browser.notifications.clear(id);
+const kZombieTimeout = 'zombieTimeout';
+const kZombieUrl = 'zombieUrl';
 
 addPublicCommands({
   /** @return {Promise<string>} */
-  async Notification({ image, text, tag, title, silent, onclick, zombieTimeout }, src) {
+  async Notification({
+    image,
+    text,
+    tag,
+    title,
+    silent,
+    onclick,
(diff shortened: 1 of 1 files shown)


gera2ld:
Is there an easy way to open a URL on click no matter whether the script is dead?

tophf:
There's no such API.

gera2ld:
But the TM API seems to support that easily:

```js
GM_notification({
  url: OpenUrlNoMatterWhetherScriptIsDead,
});

GM_notification({
  url: OpenOnlyWhenScriptIsDead,
  onclick: e => e.preventDefault(),
});

GM_notification({
  onclick: () => GM_openInTab(OpenUrlOnlyWhenScriptIsNotDead),
});
```

CMIIW

tophf:
As I stated in the issue, I don't like how Tamperonkey implements it:
* it doesn't play well with an already existing solution we provide via zombieTimeout
* using a DOM method preventDefault is super weird by itself and not obvious 
* the name `url` is too generic, it's not obvious what it does

gera2ld:
Reasonable :+1:

tophf:
I wondered if it can generalized as `zombieAction`, but didn't come up with anything other than opening of a URL...

Issue #1663, opened 2022-11-24
[Feature] 文字说明挡住最后面一个脚本...


Noah1995:
![image](https://user-images.githubusercontent.com/25548230/203758081-6f9f4896-696c-420a-94c6-388291d4f6a0.png)
文字说明挡住最后面一个脚本...
有时候开关都点不了...
请不要重叠...

tophf:
I can't reproduce this problem. It's also weird that you have two scrollbars.

To help us investigate please do the following:
1. right-click `2.13.3`
2. click "Inspect"
3. right-click `<html>`
4. click "Copy"
5. click "Copy outerHTML"

![image](https://user-images.githubusercontent.com/1310400/203772205-52a4de60-204a-4077-9cdc-7583f4684503.png)

Noah1995:
Many scripts appear double rolling bar
[Copy outerHTML.txt](https://github.com/violentmonkey/violentmonkey/files/10088142/Copy.outerHTML.txt)

Noah1995:
![录制_2022_11_25_10_32_09_453](https://user-images.githubusercontent.com/25548230/203888700-7a0adf2e-03c0-439f-9aae-88eeb2cddf51.gif)

tophf:
I still can't reproduce these problem even when using your html.

What is your browser?
Did you change the scale setting in OS to 125% or more?

Noah1995:
Can you remove the text prompt...
![录制_2022_11_26_00_17_53_923](https://user-images.githubusercontent.com/25548230/204024627-0ac7d259-58b6-4664-bf1f-85a3ad278188.gif)

tophf:
We can't remove something unless we can understand what causes the problem.
You can remove it yourself by adding the following code to "Custom CSS" in Violentmonkey settings:

```css
.page-popup .message {
  display: none;
}
```

tophf:
> My question is not about double scrollbars...

It's the source of the problem.

Noah1995:
It is indeed related to the scale setting
100% no problem
![image](https://user-images.githubusercontent.com/25548230/204079914-c854c447-af14-49e3-b7ac-0b48aaedf116.png)

tophf:
I can reproduce it now if I reduce the screen resolution and increase OS zoom, so that the popup becomes less than 600px.

PR #1952, opened 2023-11-11
use window.innerHeight on desktop browsers with zoom gt than 100%

magicdawn:
screenshot on #1663 

image

when chrome global zoom set gt than 100%
in chrome://settings/appearance page

e.g when `110%`

```
> document.documentElement.clientHeight
545
> window.innerHeight
545
> document.body.clientHeight
600
```

which turns `html` element to a scroll container, the outer scrollbar shows.

for desktop chrome, `$0.style.maxHeight = window.innerHeight + 'px'` did remove outer scrollbar.

----

more info on chrome global zoom influence extension's popup

- https://github.com/bitwarden/clients/issues/1008
- https://bugs.chromium.org/p/chromium/issues/detail?id=69265

### DIFF
--- a/src/popup/views/app.vue
+++ b/src/popup/views/app.vue
@@ -556,9 +556,17 @@ export default {
   mounted() {
     focusMe(this.$el);
     keyboardService.enable();
+
+    // from https://stackoverflow.com/a/11381730
+    function mobileCheck() {
+      let check = false;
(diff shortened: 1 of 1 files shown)


src/popup/views/app.vue
```diff
@@ -556,9 +556,17 @@ export default {
   mounted() {
     focusMe(this.$el);
     keyboardService.enable();
+
```
tophf, line 559: remove this

src/popup/views/app.vue
```diff
@@ -556,9 +556,17 @@ export default {
...
+      return check;
+    }; 
     this.$el.style.maxHeight = Math.min(
       // innerHeight may exceed 600px in a mobile browser which displays the popup as a fullscreen page
-      Math.max(600, innerHeight),
+      // on desktop browsers a global zoom > 100% cause multiple scrollbars
+      mobileCheck() ? Math.max(600, innerHeight) : Math.min(600, innerHeight),
```
tophf, line 569: Find a different way to implement mobileCheck.

tophf: Or maybe simplify the check to `android`...

tophf:
I found a more reliable method.

magicdawn:
Thanks for resolving popup scrollbars with zoom set. It's quite common for large screens.

Issue #1939, opened 2023-10-20
[BUG] Not yet supported in Orion (macOS)


cyfung1031:
Orion is a newly developed browser in macOS using Webkit.

It can support both Chrome and Firefox extensions.
<img width="206" alt="Screen Shot 2023-10-20 at 14 34 13" src="https://github.com/violentmonkey/violentmonkey/assets/44498510/53c54401-fb6d-4ec9-81bc-72493b6613c7">

However not every extension is currently supported due to its support on only 70% WebExtensions APIs.
https://help.kagi.com/orion/browser-extensions/macos-extensions.html
<img width="556" alt="Screen Shot 2023-10-20 at 14 36 30" src="https://github.com/violentmonkey/violentmonkey/assets/44498510/fead7b0b-30cf-4727-837c-dd59a63e41c8">

So it would be great that Violentmonkey can be one of the supported extensions in Orion.

<img width="277" alt="Screen Shot 2023-10-20 at 14 39 05" src="https://github.com/violentmonkey/violentmonkey/assets/44498510/849059bf-61da-498a-a798-1a4d800337d3">

Right now the problem is, the script can be installed in Violentmonkey, but never match the page and no execution of userscripts.
There is no message in violentmonkey's background console.

System: MacBook Pro (macOS Monterey 12.6.1 build 21G217)
Browser: Orion Version 0.99.125-beta (WebKit 616.1.22) Build date Jul 31 2023
Violentmonkey: Stable 2.15.0 and Beta 2.15.9

tophf:
I don't have a Mac device so there's nothing I can do. Someone will have to debug it using [remote devtools](https://www.google.com/search?q=chrome+remote+devtools+iphone).

cyfung1031:
@tophf 

As per their community discussion [here](https://orionfeedback.org/d/547-tampermonkey-violentmonkey-do-not-work-properly/24), **Violentmonkey 2.13.0.8** is mentioned to be the last working version for Orion.

I have verified by myself.

```js
// ==UserScript==
// @name        New script 
// @namespace   Violentmonkey Scripts
// @match       *://*/*
// @grant       none
// @version     1.0
// @author      -
// @description 11/30/2023, 8:07:07 PM
// ==/UserScript==

console.log("VM123",GM.info)
```

Created this script in both VM 2.16.0 and VM 2.13.0.8.
Only the log from VM 2.13.0.8 is shown.

[...]

PR #1971, opened 2023-12-02
`browserAction.setIcon` by `path` only for Orion

cyfung1031:
see #1939

### DIFF
--- a/src/common/safe-globals.js
+++ b/src/common/safe-globals.js
@@ -23,6 +23,7 @@ export const hasOwnProperty = safeApply.call.bind(({}).hasOwnProperty);
 export const safeCall = Object.call.bind(Object.call);
 export const IS_APPLIED = 'isApplied';
 export const IS_FIREFOX = 'geckoProfiler' in chrome;
+export const IS_ORION = typeof webkitIndexedDB === 'object' && chrome.contentScripts && !navigator.userAgentData;
 export const ROUTE_SCRIPTS = '#' + SCRIPTS;
 export const extensionRoot = chrome.runtime.getURL('/');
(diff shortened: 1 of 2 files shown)


tophf (changes requested):
As the comment for `iconDataCache` says caching is necessary.

tophf:
Here's the comment:

> Caching own icon to improve dashboard loading speed, as well as browserAction API
> (e.g. Chrome wastes 40ms in our extension's process to read 4 icons for every tab).

cyfung1031:
> Here's the comment:
> 
> > Caching own icon to improve dashboard loading speed, as well as browserAction API
> > (e.g. Chrome wastes 40ms in our extension's process to read 4 icons for every tab).

OK. Can I make a boolean like "isOrion" then bypass the imageData and just pass the filepath for Orion?

tophf:
Ideally, the bug should be reported to Orion makers so if they fix it there'll be no need for kludges.

src/common/safe-globals.js
```diff
@@ -23,6 +23,7 @@ export const hasOwnProperty = safeApply.call.bind(({}).hasOwnProperty);
 export const safeCall = Object.call.bind(Object.call);
 export const IS_APPLIED = 'isApplied';
 export const IS_FIREFOX = 'geckoProfiler' in chrome;
+export const IS_ORION = typeof webkitIndexedDB === 'object' && chrome.contentScripts && !navigator.userAgentData;
```
tophf, line 26: This may change suddenly in th next version of Orion...

tophf:
So they say they'll fix this bug, which means there's no need for IS_ORION, assuming people use the latest version, which should be the case normally.

PR #1976, opened 2023-12-04
Add codespell: config, workflow + make it fix "preferred" typo

yarikoptic:
workflow would only detect new typos, not fix them

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,5 @@
+[codespell]
+skip = .git,*.pdf,*.svg,*.lock,*.css,.codespellrc
+check-hidden = true
+# ignore-regex = 
+# ignore-words-list =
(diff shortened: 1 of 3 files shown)


tophf:
It should be a part of the test if it's not too huge, definitely not an action, as there's an elevated probability of it stabbing us in the back rather regularly for English isn't our native language.

yarikoptic:
sorry, I don't see you using pre-commit (e.g. like we do [here](https://github.com/dandi/dandi-cli/blob/master/.pre-commit-config.yaml#L21) and otherwise not familiar with the test setup -- could you provide more guidance?

Frankly speaking, given that you had only 1 typo (kudos!), I would not too worry for stabbing you in the back regularly ;)

tophf:
Test is executed in pre-push hook and manually via https://github.com/violentmonkey/violentmonkey/blob/master/package.json#L17

Not sure we need codespell then if we had just one typo... 

@gera2ld?

gera2ld:
An action might be a little too heavy? Can we make it part of `pre-push`, or `pre-commit` with `lint-staged`?

tophf:
> How about making it part of the CI workflow

I really really really don't like not being sure that the thing I push is good or may be rejected by a test.

Issue #1586, opened 2022-09-24
Newer zip.js needs babel transforms for old browsers


tophf:
When I reported that the new zip.js uses globalThis which is unsupported in Chrome 55-70, the author fixed it but said it's up to us to maintain compatibility with the old browsers.

There are more problems like destructuring in objects (Chrome 60+) and async generators (Chrome 63+). I guess we'll have to babelize it ourselves, which was something that a reviewer on AMO disallowed, but it might be a personal mistake because they don't object to our custom build of codemirror.js.

Currently, as a temporary workaround, I've pinned an old zip.js in package.json.

cyfung1031:
change to jszip?

https://github.com/Stuk/jszip - 9.1k Star
https://github.com/gildas-lormeau/zip.js - 3k Star

tophf:
It doesn't support workers, which means it'll be slow on a device like smartphone. Anyway, this problem isn't critical for now as the old version works just fine.

cyfung1031:
@tophf  please be noted that jszip can zip more than 65,535 files but fflate cannot.

[jszip-limit-test.html](https://github.com/user-attachments/files/28417783/jszip-limit-test.html)
[fflate-zipsync-limit-test.html](https://github.com/user-attachments/files/28417802/fflate-zipsync-limit-test.html)

See https://github.com/101arrowz/fflate/issues/272

tophf:
Good to know that fflate has sane limits.

Rikk:
Hi @tophf. I think this change of library has caused a loss of timestamp precision (subseconds) in exported files inside the zip:

<img width="637" height="453" alt="Image" src="https://github.com/user-attachments/assets/8d79b469-39d8-44fd-9830-e2494934c32e" />

Also I see a very very small reduction in compression of each file (only a few bytes), insignificant.

cyfung1031:
I have checked the current JavaScript libraries for zip creation... none of them satisfied me.
So I decided to make a new one.
It is almost finish... just to update some doc and benchmark...
I will soon publish it to npm. (size & speed outperform both jszip and fflate)

Yes. `fflate` does not store the extended timestamp. this information is per file entry so the total zip file size will be increased.
both `uzip.js` and `fflate` did not take it into account.

the DOS timestamp just store local time and always even second.

---

<img width="1138" height="498" alt="Image" src="https://github.com/user-attachments/assets/cceb2ad6-b4d5-4a16-997a-d7b21f8e8be9" />

[...]

PR #1978, opened 2023-12-08
chore: update dependencies

cyfung1031:


tophf:
We can't update zipjs until #1586 is fixed and overall, updating dependencies is dangerous as everything needs to be tested carefully, so I do it only in case the update is bringing a substantial benefit. I don't see such benefit in this update.

PR #1996, opened 2024-01-02
feat: add "Reset settings" and "Remove all scripts"

### LINKED ISSUE #1995: [Feature] Request to add two new features
**Motivation**
 I have 25 scripts installed but there is no option to delete all scripts at once.

**Proposed Solution**
 Add a function to delete all scripts, because currently you have to manually delete each script one by one.

**Use Cases**
 When I want to delete all the scripts I have, your extension does not provide a button/option to delete all scripts at once.

Hello, happy holidays.

I suggest adding two functions in your extension.

1. Add a button/option to make it possible to delete all custom scripts. Currently, it is extremely inconvenient to delete each personal script one by one, because it is currently impossible to delete all of the number of scripts easily and at once.

2. In your very nice extension, I terribly miss a "Reset" button/function. Let's say some user didn't back up their settings beforehand, but made changes to your extension, but can't figure out what to change to return the default settings?!

I really miss the option to "Reset" the default settings. There is currently no such option. My only option to return all settings to default is to uninstall your extension. To go to the site to reinstall your extension and that way I will get your extension with default settings. I'm surprised there isn't such an option at all...

___________________________

Well, I can't edit the title anymore, but:

3. Add option for manual user language change. When the user's nationality is such that your extension does not have a translation in that language, then your extension automatically selects a default language of "English".

Add an opportunity for us users to choose a language close to us from the available languages, but to be able to choose a language ourselves.

Please consider adding these buttons/features.

tophf:
* fixes #1995: add `Reset settings` and `Remove all scripts` buttons in VM settings page.
   * Both show a confirmation first
   * Resetting, like vacuuming, shows the number of changes in the button's text + json values in the tooltip
   * Removing uses Recycle bin to make it harder to shoot oneself in the leg by accident

* show `Undo` as a button next to `Import` button after importing instead of a post-import confirmation, which was both redundant and inconvenient as closing it would lose the button so the users had to open a new tab for Violentmonkey dashboard or settings if they wanted to test something and possibly undo it.

@gera2ld, WDYT? Also, editing+saving a locale doesn't run `copyI18n` in gulpfile anymore (even though it has gulp.watch for the paths) while running `gulp dev`, probably after updating gulp to v4.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -78,6 +78,9 @@ buttonReplaceAll:
 buttonReset:
   description: Button to reset to default values.
   message: Reset
+buttonResetSettings:
+  description: Button in settings page to reset all settings
+  message: Reset settings
 buttonRestore:
   description: Button to restore a removed script.
   message: Restore
@@ -306,6 +309,10 @@ labelAutoUpdate:
 labelBackup:
   description: Label of the import/export section in settings.
   message: Backup
+  touched: false
+labelBackupMaintenance:
+  description: Label of the import/export/maintenance section in settings.
+  message: Backup and maintenance
 labelBadge:
   description: Label for option to show number on badge.
   message: 'Display on badge: '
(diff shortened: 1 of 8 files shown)


gera2ld:
LGTM.
Let me check the gulp script.

PR #2015, opened 2024-01-22
feat: add tags and enhance searching

gera2ld:
Editing tags:
image
Display of tags:
image
Search:
image


tophf:
Awesome!

But the hint is too big to be shown by default and it is overlaid by the built-in autocomplete dropdown. I suggest adding an (i) icon that shows this info on hover and maybe also on click (via showMessage) for mobile users.

image

The input is too wide. The previous width is sufficient + the new `field-sizing` CSS or a polyfill for older browsers to autosize the field on typing.

The config icon after the input is not vertically centered:
image

The sort order selector is hidden again, but I revealed it intentionally because I think it's just as important as search.

The tooltip is bugged at the edge:
chrome_2024-01-22_18-43-52

`/regexp/` notation was removed, but this form is widely used in software e.g. in devtools filters, while `re:` prefix is unusual.

gera2ld:
> The input is too wide. The previous width is sufficient + the new field-sizing CSS or a polyfill for older browsers to autosize the field on typing.

I thought it could be wider to contain more combined conditions, but yeah the previous width seems okay. BTW the previous layout overlaps with the refresh button in small screens. I'll see if I can find a better way to define the width.

> `/regexp/` notation was removed, but this form is widely used in software e.g. in devtools filters, while `re:` prefix is unusual and it makes searching for regexp in code impossible or weird like `code:re:foo`

The reason I chose `code+re:foo` over `code:/foo/` is to avoid tedious escaping, e.g. `code+re:https://www.google.com/` vs `code:/https:\/\/www.google.com\//`. After a second thought, I guess we can search with plain text in such cases, so it should not be a problem.

tophf:
There's no need to escape `/` inside regexps currently, i.e. `/\.com/foo/bar//` works correctly and it should remain so.

gera2ld:
Then how to tell if `/foo/ /bar/` is one or two regular expressions?

tophf:
An embedded space is an unsupported case, same as in devtools, and the common workaround is `\s`.

tophf:
Just checked our current implementation and it allows spaces because it only supports regexp as the entire search input value.

gera2ld:
Updated UI.
Also added support for simple regexp without spaces: `/foobar/ code:/foobar/`, but still kept `code+re:"foo bar"` to support regexps with spaces.

src/options/utils/search.js
```diff
@@ -0,0 +1,90 @@
...
+      // Strip ':'
+      prefix = prefix.slice(0, -1);
+      if (prefix.endsWith('+re')) {
+        rules.push({
+          scope: prefix.slice(0, -3),
+          pattern: new RegExp(pattern, 'i'),
+          negative,
+        });
+      } else {
+        const reMatches = pattern.match(/^\/(.*?)\/(\w*)$/);
```
tophf, line 55: I think this should be the first thing to try, otherwise it won't match when the entire input is `/foo #bar/`

src/_locales/en/messages.yml
```diff
@@ -914,6 +920,21 @@ titleSearchHint:
...
+  message: |-
+    * <kbd>Enter</kbd> key adds the text to autocomplete history
+    * All conditions are case-insensitive
+    * Space separated conditions can be combined
+    * Search by metadata: <code>"Awesome Script" "Description"</code>
+    * Search by tags: <code>#tag1 #tag2</code>
+    * Search by script name: <code>name:"awesome name"</code>
+    * Search by script code: <code>code:"awesome code"</code>
+    * Negative search: <code>!#tag2 !name:"unwanted"</code>
+    * Regular expressions: <code>name:/awesome_\w+/ name+re:"awesome \w+ with space"</code>
```
tophf, line 937: I'd like to separate them visually and restore the old examples:
```html
<code>/\w+?/</code>, <code>/\w+?/gi</code>, <code>name:/\w+?/</code>, <code>name+re:"\w+? with space"</code>
```

tophf (approved):

src/options/utils/search.js
gera2ld: This is intended. I don't want to match regexp with spaces this way, as it is ambiguous and confusing.
So `/foo #bar/` will be parsed into string `/foo` and tag `bar/`. To input a regexp we can use `re:"foo #bar"` or `/foo\s#bar/`.

tophf: Strictly speaking the new behavior is correct, but it's not really useful because tags aren't likely to end in `/`...

gera2ld: Yeah, by right they should be two strings since `/` is not allowed in tags. But an invalid tag matches nothing so it does not hurt. I think it might be a bad idea for `#...` to be sometimes a tag and sometimes not.

PR #2020, opened 2024-01-30
Add batch action menu

### LINKED ISSUE #2006: [Feature] Added a feature/option to disable/enable all installed scripts
Guys, thanks for adding the feature I had requested re: https://github.com/violentmonkey/violentmonkey/issues/1995/

Guys I would like you to add one more feature.

It's about adding a `function/option` to `disable/enable` **all installed scripts**. Some extensions have this feature/option which can enable/disable extensions - this is an example of installed browser extensions.

Why am I suggesting the introduction of this feature? A while ago I had to disable all the scripts one by one because I can't disable all the scripts with one click. The goal was to find out which of the installed scripts gave me a defect.

Please consider adding a feature/option to disable/enable all installed scripts.

gera2ld:
Close #2006 

image

### DIFF
--- a/src/_locales/ar/messages.yml
+++ b/src/_locales/ar/messages.yml
@@ -236,6 +236,9 @@ headerRecycleBin:
 helpForLocalFile:
   description: Label of the checkbox.
   message: ''
+hintForBatchAction:
+  description: Hint shown next to the buttons for batch actions.
+  message: ''
 hintInputURL:
   description: Hint for a prompt box to input URL of a user script.
   message: ''
--- a/src/_locales/cs/messages.yml
+++ b/src/_locales/cs/messages.yml
@@ -240,6 +240,9 @@ headerRecycleBin:
 helpForLocalFile:
   description: Label of the checkbox.
   message: ''
+hintForBatchAction:
+  description: Hint shown next to the buttons for batch actions.
+  message: ''
 hintInputURL:
   description: Hint for a prompt box to input URL of a user script.
   message: 'Adresa skriptu:'
--- a/src/_locales/de/messages.yml
+++ b/src/_locales/de/messages.yml
@@ -272,6 +272,9 @@ headerRecycleBin:
 helpForLocalFile:
   description: Label of the checkbox.
   message: Anleitung für per Drag-and-Drop hinzugefügte lokale Skripte anzeigen
+hintForBatchAction:
+  description: Hint shown next to the buttons for batch actions.
+  message: ''
 hintInputURL:
   description: Hint for a prompt box to input URL of a user script.
   message: 'URL eingeben:'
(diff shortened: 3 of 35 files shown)


tophf:
It looks weird for some reason, probably because the line spacing is different or maybe because of the indent without bullets, I don't know... I wonder if icons would look better: 
off on del nuke

Maybe also add "update" and "export to zip"?

gera2ld:
Do you mean to display the icons on the top bar and apply to only filtered scripts?
My concern is that there are too many buttons that are rarely used.

tophf:
No, I meant showing them inside the dropdown. If some users want to always see the buttons they can use a custom CSS or maybe we'll add a "pin" icon in the menu.

What if instead the `>` icon it'd be `Batch` button, clicking or hovering which would show a horizontal div with the buttons?

gera2ld:
A horizontal div cannot clearly tell that the buttons are for filtered scripts only, which is why I preferred a dropdown or a separate window.
How about adding a toggle switching between "all" / "filtered only" and reusing the same icons, such as refresh, enable, disable, refresh, etc.?

tophf:
> A horizontal div cannot clearly tell that the buttons are for filtered scripts only

There's a solution: add `for 10 scripts` text in the same row.

> How about adding a toggle

Not sure it'll be useful (as it's easy to clear the filter), it might be actually confusing and cause mistakes if the user forgets to check/uncheck it, but maybe I'm too pessimistic. Anyway, it can be added to `for 10 scripts` text (and make the entire text bold or colored when it's enabled).

----

An `Undo` button might be very useful here.

gera2ld:
Yeah the extra text looks like a good idea. So we will always apply to filtered scripts only, making things easier.

I'll keep the refresh button for all, as it doesn't make sense to check updates for just a portion of scripts, and split the general buttons and batch action buttons into different groups, roughly like this:

image

tophf:
Some UI nits:
1. The underline below the actions block looks weird to me because I don't understand its meaning.
2. Maybe remove the `Remove all scripts` button from Settings?
3. The I-beam cursor over `for # matching scripts` may imply we can edit the field, particularly because of the underline, but we can't. A possible solution is to either add `cursor: default` or allow actually typing tags there.

tophf:
4. Set `store.importing` during the action the way it's done in vm-maintenance.vue to speed up processing by, like, 100x.

gera2ld:
> The underline below the actions block looks weird to me because I don't understand its meaning.

The underline indicates that the buttons above it belongs to one group and are all `for # matching scripts`. Otherwise it looks like only the last button applies to the matching scripts.

tophf:
The underline doesn't convey that purpose.

A full-height separator might work:
image

tophf:
After the action completes the button stays highlighted - why?

image

tophf:
I see it's because it's still the active element, but maybe it would make more sense conceptually to blur() it? If not, then the inter-button gap should be reduced probably by 1px to avoid an unusually wide border:

image

tophf (approved):

PR #2038, opened 2024-02-21
fix: set lastUpdated only if code changed

### LINKED ISSUE #2037: [Feature] When only change script's Settings, preserve script's timestamp
**Motivation**
When Export to zip, created archive contains scripts & settnig data("violentmonkey" file).
If I change only script's `Settings` section, created archive contains modified script & violentmonkey file.
If `Code` has not been changed, I would like you to change only violentmonkey file.

**Proposed Solution**
Only modify "violentmonkey" file.
Preserve script's timestamp.

**Use Cases**
Makes it easier to see which scripts have actually changed from the backup file.

tophf:
Fixes #2037.

I wonder if we should show both dates in the tooltip in the dashboard and add a sorting mode for lastModified? It might be convenient to find the recently customized scripts.

OTOH, this might be rare.

### DIFF
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -641,6 +641,7 @@ export async function parseScript(src) {
   if (!src.update) storage.mod.remove(getScriptUpdateUrl(script, { all: true }) || []);
   buildPathMap(script, src.url);
   const depsPromise = fetchResources(script, src);
+  // DANGER! Awaiting here when all props are set to avoid modifications made by a "concurrent" call
   const codeChanged = !oldScript || code !== await storage[S_CODE].getOne(id);
   if (codeChanged) props.lastUpdated = now;
   // Installer has all the deps, so we'll put them in storage first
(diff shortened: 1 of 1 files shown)


src/background/utils/db.js
```diff
@@ -608,7 +610,7 @@ export async function parseScript(src) {
     }
     delete script[INFERRED];
   }
-  props.lastModified = props.lastUpdated = src.lastModified || Date.now();
+  props.lastModified = now;
```
tophf, line 613: `src.lastModified` isn't set anywhere AFAICT, only `src.props.lastModified`

src/background/utils/db.js
```diff
@@ -639,11 +641,14 @@ export async function parseScript(src) {
   if (!src.update) storage.mod.remove(getScriptUpdateUrl(script, { all: true }) || []);
   buildPathMap(script, src.url);
   const depsPromise = fetchResources(script, src);
+  // DANGER! Awaiting here when all props are set to avoid modifications made by a "concurrent" call
```
gera2ld, line 644: Just curious, when will there be a concurrent call?

gera2ld: got it, thanks

PR #2040, opened 2024-02-26
fix: update tooltip alignment to avoid overflow

### LINKED ISSUE #2036: [BUG] can not scroll script settings page
### Sequence of actions: 
1.Create a new script
2.Click Settings
3.Input strings to `@include rules` or expand input area

### Problem: 
Overflowed elements (`@match rules`,`@exclude rules`,`@exclude-match rules`) are not visible and cannot be scrolled "Settings" page.

### Expected result:  
Overflowed elements are not visible, but they can be scrolled "Settings" page.

#### Environment:
- OS: Windows
- Browser: Firefox 115.8.0esr
- Violentmonkey Version: v2.18.0 (BETA v2.15.9 and later)

gera2ld:
Close #2036

### DIFF
--- a/src/common/ui/code.vue
+++ b/src/common/ui/code.vue
@@ -31,22 +31,22 @@
         <span v-text="i18n('labelReplace')"></span>
         <!-- id is required for the built-in autocomplete using entered values -->
         <input class="flex-1" type="search" id="editor-replace" v-model="search.replace">
-        <tooltip :content="tooltip.replace">
+        <tooltip :content="tooltip.replace" align="end">
           <button type="submit" v-text="i18n('buttonReplace')"></button>
         </tooltip>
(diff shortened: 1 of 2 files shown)


tophf:
It also happens on `<` `>` buttons.

tophf:
Actually align=end doesn't solve the problem when the button is wrapped to the left:
image

gera2ld:
> Actually align=end doesn't solve the problem when the button is wrapped to the left

Yeah but fixing that requires a refactor of the tooltip component, I will just fix the most common cases here first.

tophf:
Did you see my comment?

> It also happens on < > buttons.

gera2ld:
Oh I thought I fixed 2 more buttons, just realized they were the replace buttons. Let me update.

PR #2079, opened 2024-04-21
[WIP] Add support for nativeConnect

innovate-invent:
There are a number of issues/discussions/feature requests for various mechanisms that allow accessing host resources from ViolentMonkey or other User Script managers. 

This adds the `GM_nativeConnect` API that allows a user to configure access to an application on the host to communicate with that doesn't require running a separate web server. 

It forwards the [Native Messaging](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging) API to user scripts with restrictions.

Due to how potentially dangerous this feature is I have added a global setting to allow access to GM_nativeConnect that is disabled by default. With that I have also extended the `@grant` meta data to require listing native application ids with the grant.

```
@grant GM_nativeConnect myapp1 node
```

Additionally this requires the user to configure apps with the browser by placing manifest files in the [correct locations](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_manifests#manifest_location) on their system.

```json
{
  "name": "node",
  "description": "NodeJS",
  "path": "/usr/bin/node",
  "type": "stdio",
  "allowed_extensions": ["{aecec67f-0d10-4fa7-b7c7-609a2db280cf}"]
}
```

An example script that allows reading a file from the system:

```js
// ==UserScript==
// @name        New script
// @match       *://example.org/*
// @grant       GM_nativeConnect node
// @version     1.0
// @author      -
// @description 20/04/2024, 14:33:18
// ==/UserScript==

const port = GM_nativeConnect("node");
port.onMessage.addListener(data=>console.log(data));
port.postMessage(`
const fs = require('fs');
console.log(fs.readFileSync('/path/to/input.txt',{ encoding: 'utf8', flag: 'r' }));
`);
```

NodeJS is just one example, any application that communicates via stdin/stdout can be used as a app.

### DIFF
--- a/src/common/options-defaults.js
+++ b/src/common/options-defaults.js
@@ -92,4 +92,5 @@ export default {
   valueEditor: defaultsValueEditor,
   /** @type {'' | 'dark' | 'light'} */
   uiTheme: '',
+  allowNativeMessaging: false,
 };
--- a/src/injected/content/gm-api-content.js
+++ b/src/injected/content/gm-api-content.js
@@ -8,6 +8,7 @@ const { toLowerCase } = '';
 const { [IDS]: ids } = bridge;
 let setPopupThrottle;
 let isPopupShown;
+const ports = new Map();
 
 addBackgroundHandlers({
   async PopupShown(state) {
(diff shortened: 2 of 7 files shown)


innovate-invent:
I need some help wiring in the global configuration into https://github.com/violentmonkey/violentmonkey/pull/2079/files#diff-92b27ade356507a03a6ffe0bb9bd5dcc698e1fde07de83e1e1b3df9b1dac0ebcR58

tophf:
I don't think we want to expose such capabilities to userscripts.
@gera2ld?

gera2ld:
I think Chrome has made a lot of effort to limit how extensions interact with the host. So I think we should avoid acting as a proxy to bypass those limits. Perhaps a separate web extension would be more appropriate for this purpose?

innovate-invent:
While nodejs is an extreme example, how many people are running webservers locally to augment their user script functionality? It would be more secure to communicate with the external logic via the Native Messaging API rather than having a server open to all scripts/local clients.

I can switch to rigging up a local server if you feel strongly about excluding this.

tophf:
> I think Chrome has made a lot of effort to limit how extensions interact with the host.

Chrome's limits don't apply to nativeMessaging, so it's not a concern.

> While nodejs is an extreme example, how many people are running webservers locally to augment their user script functionality?

0.1% of all userscript users? I don't think this is worth complicating our extension.

PR #2089, opened 2024-05-01
Fix editor style by using largest possible viewport

### LINKED ISSUE #2088: [BUG] Cannot edit script values in some mobile device due to UI layout issue
On Samsung Galaxy Tab S6 with Firefox for Android, when I try to edit the values (those managed by GM_get/setValue) of a script in the Script > Values tab
- I cannot save the values because the buttons are missing (both in landscape and portrait). See screenshots below.
- for scripts that do not have any values, the buttons do show up.
- The issue does not happen on another device (Google Pixel 3a).
- I cannot reproduce it using desktop Firefox in responsive design mode, simulating the tablet.

Screenshots for:
- [portrait](https://github.com/violentmonkey/violentmonkey/assets/250644/0b9891bf-653f-4fd3-a60b-f53332ac113f)
- [landscape](https://github.com/violentmonkey/violentmonkey/assets/250644/f014bb3d-ae48-4126-b957-ea6ad3c0a60f)

### Expected result:  

The Ok / Apply / Cancel buttons should show up.

#### Devtools console contents:

Did not obtain.

#### Environment:
- OS: Android 12
- Browser: Firefox for Android 125.3
- Device: Samsung Galaxy Tab S6 (1600 X 2560 pixels) 
  - On the browser: `devicePixelRatio`: 2.22, `screen.height`: 720,  `screen.width`: 1152
- Violentmonkey Version: 2.18.0

gera2ld:
Fix #2088

The current implementation is broken on a large screen because `.editor-open .page-options` is only hidden on a small screen while `position: static` is always set for a touch screen.

I try to use the same behavior for both desktop and mobile in this PR for easier debugging.

### DIFF
--- a/src/options/views/edit/index.vue
+++ b/src/options/views/edit/index.vue
@@ -448,7 +448,6 @@ function setupSavePosition({ id: curWndId, tabs }) {
   }
   &-body {
     padding: .5rem 1rem;
-    // overflow: auto;
     background: var(--bg);
     flex: 1;
   }
@@ -516,10 +515,13 @@ function setupSavePosition({ id: curWndId, tabs }) {
   }
 }
 
-.touch .edit {
-  // fixed/absolute doesn't work well with scroll in Firefox Android
-  position: static;
-  // larger than 100vh to force overflow so that the toolbar can be hidden in Firefox Android
+.touch body {
+  position: relative;
+  /*
+   * Set height to 1px larger than screen height to force overflow so that the toolbar can be hidden in Firefox Android.
+   * Use `100vh` (largest possible viewport) to avoid flashing caused by URL bar resizing.
+   * See https://developer.chrome.com/blog/url-bar-resizing
+   */
   min-height: calc(100vh + 1px);
 }
 
--- a/src/options/style.css
+++ b/src/options/style.css
@@ -181,12 +181,4 @@ textarea:not([rows="1"]):not(.h100) {
       white-space: normal;
     }
   }
-
-  .edit-open {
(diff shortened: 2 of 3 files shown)


tophf:
LGTM, but `inset` requires Chrome 87 and Firefox 66, so we should either transpile our CSS or restore the individual props.

gera2ld:
Removed `inset`.

src/common/ui/style/style.css
```diff
@@ -357,6 +345,18 @@ li {
   right: 0;
   bottom: 0;
 }
+.h-screen {
```
tophf, line 348: Looks like these 4 rules were moved by accident?

gera2ld: I was trying to override `height` of `abs-full` but seems no longer needed, now reverted.

src/options/views/edit/index.vue
```diff
@@ -516,10 +515,13 @@ function setupSavePosition({ id: curWndId, tabs }) {
   }
 }
 
-.touch .edit {
-  // fixed/absolute doesn't work well with scroll in Firefox Android
-  position: static;
-  // larger than 100vh to force overflow so that the toolbar can be hidden in Firefox Android
+.touch body {
```
tophf, line 518: Just to confirm, so now this applies to all pages in options?

gera2ld: Yes. The issue is that if the page cannot scroll, the toolbar of Firefox on Android will not hide automatically and will cover part of the viewport. This happens to all pages in options. This hack should fix all full screen pages or modals.

PR #2093, opened 2024-05-05
fix: set header bar to scrollable for small screen only

gera2ld:
follow-up to #2089

### DIFF
--- a/src/common/ui/style/style.css
+++ b/src/common/ui/style/style.css
@@ -512,7 +512,7 @@ input[type].has-error {
   }
 }
 
-@media (max-width: 500px) { /* This value is also used in other files */
+@media (max-width: 550px) { /* This value is also used in other files */
   .hidden-sm {
     display: none !important;
   }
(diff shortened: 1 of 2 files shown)


tophf:
I don't think it helps because mobile devices may be wider than 550px:

image

gera2ld:
When it's wider than 550px, it's still the same as desktop.
When it's smaller than 550px, the header bar will be scrollable, and the dropdown is fixed at `left: 0` same as before.

tophf:
The emulator's width is set to 800px and devicePixelRatio is 1.5, so the ffective width is 533, which is why `@media` fails?

tophf:
Maybe it'd be more reliable to teleport the menu and use position:absolute?

gera2ld:
> Maybe it'd be more reliable to teleport the menu and use position:absolute?

Yes, but that requires a rewrite of component and some more details to deal with (position sync, page overflow).

gera2ld:
Maybe I should check if `header` is scrollable by js instead of using a hard-coded breakpoint?

tophf:
Maybe `flex-wrap: wrap` in the header [for small widths]?

tophf:
>I cannot reproduce what you got above.

Here it is in a desktop version after shrinking the window:

image

gera2ld:
My bad, wrong class name.

```diff
-    .vl-dropdown-right .vl-dropdown-menu {
+    .vl-dropdown-menu {
```

tophf:
LGTM.

PR #2106, opened 2024-05-26
fix: improve auto-size in popup

tophf:
Now this should work with all types of zoom:
1. OS zoom
2. Browser settings -> page zoom
3. Ctrl-wheel zoom in Violentmonkey dashboard (either below or above 100%) then open the popup for a web tab

@gera2ld, please test it on a real mobile, I only tested in the emulator. It should fit the entire height of the device as it did previously, I just want to confirm.

### DIFF
--- a/src/popup/views/app.vue
+++ b/src/popup/views/app.vue
@@ -541,20 +541,16 @@ function showButtons(item) {
 
 onMounted(async () => {
   const $el = $root.value;
-  const $elStyle = $el.style;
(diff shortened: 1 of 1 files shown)


tophf:
Unfortunately, it's glitchy in Chrome: resizing is visible and sometimes leaves an empty disabled scrollbar if I open the popup many times...

tophf:
667e88c5 seems to fix it.

tophf:
No, it may still flash an oversized popup...

PR #2119, opened 2024-06-15
feat: create global option that allows XHR requests to set cookies

### LINKED ISSUE #2118: [Feature] Setting to allow XHR responses to set cookies
**Motivation**
It seems that [this release](https://github.com/violentmonkey/violentmonkey/releases/tag/v2.15.0) (this commit: https://github.com/violentmonkey/violentmonkey/commit/712df2b67a31d66ca0d93b65549045a7c3ebd535) last year was made to improve parity with Tampermonkey, which changed how it allows XHR responses to set cookies. I have been using an old TM version until recently, when I upgraded from TM 5.1.1 to 5.2.0, which included the change in behavior I am referring to (TM 5.1.1 did not filter out `set-cookie` headers). In response, I started looking for other userscript managers to determine if that behavior change is isolated to TM, but it seems VM also adopted the same behavior change.

This is breaking many of the userscripts used within the company I work for. All of our internal sites perform SSO authentication in the following way:

1. Go to target URL
2. Service checks to see if there are SSO cookies present. If so, validates them and then determines authZ. If authorized, process request.
3. If the SSO cookies are not present, or if they are expired, then the target site responds with a `set-cookie` header, containing an SSO forgery protection token. It then redirects the user to our internal auth service. 
4. The internal auth service authenticates the user (via credentials or session cookies), generates a token for the initial target URL, and then redirects the user back to the initial target URL with the token added to the request.
5. The initial site takes the token from the auth request and stores it in a cookie, and redirects it to the original URL, without the token in the request.
6. Return to step 2 (since the SSO tokens are now present and valid, the request is then processed)

By filtering out `set-cookie` headers, we just get put on an infinite loop, since no SSO cookies would have been set. You can temporarily alleviate this by going to the target site manually, but our SSO tokens don't have a long liveness (5-10 mins) before they need to be refreshed, so it won't last very long. Additionally, since we have hundreds of internal sites which each store their own SSO tokens, the user would have to visit every one, every 5-10 minutes.

**Proposed Solution**

Short-term: some global setting that allows cookies to be set by background XHR response headers. That should unblock the people in my company.

Long-term: not sure...

There's a possibility I am just missing some setting, and I apologize if that is the case!

The change to remove the `set-cookie` header was a fairly big backwards-incompatible change, though I do understand the motive. Are there any processes in place for the various userscript manager teams to come to consensus on these types of behaviors?

**Use Cases**
Covered in the motivation section above

grablair:
Fixes #2118

Details of motivation and use case can be found in the issue above. Happy to discuss this!

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -324,6 +324,9 @@ labelAbout:
 labelAdvanced:
   description: Label for button to show advanced settings.
   message: Advanced
+labelAllowGmXhrResponseCookies:
+  description: Option to allow GM.xhr responses to set browser cookies
+  message: 'Allow GM.xhr responses to set browser cookies'
 labelAllowUpdate:
   description: Option to allow checking updates for a script.
   message: Allow update
--- a/src/background/utils/requests-core.js
+++ b/src/background/utils/requests-core.js
@@ -1,5 +1,6 @@
 import { buffer2string, getUniqId, isEmpty, noop } from '@/common';
 import { forEachEntry } from '@/common/object';
+import { getOption } from './options';
 import { CHROME } from './ua';
 
 let encoder;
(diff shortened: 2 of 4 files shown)


tophf:
It should be an option in GM_xmlhttpRequest, not in the extension, otherwise it'll break scripts that don't expect it. And I'm not convinced we should add it. Try opening an issue in Tampermonkey, its author likes to add features, so if we like his solution we can implement it here as well.

tophf:
Already fixed in a617df2b.

PR #2146, opened 2024-07-20
Add GM_messageExtension allowing userscript to trigger other WebExtension

GHolk:
This API allow userscript to communicate with other extensions
if they do listen on the (browser.runtime.on)MessageExternal event.

The script should also declare the extension id it want to communicate with
in the `@connect` metadata block.

Different from [Tampermonkey's connect field][tmc],
the ajax will not restricted to the URL in the connect field.
These commits only check the connect field with `web-extension://` prefix
and restrict the GM_webextSendMessage's destination id.

[tmc]: https://www.tampermonkey.net/documentation.php#meta:connect

I am not very familiar with this project,
any advice or improvement is wellcome!

## future work
 1. Maybe we should prevent userscript use this API to send message to violent monkey itself?
   Even if they declare `web-extension://*` or violentmonkey's id.
 2. include the userscript's information (namespace and name?) in the payload?
     so the other will not only see the sender is violent monkey.

### DIFF
--- a/src/injected/content/index.js
+++ b/src/injected/content/index.js
@@ -4,6 +4,7 @@ import { injectPageSandbox, injectScripts } from './inject';
 import './notifications';
 import './requests';
 import './tabs';
+import './webext-send-message';
 import { sendCmd } from './util';
 import { isEmpty } from '../util';
 import { Run, finish } from './cmd-run';
--- a/src/injected/content/index.js
+++ b/src/injected/content/index.js
@@ -4,7 +4,7 @@ import { injectPageSandbox, injectScripts } from './inject';
 import './notifications';
 import './requests';
 import './tabs';
-import './webext-send-message';
+import './message-extension';
 import { sendCmd } from './util';
 import { isEmpty } from '../util';
 import { Run, finish } from './cmd-run';
--- a/src/injected/content/message-extension.js
+++ b/src/injected/content/message-extension.js
@@ -2,7 +2,7 @@ import bridge, { addHandlers } from './bridge';
 import { browser } from '../util';
 
 addHandlers({
-  async WebextSendMessage(options, realm) {
+  async MessageExtension(options, realm) {
     let response;
     let ok = true;
     try {
--- a/src/injected/content/webext-send-message.js
+++ b/src/injected/content/webext-send-message.js
@@ -1,8 +1,20 @@
-import { addHandlers } from './bridge';
+import bridge, { addHandlers } from './bridge';
 import { browser } from '../util';
 
 addHandlers({
-  WebextSendMessage(options) {
-      browser.runtime.sendMessage(options.id, options.message, null);
+  async WebextSendMessage(options, realm) {
+    let response;
+    let ok = true;
+    try {
+      response = await browser.runtime.sendMessage(
+        options.extId, options.message, null
+      );
+    }
+    catch (error) {
+      ok = false;
+      response = error;
+    }
+    const data = { ok, response };
+    bridge.post('Callback', { id: options.cbId, data }, realm);
   },
 });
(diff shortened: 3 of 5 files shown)


tophf:
There's no need for GM_webextSendMessage: if another extension can listen to external messages (by declaring `externally_connectable` in its manifest.json), userscripts and web pages can use `chrome.runtime.sendMessage()` directly.

GHolk:
@tophf No it can't, browser or chrome both are undefined.
Userscript can not access web-extension API like content-script.

I am not meaning communicate from webpage, which require specifying url in `externally_connectable.matches[]`
(which does not support in firefox now).
I want to communicate with the web-extension level api.

tophf:
I see. We don't plan to expose WebExtensions API to userscripts. There are other extension(s) that can do it, but I don't remember the name.

GHolk:
We already wrap and expose some, like notification and xhr to any domain.
Maybe we can wrap and add some restriction and expose it?

tophf:
Those are parts of the userscript API that existed for like 20 years. If you want to convince me or other userscript extension teams please provide some motivational use cases that demonstrate how it can be used.

GHolk:
For myself, I need this API because I want to trigger the extension [SingleFile] download the webpage easily.
Eg, I can put a `display:fixed` button on some domain with userscript, and click it to trigger the downloading.
Without this, I have to *open the browser menu / open extension list / (scroll and) click SingleFile* .

[SingleFile]: https://addons.mozilla.org/en/firefox/addon/single-file/

I write code, so I add this feature to make my life easy.
I already solve my problem, and I think maybe I can contribute back.
This is the opensource; I can use other's code, and I can give back.

In fact, I does not use this feature often.
It does great job in 1 week, and I have not used it for months.

I know tha add a new `GM_*` API is a serious thing,
(though TamperMonkey has a bunch of different API.)
and new features usually mean more burden for maintainers; this is not my personal project.
Anywan, thank you all for this great add-on.

tophf:
Sounds useful, but since it's the first time I see this feature being suggested, I'm not sure it should be added at all. Maybe you could open a feature request in Tampermonkey and [Firemonkey](https://github.com/erosman/support/issues/)? Make sure to describe the use case and the fact that this solution doesn't require `externally_connectable`.

I also don't like the `webext` part in the name. I'd probably prefer something `GM_messageExtension`. It would also make sense to add `GM_connectExtension` that uses chrome.runtime.connect + onConnectExternal.

@gera2ld, WDYT?

GHolk:
> Sounds useful, but since it's the first time I see this feature being suggested, I'm not sure it should be added at all. Maybe you could open a feature request in Tampermonkey and [Firemonkey](https://github.com/erosman/support/issues/)? Make sure to describe the use case and the fact that this solution doesn't require `externally_connectable`.

ok

> I also don't like the `webext` part in the name. I'd probably prefer something `GM_messageExtension`.

That's a better name than mine. I will use it before the PR get merged.

> It would also make sense to add `GM_connectExtension` that uses chrome.runtime.connect + onConnectExternal.

It may take large effort to wrap the connect API since it will be a long connection,
but I have not seen any extension listens to the onConnectExternal.
I will consider on it if anyone needs this.

tophf:
`@connect-extension id` seems simpler than `@connect web-extension://id`, moreover there's no technical reason we should reuse `@connect`.

douglasg14b:
I am also interested in this, the arguments against it don't appear to have solid footing.

A toolbox without niche, but essential, tools doesn't make for a great toolbox. Tools will be used when made available, similarly sending runtime messages is crazy useful for "hacking" your own behavior triggers into extensions via userscripts.

GHolk:
For anyone interested in this:
If you want to send message to other extensions on desktop, [Tridactyl] let you do this with the `:js` command.
It does not work on firefox for android so I forked violent monkey.

[Tridactyl]: https://addons.mozilla.org/zh-TW/firefox/addon/tridactyl-vim/

PR #2240, opened 2024-11-27
feat: HiDPI/Retina icons

p-mng:
This pull request adds HiDPI/Retina icon support for high resolution displays. The difference should be visible in the screenshots below.

before (no HiDPI support) | after (HiDPI support)
--- | ---
before | after

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -96,10 +96,8 @@ async function createIcons() {
     .toFile(`${dist}/icon16${type}.png`);
   };
   return Promise.all([
-    handle(48),
-    handle(128),
     ...types.map(handle16),
-    ...[32, 38].flatMap(size => types.map(t => handle(size, ...t))),
+    ...[32, 38, 48, 64, 96, 128].flatMap(size => types.map(t => handle(size, ...t))),
   ]);
 }
 
--- a/src/manifest.yml
+++ b/src/manifest.yml
@@ -6,7 +6,10 @@ author: Gerald
 homepage_url: 'https://violentmonkey.github.io/'
 icons:
   '16': public/images/icon16.png
+  '32': public/images/icon32.png
   '48': public/images/icon48.png
+  '64': public/images/icon64.png
+  '96': public/images/icon96.png
   '128': public/images/icon128.png
 default_locale: en
 browser_action:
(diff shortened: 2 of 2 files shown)


tophf:
Thanks. There's one thing though: browser_action doesn't use 96 and 128.

tophf:
Actually, this PR is only correct for one 96px icon shown in the list of extensions, but wrong for all the other icons, so I'll have to rework it.

p-mng:
Wrong in what sense?

tophf:
Our icon is dynamically assigned depending on the state and that code still uses the small icons.

tophf:
I'm testing right now in an emulator on 2x DPI and I see Firefox uses 48px for the browser_action icon, which is what needs to be added to our state routine.

p-mng:
Apologies for not testing that, I didn't how about the code changing the icons. Are you still working on this? The icons are still blurry on my Mac, but adding a 96px (2*48px) icon fixes that: <https://github.com/p-mng/violentmonkey/commit/26624937ac0890a8228ec22684f5dba4020f7bf5>

Current master:

before | after
--- | ---
before gray | after gray
before color | after color

tophf:
I don't have a Mac, but the size of the icon in your screenshot is 64px. Try using it instead of 96px and see if it stays crisp.

tophf:
I'd also like to know why did you add 64 to the desktop set of icons, not the mobile set, what browser and what part of its UI is shown on your screenshots?

p-mng:
> Try using it instead of 96px and see if it stays crisp.

Like this (<https://github.com/p-mng/violentmonkey/commit/680620bff316a03cb8e8650f131a1724457daeb2>)? Then the icon does indeed stay crisp.

> I'd also like to know why did you add 96 to the desktop set of icons, not the mobile set

I'm not using the addon on Firefox mobile, so I didn't want to make any changes to code I cannot test.

> what browser and what part of its UI is shown on your screenshots?

Firefox 132, [using the extensions button and panel](https://support.mozilla.org/en-US/kb/extensions-button).

tophf:
Ah, that makes sense, thanks. I didn't test the new UI in Firefox.

p-mng:
So I assume you thought the issue was specific to Android, meaning we can remove the mobile-specific sizes (unless there is another reason to keep it) and just add the 64px icon?

tophf:
I think I'll use this:
```js
const SIZES = !FIREFOX
  ? [16, 32]
  : ua.mobile
    ? [32, 38, 48] // 1x, 1.5x, 2x
    : [16, 32, 48, 64]; // 16+32: toolbar, 32+48+64: extensions panel
```

PR #2245, opened 2024-11-30
Update actions

larziwau:


tophf:
Please remove all changes in formatting and restore the comments, so the only changes that remain are the versions.

larziwau:
Well this formatting looks a lot better but its up to you.

larziwau:
Added comments back.

tophf:
"looks better" implies there's an absolute standard of beauty, which is not true. Please also remove changes like "- name: Checkout repository", we don't need these verbose names. There's also no need to change the letter case in the comments. The only changes that should remain are the versions after `@`.

tophf:
Also, it looks like we can't get rid of `checkout@v1` because "v1 keeps tags" as pointed out in the comments.

PR #2274, opened 2025-02-10
fix: improve unicode support of  globalVars of hint

tpsnt:
when mode set to{
    "name": "javascript",
    "globalVars": true
  }, unicode support(e.g. Chinese chars) of  globalVars of hint will be improved

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Get version and SHA
         run: node scripts/action-helper.js ci
       - name: Upload Artifact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         continue-on-error: true
         with:
           name: 'Violentmonkey-test-webext-${{ env.GIT_DESCRIBE }}'
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "violentmonkey",
   "description": "Violentmonkey",
-  "version": "2.29.0",
+  "version": "2.30.0",
   "scripts": {
     "prepare": "husky install",
     "dev": "gulp dev",
@@ -113,6 +113,5 @@
       "./test/mock/index.js"
     ],
     "testEnvironment": "./test/mock/env.js"
-  },
-  "beta": 1
+  }
 }
--- a/src/_locales/id/messages.yml
+++ b/src/_locales/id/messages.yml
@@ -133,13 +133,17 @@ confirmUndoImport:
   description: >-
     Tooltip of the Undo button after importing a zip in settings. The same text
     is also shown as a confirmation dialog after clicking this button.
-  message: ''
+  message: >-
+    Mengembalikan semua perubahan yang dibuat pada basis data (mengimpor,
+    memperbarui, mengedit, menyesuaikan)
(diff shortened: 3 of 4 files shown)


tophf:
I don't see any practical improvement for Violentmonkey. Can you elaborate?

tophf:
Ah, I see now. You want userOpts to override the internal mode... I got confused by the irrelevant `mode: [String, Object],`, please remove that.

tophf:
As for the main objective, I don't see this as a good solution, because most users won't specify `javascript-mixed` and thus would lose the proper syntax highlight.

tpsnt:
however, imho, by default, javascript-mixed won't be lost, isn't it?



---Original---
From: ***@***.***&gt;
Date: Mon, Feb 10, 2025 18:19 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [violentmonkey/violentmonkey] fix: improve unicode support of globalVars of hint (PR #2274)




 
As for the main objective, I don't see this as a good solution, because most users won't specify javascript-mixed and thus would lose the proper syntax highlight.

tophf:
I guess the correct fix would be to apply globalVars explicitly at the end:
```js
  const opts = {
    ...cmDefaults,
    ...userOpts,
    ...theme && { theme },
    ...internalOpts, // internal options passed via `props` have the highest priority
    mode: {
      name: props.mode || cmDefaults.mode,
      globalVars: userOpts.globalVars,
    },
  };
```

tpsnt:
ok, i will check that.



---Original---
From: ***@***.***&gt;
Date: Mon, Feb 10, 2025 18:28 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [violentmonkey/violentmonkey] fix: improve unicode support of globalVars of hint (PR #2274)




 
I guess the correct fix would be to apply globalVars explicitly at the end:
   const opts = {     ...cmDefaults,     ...userOpts,     ...theme &amp;&amp; { theme },     ...internalOpts, // internal options passed via `props` have the highest priority     mode: {       name: props.mode || cmDefaults.mode,       globalVars: userOpts.globalVars,     },   };

tpsnt:
image
turns out that it does not work. Perhaps javascript-mixed mode does not recognize globalVars like javascript mode does?
image

I will dig into javascript-mixed mode to find a way out.

tophf:
Actually, a much simpler solution might be to just enable `globalVars` by default in Violentmonkey. Do you know if there any drawbacks?

tpsnt:
i have no idea. Maybe i will look into some posts or threads of CodeMirror or javascript-mixed tomorrow.

tpsnt:
Don't see much drawbacks, maybe too many globalVars shown?
image

A dirty way would be like this:  
`const jsMode = CodeMirror.getMode(config, { name: 'javascript', globalVars: config.globalVars });`

tophf:
Looks fine. Also, normally people type something before invoking the completion.

cyfung1031:
@tophf It looks great. When will this PR could be merged?

tophf:
I've re-implemented @tpsnt's great work as 067769f1 to accommodate for our code style, please try the CI builds.

PR #2427, opened 2026-01-27
Fix GM.xmlHttpRequest not resolving

uohhh:


tophf:
@elchio, "resolving" is about Promise resolving, but in your script the problem is `@grant none` that disables the sandbox. The editor should explain it at the bottom:

>`@grant none` disables all other `@grant` keys.
>https://violentmonkey.github.io/api/metadata-block/

PR #2449, opened 2026-02-24
Update tab-installed.vue

D4n2021:
Added size info of to-be-deleted userscripts (shows before emptying bin and after having emptied it)

### DIFF
--- a/src/options/views/tab-installed.vue
+++ b/src/options/views/tab-installed.vue
@@ -526,9 +526,11 @@ async function getCodeFromStorage(ids) {
   onUpdate();
 }
 async function handleEmptyRecycleBin() {
-  if (await showConfirmation(i18n('buttonEmptyRecycleBin'))) {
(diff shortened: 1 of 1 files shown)


src/options/views/tab-installed.vue
```diff
@@ -526,9 +526,11 @@ async function getCodeFromStorage(ids) {
   onUpdate();
 }
 async function handleEmptyRecycleBin() {
-  if (await showConfirmation(i18n('buttonEmptyRecycleBin'))) {
+  const totalKB = (store.removedScripts.reduce((sum, s) => sum + (s.$cache.sizeNum || 0), 0) / 1024).toFixed(2);
```
tophf: It's better to show this total size in the title of the page to help the user decide if it's worth emptying the bin without clicking. Also it would avoid the need to translate a new confirmation message.

PR #2458, opened 2026-03-04
feat: Add button to publish to Greasy Fork

### LINKED ISSUE #2425: [Feature] Button to publish to GreasyFork/OpenUserJS
**Motivation**
It’s repetitive to open a GreasyFork/OpenUserJS tab, click many times, and copypaste the userscript from the Violentmonkey tab over.

**Proposed Solution**
Add a publish/push-update button like the Stylus extension has with userstyles.world. Ideally upon update, it’d handle everything in the background like Stylus, but I’dn’t be upset if it just opened a prefilled GreasyFork form.

**Use Cases**
I publish/update/maintain userscripts on GreasyFork. I’d like to perform those actions in 1 click.

danielzgtg:
See #2425

#### Before publish: Publish userscript button appears

publish userscript button appears above editor

#### After successful publish: Push update button appears

push update button appears above editor

#### 3rd-party read-only: No button added

no extra button next to READ-ONLY

#### 3rd-party modified: No button added

Anything that didn't go through my workflow is considered 3rd-party.

no extra button when we didn't go through my process

If you want to have it recognize a userscript as 1st-party:

1. Remove `@updateURL` and `@downloadURL`
1. Click "Publish userscript"
1. Close the tab that appears
1. Refresh the Greasy Fork tab that contains the green "Reinstall version" button

This way, upon reopening, the "Push update" button should appear

### DIFF
--- a/_locales/en/messages.yml
+++ b/_locales/en/messages.yml
@@ -45,6 +45,9 @@ buttonFilter:
   description: Button to show filters menu.
   message: Filters
   touched: false
+buttonGreasyForkPublish:
+  description: Button to upload the current userscript, creating a new one on Greasy Fork
+  message: Publish
 buttonHome:
   description: Button to open homepage.
   message: Homepage
@@ -64,6 +67,9 @@ buttonNew:
 buttonOK:
   description: OK button on dialog.
   message: OK
+buttonPushUpdate:
+  descrption: Button to upload the current code to replace a userscript's contents on Greasy Fork or etc.
+  message: Push update
 buttonRecycleBin:
   description: Button to list scripts in recycle bin.
   message: Recycle Bin
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -26,6 +26,7 @@ import storage, {
 import { storageCacheHas } from './storage-cache';
 import { reloadTabForScript } from './tabs';
 import { vetUrl } from './url';
+import { kDownloadURL, kPublishedByMe, kUpdateURL } from '@/options/utils';
 
 let maxScriptId = 0;
 let maxScriptPosition = 0;
@@ -42,15 +43,53 @@ export const sizesPrefixRe = RegExp(
 /** @type {{ [type: 'cache' | 'require']: { [url: string]: Promise<?> } }} */
 const pendingDeps = { [S_CACHE]: {}, [S_REQUIRE]: {} };
 const depsPorts = {};
+let listenForPublishScriptId = 0;
 
 addPublicCommands({
   GetScriptVer(opts) {
-    const script = getScript(opts);
-    return script
-      ? script.meta.version
-      : null;
+    try {
+      const script = getScript(opts);
+      if (!script) return null;
+      if (script.props.id === listenForPublishScriptId) void onScriptPublished(opts, script);
+      return script.meta.version;
+    } finally {
+      listenForPublishScriptId = 0;
+    }
   },
 });
+
+/**
+ * Adds @downloadURL after going through "Publish userscript" button
+ */
+async function onScriptPublished(opts, script) {
+  // For security, validate meta and edit local code, instead of fetching the downloadURL
+  const url = opts.meta.publishDownloadURL;
+  if (opts?.meta?.name !== script.meta.name || opts.meta.namespace !== script.meta.namespace
+  || opts.meta.publishVersion !== script.meta.version || script.custom[kUpdateURL] || script.meta[kUpdateURL]
+  || script.custom[kDownloadURL] || script.meta[kDownloadURL] || script.custom.lastInstallURL
+  || !url.startsWith('https://update.greasyfork.org/scripts/') || !url.endsWith('.user.js')
+  || url.includes('\n')) throw new Error;
+  const id = script.props.id;
+  const [ head, tail, ...excess ] = (await GetScriptCode(id)).split('\n// ==/UserScript==\n');
+  if (excess.length || head.includes('@downloadURL') || head.includes('@updateURL')) throw new Error;
+  await parseScript({
+    bumpDate: true,
+    code: `${head}
+// @downloadURL ${url}
+// @updateURL ${url.substr(0, url.length - 8)}.meta.js
+// ==/UserScript==
+${tail}`,
+    config: { ...script.config, shouldUpdate: 0 },
+    custom: { ...script.custom, [kPublishedByMe]: true },
+    id,
+    isNew: false,
+    message: '',
+  });
+}
+
+function GetScriptCode(id) {
+  return storage[S_CODE][Array.isArray(id) ? 'getMulti' : 'getOne'](id);
+}
 
 addOwnCommands({
   CheckPosition: sortScripts,
@@ -74,10 +113,9 @@ addOwnCommands({
     };
   },
   /** @return {Promise<string>} */
-  GetScriptCode(id) {
-    return storage[S_CODE][Array.isArray(id) ? 'getMulti' : 'getOne'](id);
-  },
+  GetScriptCode,
   GetTags: () => getScriptsTags(aliveScripts),
+  ListenForPublish: ({ id }) => void (listenForPublishScriptId = id),
   /** @return {Promise<void>} */
   async MarkRemoved({ id, removed }) {
     if (!removed) {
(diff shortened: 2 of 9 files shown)


tophf:
...or form's `target` may be the name of an invisible iframe.

tophf:
Ah so if it's similar to Stylus integration with userstyles.world then it might make sense to specify the window as a popup:
```js
const {availWidth: w, availHeight: h} = screen;
window.open('', '_blank', `popup=1,left=${w / 4},top=${h / 4},width=${w / 2},height=${h / 2}`);
```

danielzgtg:
@tophf Your idea of a "form element in the current page" simplified my code. I added automatic detection of the "Reinstall version" button after going through "Publish userscript", so showMessage and hints aren't needed anymore.

I added screenshots at the top of the 4 cases I tested working. This PR is now ready for others to test.

tophf:
I really want it to open inside an iframe inside the editor tab and judging by the HTTP response headers greasyfork doesn't forbid opening itself inside an iframe. Assuming it's added to the Vue template and declared as `const $iframe = ref()`it will redirect similarly to what you do now:
```js
const iframe = document.createElement('iframe');
const form = document.createElement('form');
const textarea = document.createElement('textarea');
form.hidden = true;
form.method = 'post';
textarea.name = 'script_version[code]';
textarea.value = $codeComp.getRealContent();
document.body.append(iframe);
iframe.contentDocument.body.append(form);
form.submit();
```

`Publish` is sufficient as a label because we already in a userscript editing mode. You can add a `title="greasyfork.org"` if you want, but it's probably unnecessary as it's clear from the UI that opens.

I think the label doesn't need to change to `Push update` even when the script is connected to greasyfork. We can probably colorize the button in this case to something greenish or bluish.

I also have a lot of trivial suggestions for the actual code but we can deal with that later or I can do it myself after we establish the fundamental UX. For example, in `injected` folder the code must avoid using any built-in methods that weren't safely saved, so `web/index.js` should be restored and instead `content/index.js` should do something like this:
```diff
diff --git a/src/injected/content/index.js b/src/injected/content/index.js
@@ -7 +7 @@ import './tabs';
-import { sendCmd } from './util';
+import { getAttribute, querySelector, sendCmd } from './util';
@@ -44 +44,12 @@ async function init() {
-    addHandlers({ GetScriptVer: true });
+    addHandlers({
+      GetScriptVer({ meta }) {
+        setPrototypeOf(meta, null);
+        const el = document::querySelector('a.install-link');
+        if (el && el::getAttribute('data-script-name') === meta.name
+        && el::getAttribute('data-script-namespace') === meta.namespace) {
+          meta.publishVersion = el::getAttribute('data-script-version');
+          meta.publishDownloadURL = el::getAttribute('href');
+        }
+        return sendCmd('GetScriptVer', { meta });
+      },
+    });
diff --git a/src/injected/content/inject.js b/src/injected/content/inject.js
@@ -2 +2,3 @@ import bridge, { addHandlers, grantless } from './bridge';
-import { elemByTag, makeElem, nextTask, onElement, sendCmd } from './util';
+import {
+  elemByTag, getAttribute, makeElem, nextTask, onElement, querySelector, sendCmd,
+} from './util';
@@ -19,2 +20,0 @@ let nonce;
-let getAttribute;
-let querySelector;
@@ -160,4 +159,0 @@ export async function injectScripts(data, info, isXml) {
-  if (wasInjectableFF) {
-    getAttribute = Element[PROTO].getAttribute;
-    querySelector = document.querySelector;
-  }
diff --git a/src/injected/content/util.js b/src/injected/content/util.js
@@ -10,0 +11,2 @@ export const elemByTag = (tag, i) => getOwnProp(document::getElementsByTagName(t
+export const getAttribute = Element[PROTO].getAttribute;
+export const querySelector = document.querySelector;
```

danielzgtg:
> I really want it to open inside an iframe inside the editor tab

I started then stopped working on that. It could've allowed easy window.location instead of GetScriptVer, but I don't think the cookies will work. Here's a photoshopped mockup:

iframe maybe

> judging by the HTTP response

At first, Greasy Fork will redirect to a login screen. I use GitHub and Google login. There's also preexisting cookies, cache partitioning, and CHIPS. Unless extensions are special, I think it might not work. Besides, you suggested "target=_blank" above.

> Publish is sufficient as a label

Done

> title="greasyfork.org" if you want

It can be added by any PR that adds OpenUserJS publish.

> I think the label doesn't need to change to Push update

Please let us have this feature that Stylus has. I don't want to mistakenly publish my private potentially password-containing userscripts when I intended to update one of my public userscripts. Making the text the same will lead to desensitization to the seriousness of "Publish", exacerbated by Greasy Fork's layout and colours being the same on that page for both actions. 

> diff

Applied manually and rolled back `src/injected/web/index.js`. Somehow `git apply` threw `error: patch fragment without header at line 3: @@ -7 +7 @@ import './tabs';`.

tophf:
Yeah an iframe won't work because the site sets X-Frame-Options header to 'sameorigin', and I guess Jason won't be too keen on us stripping it.

What about using const wnd = window.open() with the size parameter set to something like a half of screen size in each dimension, putting the form inside via wnd.document, and submitting it?

> Making the text the same will lead to desensitization

That's a big claim, but I don't think that having different text will meaningfully reduce the possibility of a mistake. The most reliable solution is to not put sensitive info into publishable scripts in the first place e.g. maybe use GM storage instead. We can also add a confirmation on the first publish attempt within this editing session reminding about sensitive info.

JasonBarnabe:
> Yeah an iframe won't work because the site sets X-Frame-Options header to 'sameorigin', and I guess Jason won't be too keen on us stripping it.

If there's a change I can make that'll make your life easier without leaving the door wide open for bad actors, I'll do it. I'm not sure what origin I would need to allow for this to work for you.

danielzgtg:
> What about using const wnd = window.open()

I brought back my original window.open code, simplified, then made it a new window as described:

popup prefill

> not put sensitive info into publishable scripts in the first place e.g. maybe use GM storage instead

There are things like `@match` that must be in the userscript itself. Greasy Fork allows "Adult content", and user e.g. bible_thumping_republican_politician123 might be embarrassed if he forgot to switch browsers. Seeing "Publish" if he usually sees "Push update" will make it more likely that he will think twice.

tophf:
Yeah looks fine.

> things like @match that must be in the userscript itself

If we are still in the context of sensitive things that shouldn't be put into a greasyfork-connected script, such `@match` should be specified in the settings of the userscript, not in its code, because it's very easy to forget to remove it before publishing.

BTW, I wonder if authorization should be performed via browser.identity.launchWebAuthFlow the way we do it in Stylus...

@JasonBarnabe, a separate window seems fine as it's usually how it's done anyway.

danielzgtg:
> browser.identity.launchWebAuthFlow

I'm not familiar with the internals of greasyfork.org's authentication. But the current prefill workflow works fine.

Are there any remaining blockers for this feature?

tophf:
I'll get to it after publishing MV3 stable. One thing I think that needs changing is `publishedByMe` because theoretically we might support multiple publishing targets, so it might make sense to have something like `published: "greasyfork:123,blabla:345"`.

danielzgtg:
> needs changing is `publishedByMe` [...] `published: "greasyfork:123,blabla:345"`

`publishedByMe` is not the source of truth. It just unhides a button labelled "Push update".

> support multiple publishing targets

`@downloadURL` is the source of truth. The selected publishing target for the current userscript has its hostname after `@downloadURL`. Once a second target is added to ViolentMonkey, `update.greasyfork.org` would simply be replaced by a different hostname.

OpenUserJS sets `@downloadURL https://openuserjs.org/install/<username>/<title>.user.js`. Adding support involves expanding the "Publish" button, not "Push Update". I didn't design support for pushing the same userscript to both Greasy Fork and OpenUserJS. They would fight over `@downloadURL`, necessitating your hidden state. It's more complicated and risks confusion as to which version is from where and should go where. Publishing two different userscripts, each to a different website, is OK however.

tophf:
My idea is that we can push a script without `@downloadURL` to multiple sites to avoid duplication.

danielzgtg:
I remembered why I didn't implement hidden state. One could cause an identity crisis by cutting one userscript's code and pasting it into another's. I will avoid it this time by exposing the publish URLs on the Settings tab in newline-separated values in a textbox.

I'll refactor `publishedByMe: boolean` to `publishURLs: string[]`. `published: string | Record<string, number>` mightn't work for OpenUserJS. UI changes to the Publish/Push-update buttons will be deferred to whoever adds OUJS support.

PR #2459, opened 2026-03-04
MV3 Conversion

Zulenka:
Currently works with BSP. I am continuing to test other scripts and test the overall system. 

I added some improvements to the settings page to help with debugging. There is a console that displays MV3 health and an error, info, debug console for displaying script injection errors and CSP errors. 

Current working version is Chrome only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "violentmonkey",
   "description": "Violentmonkey",
-  "version": "2.35.1",
+  "version": "2.35.2",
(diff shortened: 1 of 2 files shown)


tophf:
It probably can't include the name so maybe Saturnmonkey :-) Either way, any properly performant implementation should use browser.userScripts.register API **for every userscript** and this code then will be executed by one main controller script of the extension. This is architecturally different to the current GetInjected approach and would allow a real document_start.

PR #2490, opened 2026-04-08
fix #1407

### LINKED ISSUE #1407: [Feature] GM_registerMenuCommand should display in contextmenus like Tampermonkey
### What is the problem?
Sorry, my English is very poor. The following content comes from Baidu translation

found it in Violentmonkey extension icon click popop menu item "test menu item"

but it's a new Window, It is expected that can right-click an element in HTML to further control it. However, when a new window is opened, it is difficult to directly associate and control the clicked target item

### How to reproduce it?
1. let  menuFunRef = function(){};
2. GM_registerMenuCommand('test menu item", menuFunRef);
3. open a this script can execute web page.
4.  mouse right click any element popup chrome mouse right menu item list
5.  now, 'test menu item" not exists in this menu list

Tampermonkey is support, hope violentmonkey support too!

this menu list exists 'test menu item" 

not exists

### What is the result in the [upcoming release](https://github.com/violentmonkey/violentmonkey/releases)?<br>WARNING! Do an export in settings before testing!
not exists

### Environment
- Browser: chrome
- Browser version: 94.0.4606.81（正式版本） (x86_64)
- Violentmonkey version:v2.13.0
- OS: macOS Catalina 10.15.7

117649:
This PR add right click menu commands registration implementation requested in #1407 .
I've been migrated from tampermonkey and lack of this feature was a big roadblock to me.
Thus, I let cursor implement it for me. To my surprise it spit out working codes at first go but I have to fix loose bits here and there myself after I run out free quota in like 2 prompts.
Now I've run this for several days and no problems that I could reproduce appears.
So I think its time this become a problem of not only myself.

P.S. I don't wrote all those comments, I don't really have an idea what they are talking about.

### DIFF
--- a/src/_locales/en/messages.yml
+++ b/src/_locales/en/messages.yml
@@ -694,6 +694,9 @@ menuCommands:
   description: Menu item to list script commands.
   message: Script commands
   touched: false
+labelPageMenuCommands:
+  description: Checkbox item in page context menu and option to enable script commands.
+  message: Show script commands in page menu
 menuDashboard:
   description: Label for menu item to open dashboard.
   message: Open Dashboard
--- a/src/background/utils/icon.js
+++ b/src/background/utils/icon.js
@@ -7,6 +7,7 @@ import { popupTabs } from './popup-tracker';
 import storage, { S_CACHE } from './storage';
 import { forEachTab, getTabUrl, injectableRe, openDashboard, tabsOnRemoved, tabsOnUpdated } from './tabs';
 import { testBlacklist } from './tester';
+import { tryHandlePageMenuCommand } from './page-context-menu-commands';
 import { FIREFOX, ua } from './ua';
 
 /** 1x + HiDPI 1.5x, 2x */
@@ -124,7 +125,8 @@ init.then(async () => {
   }
 });
 
-contextMenus?.onClicked.addListener(({ menuItemId: id }, tab) => {
+contextMenus?.onClicked.addListener(({ menuItemId: id, frameId }, tab) => {
+  if (tryHandlePageMenuCommand(id, tab, frameId)) return;
   handleHotkeyOrMenu(id, tab);
 });
 tabsOnRemoved.addListener(id => delete badges[id]);
--- a/src/background/utils/page-context-menu-commands.js
+++ b/src/background/utils/page-context-menu-commands.js
@@ -1,0 +1,298 @@
+import { browserWindows, getActiveTab, i18n, ignoreChromeErrors, isEmpty, sendTabCmd } from '@/common';
+import { browser } from '@/common/consts';
+import { nest } from '@/common/object';
+import { kFrameId } from '@/common/safe-globals';
+import { addPublicCommands, init } from './init';
+import { tabsOnActivated, tabsOnRemoved } from './tabs';
+import { getScriptById } from './db';
+import { getOption, hookOptions } from './options';
+
+/**
+ * Page / frame context menu entries for GM_registerMenuCommand.
+ *
+ * ## End-to-end flow
+ * 1. A userscript calls GM_registerMenuCommand in the injected (page) world; the content script
+ *    keeps a `menus` map and notifies the background via UpdateTabMenuCommands (see gm-api-content).
+ * 2. This module stores that snapshot per (tabId, frameId), merges frames like the popup does, and
+ *    mirrors the merged commands as children under a fixed parent item (VM_PAGE_CMD_ROOT).
+ * 3. When the user picks an entry, contextMenus.onClicked runs; icon.js delegates here by id prefix.
+ *    We verify the tab id baked into the item id matches the tab Chrome passed to onClicked, then
+ *    sendTabCmd(..., 'Command', ...) to the frame that registered that command so the same callback
+ *    runs as from the extension popup.
+ *
+ * ## Why we rebuild the menu instead of using contextMenus.onShown
+ * Chrome’s chrome.contextMenus API does not expose onShown. Firefox has menus.onShown, but we target
+ * Chrome as well. The extension context menu is global (not per-tab in the API), so we maintain one
+ * set of dynamic items that we remove/recreate whenever the “display target” tab should change.
+ *
+ * ## What tab the menu represents (approximation)
+ * We rebuild so the list matches the active tab of the focused browser window most of the time:
+ * - tabs.onActivated: user switched tab inside a window.
+ * - windows.onFocusChanged: user focused another window; we query that window’s active tab.
+ * - UpdateTabMenuCommands: if the tab that changed is the current getActiveTab(), refresh the list.
+ * - tabs.onUpdated (url): cleared script state for that tab; refresh if it was the active tab.
+ * - tabs.onRemoved: drop that tab’s data; rebuild for whichever tab is active now.
+ * - init: after creating the root item, one getActiveTab() so startup isn’t empty until a tab switch.
+ *
+ * ## Limitations
+ * - Right-clicking a non-active tab (e.g. split view or rare UI paths) can still show commands for
+ *   the active tab; fixing that without onShown would require per-URL patterns or other hacks.
+ * - Service worker restarts drop in-memory tabFrameMenus/routeTab until scripts call
+ *   GM_registerMenuCommand again (same class of issue as other ephemeral BG state).
+ * - Menu item ids embed tabId + scriptId + key; encodeURIComponent(key) keeps odd keys safe.
+ * - Per-script submenu (script display name) groups command captions; Chrome removes child items when
+ *   the submenu id is removed.
+ * - When a script is disabled or removed (db.updateScriptInfo), BG purges cached menus for all tabs/
+ *   frames and tabs get PurgeScriptMenus so the popup’s menus map stays in sync.
+ */
+
+const contextMenus = chrome.contextMenus;
+
+const KEY_PAGE_MENU_COMMANDS = 'pageMenuCommands';
+const VM_PAGE_CMD_ROOT = 'vmPageCmdRoot';
+const VM_SCRIPT_SUB_PREFIX = 'vmScriptSub:';
+const CMD_PREFIX = 'vmCmd:';
+const MAX_TITLE_LEN = 250;
+
+/** Per-tab, per-frame snapshots of `menus` from the content script (scriptId -> commandKey -> opts). */
+const tabFrameMenus = {};
+/**
+ * For each tab, maps `${scriptId}:${commandKey}` -> frameId where that command was last registered.
+ * Needed so sendTabCmd targets the same frame as the popup’s idMap resolution.
+ */
+const routeTab = {};
+/** Script submenu ids under VM_PAGE_CMD_ROOT (removing an id drops its command children too). */
+let dynamicSubmenuIds = [];
+let isEnabled = false;
+let isMenuCreated = false;
+
+/** Merge all frames’ menu snapshots for one tab; later frame ids overwrite same script/command keys. */
+function mergeTabMenus(tabId) {
+  const byFrame = tabFrameMenus[tabId];
+  if (!byFrame) return Object.create(null);
+  const frameIds = Object.keys(byFrame).map(Number);
+  const merged = Object.create(null);
+  for (const fid of frameIds) {
+    const m = byFrame[fid];
+    if (!m) continue;
+    for (const sid in m) {
+      const hub = merged[sid] || (merged[sid] = Object.create(null));
+      Object.assign(hub, m[sid]);
+    }
+  }
+  return merged;
+}
+
+/** Command row title (script name is the parent submenu). */
+function formatCommandTitle(item) {
+  const s = item.text || '';
+  return s.length > MAX_TITLE_LEN ? `${s.slice(0, MAX_TITLE_LEN - 1)}...` : s;
+}
+
+/** Submenu title from first registered command’s displayName, or fallback. */
+function formatScriptSubmenuTitle(hub, scriptId) {
+  for (const key in hub) {
+    const name = hub[key]?.displayName;
+    if (name) {
+      const s = `${name}`;
+      return s.length > MAX_TITLE_LEN ? `${s.slice(0, MAX_TITLE_LEN - 1)}...` : s;
+    }
+  }
+  const fb = `#${scriptId}`;
+  return fb.length > MAX_TITLE_LEN ? `${fb.slice(0, MAX_TITLE_LEN - 1)}...` : fb;
+}
+
+/**
+ * Replace dynamic submenus for `tabId`: root → script submenu → command rows.
+ * We hide the root when there are no commands for this tab.
+ */
+function rebuildPageCommandMenuTab(tabId) {
+  if (!contextMenus || tabId == null) return;
+  if (!isEnabled) {
+    return;
+  }
+  if (!isMenuCreated) return;
+  for (const id of dynamicSubmenuIds) {
+    contextMenus.remove(id, ignoreChromeErrors);
+  }
+  dynamicSubmenuIds = [];
+  const merged = mergeTabMenus(tabId);
+  let commandCount = 0;
+  for (const scriptId in merged) {
+    const hub = merged[scriptId];
+    const script = getScriptById(scriptId);
+    let title = formatScriptSubmenuTitle(hub, scriptId);
+    if (!script?.config.enabled) {
+      title = `⛔ ${title}`;
+    }
+    const subId = `${VM_SCRIPT_SUB_PREFIX}${tabId}:${scriptId}`;
+    contextMenus.create({
+      id: subId,
+      parentId: VM_PAGE_CMD_ROOT,
+      title,
+      contexts: ['all'],
+    }, ignoreChromeErrors);
+    dynamicSubmenuIds.push(subId);
+    for (const key in hub) {
+      const item = hub[key];
+      const id = `${CMD_PREFIX}${tabId}:${scriptId}:${encodeURIComponent(key)}`;
+      contextMenus.create({
+        id,
+        parentId: subId,
+        title: formatCommandTitle(item),
+        contexts: ['all'],
+      }, ignoreChromeErrors);
+      commandCount += 1;
+    }
+  }
+  contextMenus.update(VM_PAGE_CMD_ROOT, { visible: commandCount > 0 }, ignoreChromeErrors);
+}
+
+/** If the tab whose menus changed is the active tab (current window), refresh visible context items. */
+function maybeRebuildForActiveTab(updatedTabId) {
+  if (!isEnabled) return;
+  getActiveTab().then((active) => {
+    if (active?.id === updatedTabId) rebuildPageCommandMenuTab(updatedTabId);
+  });
+}
+
+addPublicCommands({
+  /** Sync in-memory menu state from a content-script frame; see file comment for merge/routing rules. */
+  UpdateTabMenuCommands({ menus, reset }, { tab, [kFrameId]: frameId, [kTop]: isTop }) {
+    if (!tab?.id) return;
+    const routes = nest(routeTab, tab.id);
+    const byTab = nest(tabFrameMenus, tab.id);
+    if (isEmpty(menus)) {
+      if (reset && isTop) {
+        delete tabFrameMenus[tab.id];
+        delete routeTab[tab.id];
+      } else {
+        delete byTab[frameId];
+        if (isEmpty(byTab)) {
+          delete tabFrameMenus[tab.id];
+          delete routeTab[tab.id];
+        }
+      }
+    } else {
+      const oldFrameMenus = byTab[frameId];
+      if (oldFrameMenus) {
+        for (const sid in oldFrameMenus) {
+          for (const key in oldFrameMenus[sid]) {
+            const rkey = `${sid}:${key}`;
+            if (routes[rkey]?.[frameId]) delete routes[rkey][frameId];
+          }
+        }
+      }
+      byTab[frameId] = menus;
+      for (const sid in menus) {
+        for (const key in menus[sid]) {
+          nest(routes, `${sid}:${key}`)[frameId] = true;
+        }
+      }
+    }
+    maybeRebuildForActiveTab(tab.id);
+  },
+});
+
+// Drop closed tab’s state, then repoint the global menu at the new active tab (if any).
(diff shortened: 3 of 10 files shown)


tophf:
Thanks, but it's better to use `mousedown` and `keydown` events on `window` for the moment context menu is invoked in the content script to send a message to rebuild the menu instead of wastefully doing it every time a tab is switched. Maybe your AI is smart enough to use my suggestion and produce acceptable code? Otherwise since this is a nontrivial AI-generated solution, where half the code is redundant and the other one might cause unknown issues, I'd have to spend approximately the same amount of time reworking this PR as I would for implementing it from scratch.

117649:
> Thanks, but it's better to use `mousedown` and `keydown` events on `window` for the moment context menu is invoked in the content script to send a message to rebuild the menu instead of wastefully doing it every time a tab is switched. 

Does chrome even allow you do that?

AI suggested use Menus.oShown for FF and fantasy of Chrome also have it until I point that out.

117649:
>  I'd have to spend approximately the same amount of time reworking this PR as I would for implementing it from scratch.

How long that may be?

tophf:
There are examples of using these events on stackoverflow. I never know when I'm about to do something, sorry.

117649:
> There are examples of using these events on stackoverflow.

> Thanks, but it's better to use `mousedown` and `keydown` events on `window` for the moment context menu is invoked in the content script to send a message to rebuild the menu instead of wastefully doing it every time a tab is switched. Maybe your AI is smart enough to use my suggestion and produce acceptable code? 

After some simple tests that way is pretty busted.
First, on some site you just can't catch a right click event. 'contextmenu' can work for most time however. 
But, here comes the Second there just not enough time for context menu to be updated before it opens.
Drop the real event then wait for update then synthesize a fake one to open the contextmenu? NO way yet to found.

tophf:
That's because you need to listen in the capture phase (`true` in the registration) before the event is processed:

```js
window.addEventListener('mousedown', e => {
  if (e.isTrusted && e.button === 2 && !e.altKey && !e.shiftKey && !e.ctrlKey && !e.metaKey) {
    sendCmd(..........);
  }
}, true);
```

117649:
> That's because you need to listen in the capture phase (`true` in the registration) before the event is processed:
> 
> ```js
> window.addEventListener('mousedown', e => {
>   if (e.isTrusted && e.button === 2 && !e.altKey && !e.shiftKey && !e.ctrlKey && !e.metaKey) {
>     sendCmd(..........);
>   }
> }, true);
> ```

Unapologetically, I did no other way.

You can as always choose to not believe but you still won't pass the 2nd point.

tophf:
The listener should be also registered in the first tick of the event loop so that it happens before the page adds its own listeners that can drop the event e.g. at the end of gm-api-content.js‎. If this doesn't help, I guess it might be a Firefox bug, but in Firefox it's better to have a separate simpler implementation via browser.menus.onShown anyway.

tophf:
Also make sure `window` is used, not `document`.

117649:
> The listener should be also registered in the first tick of the event loop so that it happens before the page adds its own listeners that can drop the event e.g. at the end of gm-api-content.js‎. If this doesn't help, I guess it might be a Firefox bug, but in Firefox it's better to have a separate simpler implementation via browser.menus.onShown anyway.

I'm not doing it on firefox.

tophf:
After testing it myself I see my idea is not reliable for more than a couple of items and it may be even OS-dependent. So apparently in Chrome the only reliable solution is the way it's done in the PR, but since recreating a potentially large menu is wasteful (some users may have dozens of scripts running in a tab) I'd like it to be disabled by default with an option to enable it. The logical place for an option is the right click menu itself and a duplicate option in the global extension options somewhere. Another thing to note is tabsOnUpdated.addListener needs to only process changeInfo parameter that has `url`.

@gera2ld, WDYT?

gera2ld:
I believe this is the correct approach, similar to the active script count badge.
I also agree that it should be disabled by default, as the list can become lengthy and not everyone may need it (including myself).

117649:
>  Another thing to note is tabsOnUpdated.addListener needs to only process changeInfo parameter that has `url`.

FYI, that's how original generated code does.
I have to remove it because it won't remove the menu item if you: disable the script then refresh the page.

Guessing no url change during that.

tophf:
Looks like it should be done inside preinject.js -> Run -> by adding something like `if (reset) resetMenus(src)`

117649:
> Then it seems better to remove the specific menu item directly when disabling a script.

Hit the badge open the popup and disable a script there. Have you removed the item there?
No.

The fact is you didn't rip script injected out from the page when you disable it.
Thus, until you refresh the page or itself call unregister the menu item remains legit.

I've worked the path you've mentioned and it's not feel right.
I choose to add a ⛔ prefix to the disabled script's submenu title. It's in the pr.

117649:
> Looks like it should be done inside preinject.js -> Run -> by adding something like `if (reset) resetMenus(src)`

I'm not understanding it. I don't know what its doing.
I tried put a breakpoint in it, only managed to hit it by refresh the page. It would hit twice every time.

tophf:
The important thing is `reset` being set, which happens on the first call, IIRC.

117649:
You mean this?

```diff
diff --git a/src/background/utils/page-context-menu-commands.js b/src/background/utils/page-context-menu-commands.js
index 3855b427..cab6d5a6 100644
--- a/src/background/utils/page-context-menu-commands.js
+++ b/src/background/utils/page-context-menu-commands.js
@@ -189,11 +189,6 @@ tabsOnRemoved.addListener((tabId) => {
   });
 });
 
-// Full navigation wipes injected state for that tab; refresh menu if that tab is still active.
-tabsOnUpdated.addListener((tabId) => {
-  sendTabCmd(tabId, 'UpdateScriptMenus');
-});
-
 // Primary signal: which tab is active in this window after the user selects a tab.
 tabsOnActivated.addListener(({ tabId }) => {
   rebuildPageCommandMenuTab(tabId);
diff --git a/src/background/utils/preinject.js b/src/background/utils/preinject.js
index 485aa984..c3be4527 100644
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -233,6 +233,7 @@ addPublicCommands({
     if (reset === 'bfcache' && hasIds) {
       addValueOpener(ids, tabId, getFrameDocId(isTop, docId, src[kFrameId]));
     }
+    sendTabCmd(tabId, 'UpdateScriptMenus');
   },
 });
 

```

tophf:
I meant `if (reset) sendTabCmd(tabId, 'UpdateScriptMenus');` but actually it's even better to do in cmd-run.js: 

```js
async function report(delay, reset = !sent) {
  ...................
  sendCmd('Run', { reset, [IDS]: runningIds, menus: reset && menus });
  ................
```

Then in preinject.js:
```js
  Run({ [IDS]: ids, menus, reset }, src) {
    ...................
    if (menus) {
      commands.UpdateTabMenuCommands(menus, src);
    }
  },
```

tophf:
Furthermore, with the above change there's probably no need to expose a separate UpdateTabMenuCommands command, but instead it can be just an exported function that will be called directly from Run.

117649:
I did.
```diff
diff --git a/src/background/utils/db.js b/src/background/utils/db.js
index 7758a082..5f5e0715 100644
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -584,13 +584,6 @@ export async function updateScriptInfo(id, data) {
     storage.api.set({ [S_SCRIPT_PRE + id]: script }),
     sendCmd('UpdateScript', { where: { id }, update: script }),
   ]);
-  if (data.config && (data.config.enabled != null || data.config.removed)) {
-    await forEachTab((tab) => {
-      if (injectableRe.test(getTabUrl(tab))) {
-        sendTabCmd(tab.id, 'UpdateScriptMenus').catch(ignoreChromeErrors);
-      }
-    });
-  }
 }
 
 /**
diff --git a/src/background/utils/page-context-menu-commands.js b/src/background/utils/page-context-menu-commands.js
index 3855b427..cab6d5a6 100644
--- a/src/background/utils/page-context-menu-commands.js
+++ b/src/background/utils/page-context-menu-commands.js
@@ -189,11 +189,6 @@ tabsOnRemoved.addListener((tabId) => {
   });
 });
 
-// Full navigation wipes injected state for that tab; refresh menu if that tab is still active.
-tabsOnUpdated.addListener((tabId) => {
-  sendTabCmd(tabId, 'UpdateScriptMenus');
-});
-
 // Primary signal: which tab is active in this window after the user selects a tab.
 tabsOnActivated.addListener(({ tabId }) => {
   rebuildPageCommandMenuTab(tabId);
diff --git a/src/background/utils/preinject.js b/src/background/utils/preinject.js
index 485aa984..32461940 100644
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -214,7 +214,7 @@ addPublicCommands({
       [S_CACHE]: envCache,
     };
   },
-  Run({ [IDS]: ids, reset }, src) {
+  Run({ [IDS]: ids, menus, reset }, src) {
     const {
       [kDocumentId]: docId,
       [kTop]: isTop,
@@ -233,6 +233,9 @@ addPublicCommands({
     if (reset === 'bfcache' && hasIds) {
       addValueOpener(ids, tabId, getFrameDocId(isTop, docId, src[kFrameId]));
     }
+    if (menus) {
+      commands.UpdateTabMenuCommands(menus, src);
+    }
   },
 });
 
diff --git a/src/injected/content/cmd-run.js b/src/injected/content/cmd-run.js
index fe9e8838..d8ba628a 100644
--- a/src/injected/content/cmd-run.js
+++ b/src/injected/content/cmd-run.js
@@ -42,7 +42,7 @@ export function Run(id, realm) {
 async function report(delay, reset = !sent) {
   while (--delay >= 0) await nextTask();
   // not awaiting to clear `pending` immediately
-  sendCmd('Run', { reset, [IDS]: runningIds });
+  sendCmd('Run', { reset, [IDS]: runningIds, menus: reset && menus });
   sendSetPopup(!!pending);
   pending = false;
   sent = true;

```

But certainly, break something. Because badge stay gray and popup keep asking reload tab.
Nothing screaming in console though.

117649:
And breakpoint set in preinject.js won't hit anymore anyways.

tophf:
Well, in your code it should be `commands.UpdateTabMenuCommands({menus}, src)` or you can replace `{menus}` with `menus` in the definition of UpdateTabMenuCommands.

117649:
Really important question is 'RegisterMenu()' goanna be run during every refresh, back, forward?

117649:
This is what I've managed to made work:
```diff
diff --git a/src/background/utils/db.js b/src/background/utils/db.js
index 7758a082..85d7f9b4 100644
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -1,6 +1,6 @@
 import {
   dataUri2text, i18n, getScriptHome, isDataUri,
-  getScriptName, getScriptsTags, getScriptUpdateUrl, isRemote, sendCmd, sendTabCmd, trueJoin,
+  getScriptName, getScriptsTags, getScriptUpdateUrl, isRemote, sendCmd, trueJoin,
   getScriptPrettyUrl, getScriptRunAt, makePause, isValidHttpUrl,
   ignoreChromeErrors,
 } from '@/common';
@@ -22,7 +22,7 @@ import storage, {
   getStorageKeys,
 } from './storage';
 import { storageCacheHas } from './storage-cache';
-import { forEachTab, getTabUrl, injectableRe, reloadTabForScript } from './tabs';
+import { reloadTabForScript } from './tabs';
 import { vetUrl } from './url';
 
 let maxScriptId = 0;
@@ -584,13 +584,6 @@ export async function updateScriptInfo(id, data) {
     storage.api.set({ [S_SCRIPT_PRE + id]: script }),
     sendCmd('UpdateScript', { where: { id }, update: script }),
   ]);
-  if (data.config && (data.config.enabled != null || data.config.removed)) {
-    await forEachTab((tab) => {
-      if (injectableRe.test(getTabUrl(tab))) {
-        sendTabCmd(tab.id, 'UpdateScriptMenus').catch(ignoreChromeErrors);
-      }
-    });
-  }
 }
 
 /**
diff --git a/src/background/utils/page-context-menu-commands.js b/src/background/utils/page-context-menu-commands.js
index 3855b427..39dab7d4 100644
--- a/src/background/utils/page-context-menu-commands.js
+++ b/src/background/utils/page-context-menu-commands.js
@@ -3,7 +3,7 @@ import { browser } from '@/common/consts';
 import { nest } from '@/common/object';
 import { kFrameId } from '@/common/safe-globals';
 import { addPublicCommands, init } from './init';
-import { tabsOnActivated, tabsOnRemoved, tabsOnUpdated } from './tabs';
+import { tabsOnActivated, tabsOnRemoved } from './tabs';
 import { getScriptById } from './db';
 
 /**
@@ -149,26 +149,31 @@ function maybeRebuildForActiveTab(updatedTabId) {
 
 addPublicCommands({
   /** Sync in-memory menu state from a content-script frame; see file comment for merge/routing rules. */
-  UpdateTabMenuCommands({ menus }, { tab, [kFrameId]: frameId }) {
+  UpdateTabMenuCommands(menus, { tab, [kFrameId]: frameId, [kTop]: isTop }) {
     if (!tab?.id) return;
     const routes = nest(routeTab, tab.id);
     const byTab = nest(tabFrameMenus, tab.id);
-    const oldFrameMenus = byTab[frameId];
-    if (oldFrameMenus) {
-      for (const sid in oldFrameMenus) {
-        for (const key in oldFrameMenus[sid]) {
-          const rkey = `${sid}:${key}`;
-          if (routes[rkey] === frameId) delete routes[rkey];
-        }
-      }
-    }
     if (isEmpty(menus)) {
-      delete byTab[frameId];
-      if (isEmpty(byTab)) {
+      if (isTop) {
         delete tabFrameMenus[tab.id];
         delete routeTab[tab.id];
+      } else {
+        delete byTab[frameId];
+        if (isEmpty(byTab)) {
+          delete tabFrameMenus[tab.id];
+          delete routeTab[tab.id];
+        }
       }
     } else {
+      const oldFrameMenus = byTab[frameId];
+      if (oldFrameMenus) {
+        for (const sid in oldFrameMenus) {
+          for (const key in oldFrameMenus[sid]) {
+            const rkey = `${sid}:${key}`;
+            if (routes[rkey] === frameId) delete routes[rkey];
+          }
+        }
+      }
       byTab[frameId] = menus;
       for (const sid in menus) {
         for (const key in menus[sid]) {
@@ -189,11 +194,6 @@ tabsOnRemoved.addListener((tabId) => {
   });
 });
 
-// Full navigation wipes injected state for that tab; refresh menu if that tab is still active.
-tabsOnUpdated.addListener((tabId) => {
-  sendTabCmd(tabId, 'UpdateScriptMenus');
-});
-
 // Primary signal: which tab is active in this window after the user selects a tab.
 tabsOnActivated.addListener(({ tabId }) => {
   rebuildPageCommandMenuTab(tabId);
diff --git a/src/background/utils/preinject.js b/src/background/utils/preinject.js
index 485aa984..e05a6a7e 100644
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -12,7 +12,7 @@ import {
 } from '@/common/object';
 import { CACHE_KEYS, getScriptsByURL, kTryVacuuming, PROMISE, REQ_KEYS, VALUE_IDS } from './db';
 import { setBadge } from './icon';
-import { addOwnCommands, addPublicCommands } from './init';
+import { addOwnCommands, addPublicCommands, commands } from './init';
 import { clearNotifications } from './notifications';
 import { hookOptionsInit } from './options';
 import { popupTabs } from './popup-tracker';
@@ -214,7 +214,7 @@ addPublicCommands({
       [S_CACHE]: envCache,
     };
   },
-  Run({ [IDS]: ids, reset }, src) {
+  Run({ [IDS]: ids, menus, reset }, src) {
     const {
       [kDocumentId]: docId,
       [kTop]: isTop,
@@ -233,6 +233,9 @@ addPublicCommands({
     if (reset === 'bfcache' && hasIds) {
       addValueOpener(ids, tabId, getFrameDocId(isTop, docId, src[kFrameId]));
     }
+    if (reset) {
+      commands.UpdateTabMenuCommands(menus, src);
+    }
   },
 });
 
diff --git a/src/injected/content/cmd-run.js b/src/injected/content/cmd-run.js
index fe9e8838..fc6e1959 100644
--- a/src/injected/content/cmd-run.js
+++ b/src/injected/content/cmd-run.js
@@ -1,5 +1,5 @@
 import bridge, { addHandlers, onScripts } from './bridge';
-import { sendSetPopup } from './gm-api-content';
+import { sendSetPopup, menus } from './gm-api-content';
 import { nextTask, sendCmd } from './util';
 
 const getPersisted = describeProperty(PageTransitionEvent[PROTO], 'persisted').get;
@@ -42,7 +42,7 @@ export function Run(id, realm) {
 async function report(delay, reset = !sent) {
   while (--delay >= 0) await nextTask();
   // not awaiting to clear `pending` immediately
-  sendCmd('Run', { reset, [IDS]: runningIds });
+  sendCmd('Run', { reset, [IDS]: runningIds, menus });
   sendSetPopup(!!pending);
   pending = false;
   sent = true;
diff --git a/src/injected/content/gm-api-content.js b/src/injected/content/gm-api-content.js
index 804a1ff5..32198d70 100644
--- a/src/injected/content/gm-api-content.js
+++ b/src/injected/content/gm-api-content.js
@@ -2,7 +2,7 @@ import bridge, { addBackgroundHandlers, addHandlers, grantless } from './bridge'
 import { addNonceAttribute } from './inject';
 import { decodeResource, elemByTag, makeElem, nextTask, sendCmd } from './util';
 
-const menus = createNullObj();
+export const menus = createNullObj();
 const HEAD_TAGS = ['script', 'style', 'link', 'meta'];
 const { toLowerCase } = '';
 const { [IDS]: ids } = bridge;
@@ -19,10 +19,6 @@ addBackgroundHandlers({
     }
     sendSetPopup();
   },
-
-  UpdateScriptMenus() {
-    sendCmd('UpdateTabMenuCommands', { menus });
-  },
 }, true);
 
 addHandlers({
@@ -50,13 +46,13 @@ addHandlers({
   RegisterMenu({ id, key, val }) {
     (menus[id] || (menus[id] = createNullObj()))[key] = val;
     sendSetPopup(true);
-    sendCmd('UpdateTabMenuCommands', { menus });
+    sendCmd('UpdateTabMenuCommands', menus);
   },
 
   UnregisterMenu({ id, key }) {
     delete menus[id]?.[key];
     sendSetPopup(true);
-    sendCmd('UpdateTabMenuCommands', { menus });
+    sendCmd('UpdateTabMenuCommands', menus);
   },
 });
 
```

117649:
> I meant `if (reset) sendTabCmd(tabId, 'UpdateScriptMenus');` but actually it's even better to do in cmd-run.js:
> 
> ```js
> async function report(delay, reset = !sent) {
>   ...................
>   sendCmd('Run', { reset, [IDS]: runningIds, menus: reset && menus });
>   ................
> ```
> 
> Then in preinject.js:
> 
> ```js
>   Run({ [IDS]: ids, menus, reset }, src) {
>     ...................
>     if (menus) {
>       commands.UpdateTabMenuCommands(menus, src);
>     }
>   },
> ```

At least, I found you'll need empty menus when disabled last script with menu then reload tab.

117649:
Updated so the command on the frame clicked would be try called first, then others that available if nothing there.
I think this is the right way doing it? @tophf

117649:
Anything?

tophf:
Yep sorry, I'm still lazy because I'd have to add an option as discussed above, so if you could implement it, it'be harder for me to procrastinate further.

117649:
> Yep sorry, I'm still lazy because I'd have to add an option as discussed above, so if you could implement it, it'be harder for me to procrastinate further.

I really don't know a thing about how to do that. Even I've already gain so much insight on this project that I never asked for.
Guess if I want to try something it'll have to be next month after my free quota on AI tools refresh.

cyfung1031:
> Thanks, but it's better to use `mousedown` and `keydown` events on `window` for the moment context menu is invoked in the content script to send a message to rebuild the menu instead of wastefully doing it every time a tab is switched. 

`chrome.contextMenus` is the correct API for making the context menu. However it is one config for every tab. So yes you need to do "menu switching" for every tab switching.

(We are doing the same in ScriptCat)

> Code Review

You should learn how to use AI to review the code...

---

@117649 

Suggest you need to test with 
1) the userscript with Iframe (so same script will execute twice):
  a) Top frame, script01.user.js register "menu option 1", "menu option 2"
  b) Iframe, script01.user.js register "menu option 1", "menu option 2"
2) also the case if two userscripts register the same menu option name.
  a) Top frame, script01.user.js register "menu option 1", "menu option 2"
  b) Top frame, script02.user.js register "menu option 1", "menu option 2"

You can compare the behavior with Tampermonkey

In ScriptCat we used the complicated behavior way to make compatible with Tampermonkey behavior, but I don't think that should be considered in Violentmonkey.

Reference UserScript Example: https://github.com/scriptscat/scriptcat/pull/1366/changes/2e4b035f5203a5f8907e69ef03fa5a73c71c75ff


---

Indeed the menu command feature is complicated...

tophf:
@cyfung1031, you seem to react to an outdated comment. The proper implementation is not particularly complicated, just wasteful in Chrome as it doesn't have browser.menus.onShown so personally I'd prefer to wait until it's added, which is why I'm so lazy and reluctant about this. As for AI review, I don't see the need, this code is not that big and if it was big I wouldn't trust it anyway. The per-frame menu separation should be taken in account, this is natural, not sure why you think it's not necessary in Violentmonkey.

117649:
>  Chrome as it doesn't have browser.menus.onShown so personally I'd prefer to wait until it's added,

Google(do all evil)'s way of say no way. ?

tophf:
Over the past 5 years they've added a number of features that were previously Firefox-only, most recently structured clone algo in messaging and support for returning a Promise from onMessage listener.

117649:
and mv3.

117649:
> Yep sorry, I'm still lazy because I'd have to add an option as discussed above, so if you could implement it, it'be harder for me to procrastinate further.

@tophf

tophf:
Yeah, thanks, I guess I can no longer postpone this, lol.

tophf:
After making some trivial changes to better match our architecture I noticed some redundancies and ended up refactoring more than I intended, but less than I thought initially, so not a big deal.

PR #2494, opened 2026-04-13
GitHub Issue\Feature Forms

WirlyWirly:
This simple pull replaces the older `bug-report.md` templates with the newer [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)

In the `config.yml`, I have left blank_issues enabled, as I wasn't sure if that is something you would like to continue to allow everyone access to.


tophf:
@gera2ld, WDYT? You can see the rendered preview in the diff tab -> `...` menu -> View file. Personally I like the idea in general but I dislike the emoji and the changed order (steps should come first as they were), the last confirmation checkbox seems redundant and it should be moved to the top to replace our old reminder to search for similar issues.

gera2ld:
LGTM. @WirlyWirly could you adjust the order as tophf suggested?

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,2 @@
+# Settings this options to 'false' will make it so that only maintainers can create blank issues
+blank_issues_enabled: true
```
cyfung1031, line 2: Creating a `config.yml` to just specify `blank_issues_enabled: true` is not necessary. It is because `blank_issues_enabled: true` is default behavior in Github.

Image

PR #2521, opened 2026-05-12
feat: add S3-compatible sync provider

Arlind-dev:
# Summary

Adds a new sync provider for S3-compatible object storage (AWS S3, Hetzner, Cloudflare R2, MinIO, etc.) following the same pattern as the existing WebDAV provider.

## Changes

- New `src/background/sync/s3.js` with AWS Signature Version 4 signing via `crypto.subtle`
- S3 constants added to `src/common/consts-sync.js`
- Settings UI fieldset in `vm-sync.vue` (Bucket, Region, Access Key ID, Secret Access Key, Endpoint, Key Prefix)
- `stateMap[SYNC_UNAUTHORIZED]` now also allows transitioning to `SYNC_INITIALIZING` because without it, saving S3 credentials while the state is unauthorized would silently fail since the state machine only allowed `SYNC_AUTHORIZING` as the next state (intended for OAuth flows)

## Notes

- Path-style URLs only (`https://endpoint/bucket/key`), works universally with S3-compatible services
- Endpoint accepts URLs with or without scheme, `https://` is prepended if missing
- `delimiter: /` used in list queries, so objects outside the configured prefix aren't picked up
- XML namespace fallback for services that omit the AWS namespace in ListObjectsV2 responses

## Testing
Tested on Firefox with Hetzner Object Storage and self-hosted Garage.


src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+    [...new TextEncoder().encode(c)]
+      .map(b => `%${b.toString(16).toUpperCase().padStart(2, '0')}`)
+      .join(''),
+  );
+}
+
+async function sha256hex(data) {
+  const buf = typeof data === 'string' ? new TextEncoder().encode(data) : data;
+  const hash = await crypto.subtle.digest('SHA-256', buf);
+  return [...new Uint8Array(hash)].map(b => b.toString(16).padStart(2, '0')).join('');
```
tophf: `toHex` can be reused here?

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+  [ACCESS_KEY_ID]: '',
+  [SECRET_ACCESS_KEY]: '',
+  [S3_PREFIX]: '',
+};
+
+// Per Sig V4: only unreserved chars (RFC 3986) are left unencoded in path segments.
+function encodeSegment(s) {
+  return s.replace(/[^A-Za-z0-9\-._~]/g, (c) =>
+    [...new TextEncoder().encode(c)]
+      .map(b => `%${b.toString(16).toUpperCase().padStart(2, '0')}`)
```
tophf: Why not simply `encodeURIComponent(s)`? Alternatively `toHex` can be reused here because percent-encoding allows lowercase too.

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+      .join(''),
+  );
+}
+
+async function sha256hex(data) {
+  const buf = typeof data === 'string' ? new TextEncoder().encode(data) : data;
```
tophf: `new TextEncoder().encode()` can be reused as a const function?

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+async function hmacSha256(key, data) {
+  const keyBuf = typeof key === 'string' ? new TextEncoder().encode(key) : key;
+  const k = await crypto.subtle.importKey(
+    'raw', keyBuf, { name: 'HMAC', hash: 'SHA-256' }, false, ['sign'],
+  );
+  return new Uint8Array(await crypto.subtle.sign('HMAC', k, new TextEncoder().encode(data)));
+}
+
+function toHex(buf) {
+  return [...buf].map(b => b.toString(16).padStart(2, '0')).join('');
```
tophf: Two arrays are created here. `Array.from()` would create just one array.

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+  region,
+}) {
+  const now = new Date();
+  const amzDate = now.toISOString().replace(/[:-]|\.\d{3}/g, '');
+  const datestamp = amzDate.slice(0, 8);
+
+  const canonicalUri = '/' + pathSegments.map(encodeSegment).join('/');
+  const sortedParams = Object.entries(queryParams)
+    .sort(([a], [b]) => a < b ? -1 : a > b ? 1 : 0);
+  const canonicalQS = sortedParams
```
tophf, line 82: Can this be replaced with `'' + new URLSearchParams(sortedParams)`?

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+  function elsByTag(parent, tag) {
+    const r = [...parent.getElementsByTagNameNS(ns, tag)];
+    return r.length ? r : [...parent.getElementsByTagName(tag)];
+  }
+  function elByTag(parent, tag) {
+    return parent.getElementsByTagNameNS(ns, tag)[0]
+      || parent.getElementsByTagName(tag)[0];
+  }
+
+  const items = elsByTag(doc, 'Contents').flatMap((el) => {
```
tophf: flatMap isn't supported in our minimum supported browser versions. You can use `.reduce` or `.map().filter(Boolean)`

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+    const key = elByTag(el, 'Key')?.textContent || '';
+    const name = key.split('/').pop();
+    if (!isScriptFile(name)) return [];
+    const size = +(elByTag(el, 'Size')?.textContent || 0);
+    return [normalize({ name, size })];
+  });
+
+  const nextToken = elByTag(doc, 'NextContinuationToken')?.textContent || null;
```
tophf: An explicit `|| null` is not necessary.

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+      secretKey: this._secretAccessKey,
+      region: this._region,
+    });
+    return this.loadData({ method, url, headers, body });
+  },
+
+  async list() {
+    const listPrefix = this._prefix ? `${this._prefix}/` : '';
+    const items = [];
+    let continuationToken = null;
```
tophf: An explicit `null` initializer is not necessary because it misleadingly implies this is important.

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+    this._prefix = c[S3_PREFIX]?.trim() ?? '';
+    return true;
+  },
+  async requestAuth() {
+    await this.list();
+  },
+  metaError(res) {
+    if (res.status !== 404) throw res;
+  },
+
```
tophf: Is the empty line a meaningful separator?

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+      endpoint: this._endpoint,
+      pathSegments,
+      body,
+      accessKeyId: this._accessKeyId,
+      secretKey: this._secretAccessKey,
+      region: this._region,
+    });
+    return this.loadData({ method, url, headers, body });
+  },
+
```
tophf: Is the empty line a meaningful separator?

src/background/sync/s3.js
```diff
@@ -0,0 +1,244 @@
...
+    const name = getItemFilename(item);
+    return this._s3('PUT', name, data);
+  },
+  remove(item) {
+    const name = getItemFilename(item);
+    return this._s3('DELETE', name);
+  },
+});
+
+register(S3);
```
tophf: The explicit initialization should be either at the very end or after the const/let block at the top.

src/options/views/tab-settings/vm-sync.vue
```diff
@@ -107,6 +107,72 @@
         />
       </div>
     </fieldset>
+    <fieldset v-if="rService && rAuthType === S3_AUTH" class="mt-1c">
+      <div class="mr-2c">
+        <label class="inline-block">
```
tophf, line 112: Might be more readable if you extract the data and use `v-for` with optional conditions inside e.g. https://github.com/violentmonkey/violentmonkey/blob/6a63e3db253b5fd2fb2e6c66f6fb72fccb7ab361/src/options/views/edit/settings.vue#L81

tophf (commented):
This is great, but I found some trivial nits I'd like to address either myself later after merging or have you do it if you find them reasonable. As for overall logic/appearance, let's wait for @gera2ld.

src/background/sync/s3.js
Arlind-dev: Done, `sha256hex` now uses `toHex`.

src/background/sync/s3.js
Arlind-dev: Can't use it, `encodeURIComponent` leaves `!` `'` `(` `)` `*` unencoded, but Sig V4 requires those to be percent-encoded too. Used `toHex` to simplify the implementation instead.

src/background/sync/s3.js
Arlind-dev: True, extracted as `const encode = new TextEncoder()` and reused throughout.

src/background/sync/s3.js
Arlind-dev: Done.

src/background/sync/s3.js
Arlind-dev: Didn't know! Fixed, using `.map().filter(Boolean)` instead.

src/background/sync/s3.js
Arlind-dev: Removed.

src/background/sync/s3.js
Arlind-dev: Removed.

src/background/sync/s3.js
Arlind-dev: Removed.

src/background/sync/s3.js
Arlind-dev: Removed.

src/background/sync/s3.js
Arlind-dev: Understood, moved to end.

src/options/views/tab-settings/vm-sync.vue
Arlind-dev: Done, extracted `S3_FIELDS` and used `v-for`.

src/background/sync/s3.js
Arlind-dev: Same issue as `encodeURIComponent`: 
`URLSearchParams` encodes spaces as + instead of `%20` and encodes `~` which Sig V4 requires to stay unencoded.

Arlind-dev:
Hope I could address all issues with the latest commit. 

I have tested the new build again, this time only with my selfhosted garage.

src/background/sync/s3.js
tophf: What about encodeURIComponent + a simple regex for these characters so there's no need for TextEncoder?

Arlind-dev: Done.

src/options/views/tab-settings/vm-sync.vue
```diff
@@ -153,6 +198,12 @@ import {
 } from '@/background/sync/state-machine';
 
 const SYNC_CURRENT = 'sync.current';
+const S3_FIELDS = [
+  { key: ACCESS_KEY_ID, label: 'labelSyncS3AccessKeyId', type: 'text' },
+  { key: SECRET_ACCESS_KEY, label: 'labelSyncS3SecretAccessKey', type: 'password' },
+  { key: S3_ENDPOINT, label: 'labelSyncS3Endpoint', type: 'url', placeholder: 'https://s3.example.com' },
+  { key: S3_PREFIX, label: 'labelSyncS3Prefix', type: 'text', placeholder: 'Violentmonkey' },
```
gera2ld: The placeholders are confusing because they look like the default value but are actually not. For example, if I leave the prefix empty, I see "Violentmonkey" in the input, but the actual value is empty. I recommend simply removing the placeholder.

src/options/views/tab-settings/vm-sync.vue
```diff
@@ -107,6 +107,44 @@
...
+            :disabled="!rCanUpdateConfig"
+          />
+        </label>
+        <label class="inline-block">
+          <span v-text="i18n('labelSyncS3Region')"></span>
+          <input
+            type="text"
+            v-model="rUserConfig[REGION]"
+            :disabled="!rCanUpdateConfig"
+            placeholder="us-east-1"
```
gera2ld: Same here, we should either make the placeholder as the default value, or remove the placeholder.

src/background/sync/state-machine.js
```diff
@@ -13,7 +13,7 @@ export const SYNC_ERROR = 9;
 const stateMap = {
   [SYNC_IDLE]: [SYNC_INITIALIZING, SYNC_UNAUTHORIZED],
   [SYNC_INITIALIZING]: [SYNC_AUTHORIZED, SYNC_ERROR_INIT, SYNC_UNAUTHORIZED],
-  [SYNC_UNAUTHORIZED]: [SYNC_AUTHORIZING],
+  [SYNC_UNAUTHORIZED]: [SYNC_AUTHORIZING, SYNC_INITIALIZING],
```
gera2ld, line 16: ~~This change doesn't look right. But it may be related to a bug where the auth status doesn't change after the config is updated. I'll check it later.~~

Never mind. This is indeed a bug.

gera2ld (commented):
It works for me. Thank you!

gera2ld (approved):

src/options/views/tab-settings/vm-sync.vue
Arlind-dev: Makes sense!

PR #2537, opened 2026-05-28
chore: switch to pnpm

tophf:
@gera2ld, WDYT? pnpm looks much faster than yarn, just 4 seconds to install the deps vs yarn's 22.


gera2ld:
I like that. We already use pnpm for other projects.
The main concern is how to ensure AMO reviewers strictly follow our workflow. Perhaps we should enable `corepack` and throw an error if the user is not using pnpm or is using a different version.

tophf:
I've added corepack [like this](https://github.com/violentmonkey/violentmonkey/commit/02c0ff4d68ae557a344f4f8fe8f73d43afd074dc#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519), please check.

tophf:
I wonder if AMO note is correct: `corepack enable pnpm && pnpm i && pnpm run ci && pnpm build`?

gera2ld:
Maybe prepend `npm i corepack@latest -g`, since I'm not sure whether the AMO reviewers are familiar with Node.js.
Otherwise LGTM.

BTW, can we update `engines.node` to `>= 24` now that Node 20 has reached its EOL?

cyfung1031:
in `pnpm-workspace.yaml`, you can also add `minimumReleaseAge` to reduce the risk of supply chain attack

AlttiRi:
Is not more suited for CI to use `ci` (not `i`)?

https://pnpm.io/cli/ci
https://docs.npmjs.com/cli/v11/commands/npm-ci

It's faster and builds will be byte-reproducible.

---

UPD. BTW, [`minimumReleaseAge`](https://pnpm.io/settings#minimumreleaseage) is already enabled by default.

tophf:
Theoretically yes and we'll use it when we migrate to v11 where this command was added. Last time I tried v11 it was complaining about some packages and I didn't investigate.

cyfung1031:
> Theoretically yes and we'll use it when we migrate to v11 where this command was added. Last time I tried v11 it was complaining about some packages and I didn't investigate.

it should be `pnpm install --frozen-lockfile` not `pnpm i`.
Not necessary to use `pnpm ci`

tophf:
Theoretically yes but I'm not convinced we really really need to use this switch until I see at least once a messed up lock file because AFAIK it's not something that can happen by itself.

PR #2547, opened 2026-06-10
Switch to JSZipp

cyfung1031:
* Follow-up for #1586.
* Compact and Fast like `fflate`
* You don't need to perform the Babel transform yourself like `zip.js`.
* You don't need to apply the DOS modification time patch yourself.
* ZIP64 support allows more than 65,535 files.
* Use the `cr61ff58` variant for Violentmonkey’s minimum browser requirements.
* Use NTFS timestamps (`modifiedAt`, `createdAt`, `lastAccess`) for compatibility with `zip.js`.

repo: https://github.com/jszipp/jszipp
benchmark: https://jszipp.github.io/JSZipp/demo/library-benchmark.html
docs: https://jszipp.github.io/JSZipp/

---

I used [CI](https://github.com/cyfung1031/violentmonkey/actions/runs/27246963932) to build and tried in Chrome. Works as expected.
I also used Waterfox Classic to test the CI build regarding the import and export - works as expected.


src/options/views/tab-settings/vm-export.vue
```diff
@@ -120,13 +120,10 @@ async function exportData() {
       vmValues[uri] = v;
     }
     vmScripts[name] = info;
-    files[`${name}.user.js`] = [
-      strToU8(code),
-      modDate && { mtime: new Date(modDate) },
-    ];
+    await writer.add({ path: `${name}.user.js`, data: `${code}`, meta: { modifiedAt: new Date(modDate) } });
```
tophf, line 85: Why adding is async? It should just accumulate the descriptors internally.

src/options/views/tab-settings/vm-export.vue
```diff
@@ -120,13 +120,10 @@ async function exportData() {
...
   }
-  files[vmZipEntryName] = [strToU8(JSON.stringify(vm, null, 2))]; // prettify to help users diff or view it
-  return new Blob([zipSync(files)], { type: 'application/zip' });
+  await writer.add({ path: `${vmZipEntryName}`, data: `${JSON.stringify(vm, null, 2)}` }); // prettify to help users diff or view it
```
tophf, line 87: same as above

src/options/views/tab-settings/vm-import.vue
```diff
@@ -174,11 +182,11 @@ async function doImportBackup(buf, zipName) {
         props: {
           lastModified: more.lastModified
             || more.props?.lastModified // Import data from Tampermonkey
-            || (decodedTime = +zipTimeToDate(time) || now),
+            || (decodedTime = +time || now),
```
tophf: Since there's no decoding now, the variable declaration can be removed and the code becomes `|| +time || now,`

src/options/views/tab-settings/vm-export.vue
```diff
@@ -28,7 +28,7 @@
 <script setup>
 import { computed, ref } from 'vue';
 import Modal from 'vueleton/lib/modal';
-import { strToU8, zipSync } from 'fflate';
+import { ZipWriter, TimestampMode } from 'web-jszipp/browser-legacy/cr61ff58';
```
tophf: There should be no need for you to maintain an additional distribution if you use `global` in your source code then your build config should automatically handle it in the standard UMD wrapper where it decides how to handle exports.

tophf (commented):
Thanks! I have a few questions though.

src/options/views/tab-settings/vm-export.vue
cyfung1031: The compression work is performed in `.add(...)`.

The implementation breaks the compression into smaller units of work instead of running it as one long CPU-bound task. After each `writer.add(...)`, execution yields back to the browser event loop, allowing the browser to handle higher-priority tasks such as rendering or user input before continuing.

---

For fflate, it just do the entire compression at the end. This is bad.

src/options/views/tab-settings/vm-export.vue
cyfung1031: same as above

src/options/views/tab-settings/vm-import.vue
cyfung1031: changed in dd8079e8fb6e92b1b4c32467800432a2c48700b8

src/options/views/tab-settings/vm-export.vue
cyfung1031: this is not UMD or ESM problem. Both modern and cr61ff58 have `.mjs` and `umd.js`
`web-jszipp/browser-legacy/cr61ff58` added the polyfill and changed the es version to es2015 in order to support old browser. 

### modern (main)
Screenshot 2026-06-10 at 17 40 22

### cr61ff58
Screenshot 2026-06-10 at 17 40 52

cyfung1031: just like `typeof AbortSignal !== "undefined" ? AbortSignal : XXXXXXXXXXXXXX`

the whole `XXXXXXXXXXXXXX` is dead code for modern build. It is just useless for the major browsers.

src/options/views/tab-settings/vm-export.vue
tophf: Since your library as a worker, let's use it.

src/options/views/tab-settings/vm-export.vue
tophf: I see but I hate polyfills conceptually so maybe we'll just bump Firefox's minimum from 58 to 65 or better yet 68, as well as chrome's. @gera2ld, WDYT?

gera2ld: Yeah, let's bump

tophf: Ok, so I'm bumping Chrome to 86 (Kafan uses 87) and Firefox to 68 (MyPal68).

tophf: @cyfung1031, regarding your cr86 distro, ReadableStream polyfill shouldn't be necessary because it's supported in the default non-BYOB mode. Polyfills for WritableStream, pipeTo or pipeThrough are necessary though.

cyfung1031: Firefox 68 does support basic non-BYOB `ReadableStream`, and native user-provided streams are still accepted through `isReadableStream_`.

`ReadableStream_` stays in the compat seam because the compat build needs one consistent stream family. Firefox 68 lacks `TransformStream`/`WritableStream`, so those have to be ponyfilled. Mixing a native `ReadableStream` with ponyfilled `TransformStream`/`WritableStream` can fail `pipeThrough`/`pipeTo` brand checks.

Using `ReadableStream_`, `WritableStream_`, `TransformStream_`, and `DecompressionStream_` from the same ponyfill family keeps the compat stream graph interoperable.

---

pipeThrough is used in the coding. so making ReadableStream polyfill makes sense.

Anyway I am not using VM now. If you don't want a polyfilled library, then just close this PR.

I made the polyfill because you said you need to support old browsers.

cyfung1031: changed from cr61ff58 to cr86ff68

https://github.com/violentmonkey/violentmonkey/pull/2547/commits/0fc6730648b3c8806de42c187fc633a32aee13e6

Issue #2227, opened 2024-11-08
[BUG] "Sync script status" issue


eight04:
![image](https://github.com/user-attachments/assets/e20d76b1-1239-4db4-9495-3257185fd52c)

### Sequence of actions: 
1. On PC, check "Sync script status" and sync to Google Drive.
2. On phone, uncheck "Sync script status" and sync to Google Drive.
3. On phone, disable some scripts that are not useful on mobile.

### Problem: 

Those scripts disabled on phone are also disabled on PC.

### Expected result:  

I guess the "Sync script status" option only works when pulling data from the cloud?

Ideally, I would expect that clients with this option can share script status. While clients without this option won't affect others.

#### Environment:
- OS: Windows 10, Android 14
- Browser: Firefox Nightly 134.0a1, Firefox Beta 133.0b4
- Violentmonkey Version: 2.28.0

berkaykrc:
I have the same problem. I have two PCs. One is old, so some scripts cannot be used there, or it causes performance problems. When I enable it on my desktop, it also enables it on my laptop, which it should not.

PR #2548, opened 2026-06-11
Refactor sync service and fix sync of enabled status

gera2ld:
## Changes

- Migrated OAuth2 and file providers to external packages (see https://github.com/usync-dev/usync)
- Added a link for managing remote files
- Fixed issue #2227


tophf:
Looks cool. For future MV3 work I wonder if this webdav implementation allows DOM processing to be called asynchronously so it can be done in an offscreen document via messaging.

gera2ld:
The [new package](https://github.com/usync-dev/usync/blob/main/packages/drive/src/providers/webdav.ts) is using `fast-xml-parser` which seems to work in a service worker.

tophf:
wow it's 55kb of minified code, not sure webdav is that complicated...

gera2ld:
You're right, 55KB is quite a hit for just WebDAV. The size comes from its standalone XML parsing engine.
Let me make the XML parser pluggable.

tophf:
Well duh, I ended up pulling the entire Drive into the offscreen document because the parser produces DOM elements which can't be sent via messaging. It works but I had to set up hackish relays for list() and authorizer. Hopefully you can take a look and find a better alternative: 
https://github.com/violentmonkey/violentmonkey/blob/acf88d4c98339ecc0055659891685d7043c8ab62/src/background/sync/sync-engine.js#L24-L40
https://github.com/violentmonkey/violentmonkey/blob/acf88d4c98339ecc0055659891685d7043c8ab62/src/background/sync/sync-engine.js#L473-L484
https://github.com/violentmonkey/violentmonkey/blob/acf88d4c98339ecc0055659891685d7043c8ab62/src/offscreen/index.js#L10-L37

PR #2552, opened 2026-06-18
add native browser downloads api support

### LINKED ISSUE #1349: Add support for setting path in GM_download
Hi, Why don't you add the ability to set optional "managing downloads" permission and the functionality to support setting path in GM_download like it is in TM?
image

Tyarel8:
Added the option in advanced settings to have `GM_download` use the [browser download api](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download).

You can pass the arguments to `GM_download` and they will be forwarded to the browser downloads api. The previous behavior with XML http request should be unchanged, but now the extension also needs an optional `downloads` permission.

I have tested that it works on firefox (152.0).

Closes #1349.

### DIFF
--- a/_locales/en/messages.yml
+++ b/_locales/en/messages.yml
@@ -701,6 +701,19 @@ labelXhrInjectHint:
 labelXhrInjectNote:
   description: Shown to the right of "synchronous page mode" in options.
   message: (except incognito and sites with disabled cookies)
+labelGmDownloadModeBrowser:
+  description: Option in Advanced settings to use the browser download API for GM_download.
+  message: Use browser download API
+labelGmDownloadModeBrowserHint:
+  description: Tooltip for the browser download option in Advanced settings.
+  message: Allows scripts to save files into subfolders inside your downloads
+    directory and shows download progress in the browser's download bar.
+labelGmDownloadModeBrowserPermissionRequired:
+  description: Shown after the browser download option label.
+  message: (requires "downloads" permission)
+labelGmDownloadModeBrowserGrantPerm:
+  description: Button to request the downloads permission.
+  message: Grant permission
 lastSync:
   description: Label for last sync timestamp.
   message: Last sync at $1
--- a/src/background/utils/preinject.js
+++ b/src/background/utils/preinject.js
@@ -190,6 +190,8 @@ function setBaggedScriptValues(bag, id, val) {
   }
 }
 
+
+
 function clearFrameData(tabId, frameId, tabRemoved) {
   clearRequestsByTabId(tabId, frameId);
   clearValueOpener(tabId, frameId);
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -10,6 +10,7 @@ import {
   FORBIDDEN_HEADER_RE, kCookie, kSetCookie, requests, toggleHeaderInjector, verify,
 } from './requests-core';
 import { getFrameDocIdAsObj, getFrameDocIdFromSrc } from './tabs';
+import { getOption } from './options';
 import { navUA, navUAD } from './ua';
 import { vetUrl } from './url';
 
@@ -49,6 +50,27 @@ addPublicCommands({
       __.MV3 && req.url && res.data && (res.data.finalUrl = req.url), // from onBeforeSendHeaders
       sendTabCmd(tabId, 'HttpRequested', res, req.frame)
     );
+    if (getOption('gmDownloadModeBrowser') && opts[kFileName]) {
+      const downloadOpts = { url: opts.url, filename: opts[kFileName] };
+      if (opts.headers) {
+        downloadOpts.headers = Object.entries(opts.headers).map(
+          ([n, v]) => ({ name: n, value: v }),
+        );
+      }
+      if (opts.saveAs != null) downloadOpts.saveAs = opts.saveAs;
+      if (opts.conflictAction) downloadOpts.conflictAction = opts.conflictAction;
+      if (opts.method) downloadOpts.method = opts.method;
+      browser.downloads.download(downloadOpts)
+        .then(() => {
+          cb({ id, type: 'load', data: { finalUrl: opts.url, readyState: 4, status: 200, response: null, responseHeaders: null } });
+          cb({ id, type: 'loadend', data: null });
+        })
+        .catch(err => {
+          cb({ id, [ERROR]: [err.message || `${err}`, 'DownloadError'], data: null });
+          cb({ id, type: 'loadend', data: null });
+        });
+      return;
+    }
     return httpRequest(opts, events, src, cb)
     .catch(err => cb({
       id,
--- a/src/injected/web/gm-api.js
+++ b/src/injected/web/gm-api.js
@@ -71,10 +71,10 @@ export const GM_API_CTX_GM4ASYNC = {
     }
     assign(opts, {
       [kResponseType]: 'blob',
-      data: null,
-      method: 'GET',
       overrideMimeType: 'application/octet-stream',
     });
+    if (opts.data == null) opts.data = null;
+    if (opts.method == null) opts.method = 'GET';
     return onRequestCreate(opts, this, name);
   },
 };
--- a/src/options/views/tab-settings/index.vue
+++ b/src/options/views/tab-settings/index.vue
@@ -82,6 +82,16 @@
           </locale-group>
         </div>
         <setting-check name="helpForLocalFile" :label="i18n('helpForLocalFile')"/>
+        <div class="ml-2c">
+          <tooltip :content="i18n('labelGmDownloadModeBrowserHint')">
+            <setting-check name="gmDownloadModeBrowser">
+              {{ i18n('labelGmDownloadModeBrowser') }} <ruby v-text="i18n('labelGmDownloadModeBrowserPermissionRequired')" />
+            </setting-check>
+          </tooltip>
+          <button class="btn-ghost" :disabled="permGranted"
+                  @click="requestDlPerm"
+                  v-text="i18n('labelGmDownloadModeBrowserGrantPerm')" />
+        </div>
       </section>
 
       <section>
@@ -136,6 +146,7 @@ const items = {
     light: i18n('optionUiThemeLight'),
   },
   xhrInject: value => value,
+  gmDownloadModeBrowser: value => value,
 };
 </script>
 
@@ -154,6 +165,13 @@ import VmEditor from './vm-editor';
 import VmBlacklist from './vm-blacklist';
 import VmDateInfo from './vm-date-info';
 import { kbdTypable } from '@/common/keyboard';
+import browser from '@/common/browser';
+
+const permGranted = ref(false);
+
+async function requestDlPerm() {
+  permGranted.value = await browser.permissions.request({ permissions: ['downloads'] });
+}
 
 const $el = ref();
 const settings = reactive({});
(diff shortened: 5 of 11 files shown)


_locales/en/messages.yml
```diff
@@ -701,6 +701,14 @@ labelXhrInjectHint:
 labelXhrInjectNote:
   description: Shown to the right of "synchronous page mode" in options.
   message: (except incognito and sites with disabled cookies)
+labelGmDownloadNative:
+  description: Option in Advanced settings to use the browser's native download API instead of the DOM-based download for GM_download.
+  message: Use native browser download API
+labelGmDownloadNativeHint:
+  description: Tooltip for the native download option in Advanced settings.
+  message: Skip XMLHttpRequest and use the browser's native download manager for
```
tophf: The description shouldn't mention technicalities for script developer but rather it should explain to the user that enabling this option allows scripts to download inside subfolders in their downloads directory, which is something that need to be aware of as it's not the usual behavior of downloads.

src/background/utils/preinject.js
```diff
@@ -441,7 +441,7 @@ async function prepareBag(cacheKey, url, isTop, env, inject, errors) {
     [MORE]: moreKey,
     [kSessionId]: sessionId,
     [IDS]: allIds,
-    info: { ua },
+    info: { ua, gmDownloadNative: getOption('gmDownloadNative') },
```
tophf: The mode should be exposed in a TM-compatible fashion as `downloadMode` because their solution makes sense, https://www.tampermonkey.net/documentation.php?q=GM_info

src/injected/content/requests.js
```diff
@@ -82,6 +82,7 @@ addHandlers({
     return sendCmd('HttpRequest', msg);
   },
   AbortRequest: true,
+  NativeDownload: true,
```
tophf: `true` should be `REIFY`, it's a magic constant so no need to import it.

src/injected/web/gm-api.js
```diff
@@ -69,6 +69,28 @@ export const GM_API_CTX_GM4ASYNC = {
       onRequestInitError(opts, new SafeError(`Required parameter "name" is ${name}.`));
       return;
     }
+    const useNative = opts.nativeDownload != null ? opts.nativeDownload : bridge.gmDownloadNative;
```
tophf: This entire code should be moved into the `background` context because the `web` is unsafe and everything including a Object.entries may be compromised. The decision to use native mode will be made by the presence of `saveAs`.

tophf (changes requested):
This is a very good start but we can't just add a new permission that has an installation warning because it disables the extension for everyone on update, so it should be in `optional_permissions` and the option should be only enabled if the permission is granted as explained in the documentation: https://developer.chrome.com/docs/extensions/reference/permissions

It'd be also nice to imitate the other events fired by GM_download/GM_xmlhttpRequest based on chrome.downloads API events, the progress events in particular: https://developer.chrome.com/docs/extensions/reference/downloads#event-onChanged, although this might be something we could do later.

src/injected/web/gm-api.js
Tyarel8: Wouldn't a explicit arg to use either mode make more sense (and go by config setting if not present)? Even if you want to use the browser api, you may not want to specify `saveAs` and just use whatever your browser defaults to.

tophf: My concern is compatibility/interoperability. Currently, judging by the documentation for TM, userscripts don't use such a separate parameter and knowing how TM's author loves adding various parameters it seems such a parameter was either unnecessary or it's conceptually wrong for a userscript to control this explicitly.

tophf: Now I remembered that in TM it's the user who controls this behavior and it actually makes sense:
image

Tyarel8: So should I remove the arg and just have it in the settings only?

tophf: Yes, I think so.

src/injected/web/gm-api.js
```diff
@@ -69,6 +69,31 @@ export const GM_API_CTX_GM4ASYNC = {
...
+          } else {
+            if (opts.onload) opts.onload(res);
+            resolve(res);
+          }
+        });
+      });
+      if (this.async) return doDownload();
+      doDownload().catch(err => logging.error(err));
+      return;
+    }
```
tophf: Make use of the existing code and just add saveAs in https://github.com/violentmonkey/violentmonkey/blob/de18aba991960fe485caccbb6b58c9e6d68e5a45/src/injected/web/requests.js#L237-L247

Then check gmDownloadModeBrowser option in https://github.com/violentmonkey/violentmonkey/blob/de18aba991960fe485caccbb6b58c9e6d68e5a45/src/background/utils/requests.js#L25

src/options/views/tab-settings/index.vue
```diff
@@ -168,6 +188,7 @@ onActivated(() => {
     ...hookSettingsForUI(items, settings, watch, 50),
   ];
   expose.value = Object.keys(options.get(EXPOSE)).map(k => [k, decodeURIComponent(k)]);
+  browser.permissions.contains({ permissions: ['downloads'] }).then(r => { permGranted.value = r; });
```
tophf, line 189: I wonder if it might make sense to use [onAdded and onRemoved events](https://developer.chrome.com/docs/extensions/reference/api/permissions#event-onAdded) to update the UI if the user has revoked the permission although I don't think users would do that realistically so I guess those who did might also figure out to refresh the page.

src/injected/web/gm-api.js
Tyarel8: Alright, I have moved it to requests.js and added saveAs and conflictAction.

src/options/views/tab-settings/index.vue
Tyarel8: I think that's a bit of an unnecessary complexity, if someone permissions aren't working, they would refresh the page/restart the browser at some point I would think.

src/options/views/tab-settings/index.vue
```diff
@@ -172,7 +172,7 @@ const permGranted = ref(false);
 async function requestDlPerm() {
   try {
     permGranted.value = await browser.permissions.request({ permissions: ['downloads'] });
-  } catch {}
+  } catch {} // eslint-disable-line no-empty
```
tophf: We usually just put a comment like `/* ignore */` to confirm intent. Although I wonder why this API can fail in the first place?

Tyarel8: You're right, the try/catch is most likely unnecessary, I was trying a bunch of things to try and make it work with the checkbox for the setting to also trigger the permission request without much success.

src/background/utils/requests.js
```diff
@@ -32,6 +33,27 @@ addPublicCommands({
...
+        downloadOpts.headers = Object.entries(opts.headers).map(
+          ([n, v]) => ({ name: n, value: v }),
+        );
+      }
+      if (opts.saveAs != null) downloadOpts.saveAs = opts.saveAs;
+      if (opts.conflictAction) downloadOpts.conflictAction = opts.conflictAction;
+      if (opts.method) downloadOpts.method = opts.method;
+      browser.downloads.download(downloadOpts)
+        .then(() => {
+          cb({ id, type: 'load', data: { finalUrl: opts.url, readyState: 4, status: 200, response: null, responseHeaders: null } });
```
tophf, line 65: @gera2ld I wonder if we should modify normal GM_download to omit the response for consistency? It would simplify the code and improve performance. After all if the script wants data it can use GMxhr.

tophf (approved):

src/background/utils/requests.js
gera2ld: Sounds good to me

Tyarel8:
I just realize that currently data/blob urls are intercepted and don't go to the browser download api even if the setting is on. I looked into it and blob urls can have problems with scoping but there shouldn't be any reason to not route data urls through the browser download api right?

tophf:
Yeah but maybe this API can just download the original blob/data URL without copying.

src/injected/content/requests.js
```diff
@@ -16,8 +16,10 @@ let U8_set;
 let safeFetch;
 let navigator, getUAData, getUAProps, getHighEntropyValues;
 let SafeDOMParser, parseFromString;
+let gmDownloadModeBrowser;
 
 onScripts.push(data => {
+  gmDownloadModeBrowser = data.info?.gmDownloadModeBrowser;
```
tophf: `info` is always present when scripts are injected. BTW the user may change the option and I'd like the existing tabs to receive the new mode the way it's done with kUseMenu, but I can do it later.

Tyarel8: I didn't notice in testing before but it doesn't actually work (at least in firefox), it reaches the download api but nothing happens and I don't see any error, but found this in stackoverflow https://stackoverflow.com/questions/40269862/save-data-uri-as-file-using-downloads-download-api

Edit: found the bug in bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1318564

Maybe I should revert and leave data and blob downloads alone (at least for now)

Tyarel8: I'll revert it, data/blob urls can be looked at later.

cyfung1031:
@Tyarel8 
i did not check the code but just to remind that, 

since you are using `browser.downloads.download`,
please check the extension compatibility carefully.

1) install Motrix Next Extension
2) install https://greasyfork.org/scripts/561953
3) go to twitter to test whether the media (e.g. video in twitter) can be downloaded **with the correct customized naming**.

In my finding, if you want to use `browser.downloads.download`, `browser.downloads.onDeterminingFilename` must be called as well.
(Other extensions hooking `browser.downloads.onDeterminingFilename` can override the `filename` setting in your `browser.downloads.download`)
this is somehow a bug in Chromium. See https://issues.chromium.org/issues/40706258

for details, see https://greasyfork.org/scripts/561953/discussions/319273 and  https://github.com/scriptscat/scriptcat/issues/1409 (it is Chinese though)

Tyarel8:
> @Tyarel8 i did not check the code but just to remind that,
> 
> since you are using `browser.downloads.download`, please check the extension compatibility carefully.
> 
>     1. install Motrix Next Extension
> 
>     2. install [greasyfork.org/scripts/561953](https://greasyfork.org/scripts/561953)
> 
>     3. go to twitter to test whether the media (e.g. video in twitter) can be downloaded **with the correct customized naming**.
> 
> 
> In my finding, if you want to use `browser.downloads.download`, `browser.downloads.onDeterminingFilename` must be called as well. (Other extensions hooking `browser.downloads.onDeterminingFilename` can override the `filename` setting in your `browser.downloads.download`) this is somehow a bug in Chromium. See [issues.chromium.org/issues/40706258](https://issues.chromium.org/issues/40706258)
> 
> for details, see [greasyfork.org/scripts/561953/discussions/319273](https://greasyfork.org/scripts/561953/discussions/319273) and [scriptscat/scriptcat#1409](https://github.com/scriptscat/scriptcat/issues/1409) (it is Chinese though)

I have looked into it briefly and it seems that the chrome bug is marked as resolved, and `onDeterminingFilename` overriding the filename is working as intended, it's supposed to allow changing the filename of any browser api download.

Tyarel8:
I managed to rebase this to master and fix the merge errors after much difficulty (git sucks), @tophf can you merge it?

tophf:
I wanted to do it after the stable MV3 is published next week.

piknockyou:
I also hit this limitation. I have a YouTube userscript with a small inline UI for selecting audio/video/subtitle tracks — it can download two files at once: a Python script with ready-to-run yt-dlp/ffmpeg commands, and a cookie file with a fixed name that the Python script loads automatically. Without `conflictAction: "overwrite"`, successive exports produce numbered copies like `cookies (1).txt` and the companion script loads the wrong file.

This was noted in the review thread above — `data:` URIs were reverted with "can be looked at later" since Firefox didn't handle them via the downloads API at the time. The follow-up is tracked in #2605 — routing `data:` URIs through the Browser API on Chromium, where it works correctly (verified with Tampermonkey and with VM's own `http(s):` Browser API path).

PR #2570, opened 2026-07-07
fix: refactor sw reinitialization for undoImport and improve tab-open…

AgentHitmanFaris:
This PR resolves state reinitialization issues on undo-import, removes experimental serialization warnings in Chrome/Brave Stable channels, and improves tab-opening error visibility under Manifest V3 (MV3).

1. **In-Memory State Reinitialization (Undo Import without Extension Reload)**
   * Avoids restarting the entire extension via `chrome.runtime.reload()` when undoing an import.
   * Introduced `onPostUndo` event hooks in `storage-cache.js` to allow background modules to clear and re-initialize their states.
   * Refactored the main database initialization block in `db.js` into an exported `initializeDatabase` function that registers to `onPostUndo`, cleaning script maps/lists in memory and reloading storage.
   * Clears option properties in `options.js` during `initOptions` to prevent stale setting values from lingering after database reinitialization.
   * Reloads extension UIs smoothly using the existing `sendCmd('Reload')` channel.

2. **Manifest Warn Mitigation (Chrome/Brave Stable Compatibility)**
   * Avoids loading errors on Stable channel Chromium browsers (such as Brave Stable) by commenting out `message_serialization: 'structured_clone'` (which is currently restricted to Nightly/Canary channels in Chromium).
   * Restored fallback binary stringification logic in `requests.js` on stable channels to ensure compatibility.

3. **Tab-Opening Error Reporting**
   * Wrapped unhandled exceptions from `browser.tabs.create` inside `tabs.js`.
   * Triggers a user-facing desktop notification detailing the error (e.g., when opening invalid schemes or restricted protocols) instead of failing silently in the background script.

### Verification & Testing
* **CI Verification:** All 7 test suites (48 individual tests) pass successfully (`pnpm test`).
* **Build Status:** Build compiles cleanly under MV3 mode (`pnpm build-mv3`).

### DIFF
--- a/src/background/utils/db.js
+++ b/src/background/utils/db.js
@@ -11,7 +11,7 @@ import { deepSize, forEachEntry, forEachKey, forEachValue } from '@/common/objec
 import pluginEvents from '../plugin/events';
 import {
   aliveScripts, getDefaultCustom, getNameURI, inferScriptProps, newScript, parseMeta,
-  removedScripts, scriptMap,
+  removedScripts, scriptMap, scriptSiteVisited,
 } from './script';
 import { testBlacklist, testerBatch, testScript } from './tester';
 import { getImageData } from './icon';
@@ -119,7 +119,16 @@ addOwnCommands({
   Vacuum: vacuum,
 });
 
-(async () => {
+export async function initializeDatabase() {
+  maxScriptId = 0;
+  maxScriptPosition = 0;
+  dbKeys.clear();
+  aliveScripts.length = 0;
+  removedScripts.length = 0;
+  for (const key in scriptMap) delete scriptMap[key];
+  for (const key in scriptSizes) delete scriptSizes[key];
+  for (const key in scriptSiteVisited) delete scriptSiteVisited[key];
+
   /** @type {string[]} */
   let keys;
   let [allKeys, data] = await Promise.all([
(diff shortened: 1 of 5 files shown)


src/background/utils/storage-cache.js
```diff
@@ -242,8 +244,16 @@ async function undoImport(port) {
...
+      clearStorageCache();
+      for (const cb of onPostUndo) {
+        try {
+          await cb();
+        } catch (e) {
+          console.error(e);
+        }
+      }
+      undoing = false;
+      await sendCmd('Reload', delay);
```
tophf: This code can be used for MV2 path too so `Reload` command can be removed from here and from `handlers`.

src/background/utils/db.js
```diff
@@ -206,7 +215,12 @@ addOwnCommands({
     setInterval(checkRemove, TIMEOUT_24HOURS);
   }
   resolveInit();
-})();
+}
+
+initializeDatabase();
+if (__.MV3) {
```
tophf: You've already exported initializeDatabase so it seems better to call it directly from storage-cache without onPostUndo as it's the only caller.

tophf (changes requested):
Thanks, this is great, I was afraid it'd be much more involved. I'd like some minor changes though and I wonder if you could split the notification fix into a separate commit?

PR #2610, opened 2026-07-25
Move Drive from offscreen to service worker with custom XML parser

gera2ld:
This is a follow-up to #2548, addressing https://github.com/violentmonkey/violentmonkey/pull/2548#issuecomment-4870796717.

DOMParser is unavailable in service workers, so Drive had to run in the offscreen document. This required a proxy layer that routed every method call, async iterator, and authorizer interaction through the offscreen bridge — adding complexity and a failure point.

This PR replaces DOMParser with [@violentmonkey/xml-parser](https://github.com/violentmonkey/xml-parser), which works natively in the service worker. With that, Drive can run directly in the SW.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
     "@usync/oauth2": "^0.1.7",
     "@usync/sync": "^0.1.1",
     "@violentmonkey/shortcut": "^1.4.4",
+    "@violentmonkey/xml-parser": "^1.0.0",
     "codemirror": "^5.65.20",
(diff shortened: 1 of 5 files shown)


tophf (approved):
Great! The overall js size is 6kB smaller. One thing for the future I'd like to explore is decreasing the sw size by making sync a separate js file loaded via importScripts like tldjs because even though js compilation is cacheable now in Chrome it still takes time.

PR #2625, opened 2026-08-03
fix #2624: read a page's blob: URL in the content realm for GM_download

### LINKED ISSUE #2072: [BUG] blob url download with GM_download doesn't work
```js
// ==UserScript==
// @name         GM_download sample
// @author       Eai <eai@mizle.net>
// @license      MIT
// @version      1.0.0
// @match        https://example.com/
// @grant        GM_registerMenuCommand
// @grant        GM_download
// ==/UserScript==
/* global GM_registerMenuCommand */
/* global GM_download */

function main() {
    const sampleBlob = new Blob(["Hello, world!"], { type: "text/plain" });
    const url = window.URL.createObjectURL(sampleBlob);
    GM_download(url, "hello.txt");
}

(() => {
    GM_registerMenuCommand("Download", main);
})();
```

### Sequence of actions: 
1. Install the script and open example.com
2. Select Download from the menu

### Problem: 

File will not be downloaded.

For release version, nothing is displayed.
For beta version, an error will be output to console.

### Expected result:  

The file will be downloaded.

Works as expected with Tampermonkey in Firefox, Violentmonkey in Chrome.

#### Devtools console contents:
```
Uncaught TypeError: yn is not a function
    Pe moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/sandbox/injected-web.js:1
    post moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/sandbox/injected-web.js:1
    $e moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/sandbox/injected-web.js:1
    Ye moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/sandbox/injected-web.js:1
    GM_download moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/sandbox/injected-web.js:1
    main moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/ GM_download sample.user.js#1:16
    Command moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/sandbox/injected-web.js:1
    onHandle moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/sandbox/injected-web.js:1
    Le moz-extension://fef9b40d-d2bd-4e39-bf7f-b2e258cf5346/sandbox/injected-web.js:1
injected.js:formatted:1193:19
```

image

Perhaps yn means getTabUserAgent?

https://github.com/violentmonkey/violentmonkey/blob/8626cd9bb8fd2533ce6e221939d6633389cfc8cc/src/injected/content/requests.js#L53

#### Environment:
- OS: Windows
- Browser: Firefox 124.0.2 
- Violentmonkey Version: BETA v2.18.2

### LINKED ISSUE #2605: GM_download with data: URI doesn't route through Browser API
### Description

`data:` URIs passed to `GM_download` do not route through the Browser API even with the "Use browser download API" setting enabled (v2.45.2). This was explicitly deferred when Browser API was added in #2552 — this issue picks up that deferred work.

For `http(s):` URLs (real web addresses), the Browser API works correctly — `saveAs: true` opens a Save As dialog, `conflictAction: "overwrite"` replaces old files. For `data:` URIs (content generated by the script itself), the Browser API is bypassed — `saveAs`, `conflictAction` have no effect.

Working correctly in Tampermonkey. Partially addressed but not resolved in Violentmonkey.

### Related threads

| Link | Status | Relevance |
|---|---|---|
| [#2552](https://github.com/violentmonkey/violentmonkey/pull/2552) (PR) | **Merged** | Added Browser API for `http(s):` URLs; `data:` URIs explicitly excluded (see [comment](https://github.com/violentmonkey/violentmonkey/pull/2552#issuecomment-4755374545) and [follow-up](https://github.com/violentmonkey/violentmonkey/pull/2552#issuecomment-4756522230): *"I'll revert it, data/blob urls can be looked at later"*) |
| [#1349](https://github.com/violentmonkey/violentmonkey/issues/1349) | **Closed** (by #2552) | Original feature request for `saveAs`/paths. Closed as completed, but `data:` URIs were never handled |
| [#1849](https://github.com/violentmonkey/violentmonkey/issues/1849) | **Closed** (fixed) | Blob URL download bug — unrelated, fixed separately |

### Reproduction

```js
// ==UserScript==
// @name        GM_download data: URI test
// @match       *://*/*
// @grant       GM_download
// @run-at      document-idle
// ==/UserScript==

const csv = "a,b,c\n1,2,3";
const dataUrl = `data:text/csv;charset=utf-8,${encodeURIComponent(csv)}`;
GM_download({
  url: dataUrl,
  name: "test.csv",
  saveAs: true,
  onload: () => console.log("onload fired"),
  onerror: (e) => console.error("onerror:", e),
});
```

1. Enable "Use browser download API" in VM Advanced settings
2. Visit any site, run the script
3. `onload` fires but `saveAs` is ignored — no dialog appears

### Use case

I have a YouTube userscript with a small inline UI for selecting audio/video/subtitle tracks. When triggered, it can download two files at once:

- **A Python script** containing ready-to-run yt-dlp commands (with ffmpeg, mkvmerge, etc.)
- **A cookie file** with a fixed name that the Python script loads automatically

The cookie file must keep its exact filename — the Python script expects it by that name. Without `conflictAction: "overwrite"`, successive exports produce `cookies (1).txt`, `cookies (2).txt`. The Python script either won't recognize the numbered file or will load the previous (stale) cookie file that still has the original name. Overwriting is crucial for a reliable one-click workflow.

(There are also other cases where being able to `saveAs` generated content would be useful, but the cookie overwrite case is the concrete example.)

### Expected behavior

`data:` URIs passed to `GM_download` should route through `chrome.downloads.download` when the Browser API setting is enabled, just like `http(s):` URLs already do. This would enable `saveAs`, `conflictAction`, and deterministic filenames — completing the work deferred in #2552.

### LINKED ISSUE #2624: [BUG] Firefox: GM_download fails with status 0 for a blob: URL created in the page (regression of #2072)
### Sequence of actions:

1. Install this userscript:

```js
// ==UserScript==
// @name         blob: repro
// @match        https://example.com/*
// @grant        GM_download
// @version      1.0
// ==/UserScript==

const url = URL.createObjectURL(new Blob(["hello\n"], { type: "text/plain" }));
GM_download({
  url,
  name: "repro.txt",
  onload: () => console.log("OK"),
  onerror: e => console.error("FAIL:", e),
});
```

2. Open `https://example.com/`.
3. Read the devtools console.

### Problem:

`GM_download` saves no file. `onerror` runs instead.

This happens in Firefox only. Brave downloads the file correctly with the same script. I installed Violentmonkey (v2.46.0) in Brave on 2026-08-03, so that copy includes commit `db1b5c3e` below. The difference is therefore the browser, not the Violentmonkey version.

### Expected result:

Violentmonkey saves `repro.txt` and the console shows `OK`.

#### Devtools console contents:

```
FAIL: 
Object { finalUrl: "", readyState: 4, status: 0, statusText: "", lengthComputable: false, loaded: 0, total: 0, response: Getter, responseHeaders: "", responseText: null, … }
blob: repro.user.js:16:25
```

The same object, expanded:

```json
{
  "finalUrl": "",
  "readyState": 4,
  "status": 0,
  "statusText": "",
  "lengthComputable": false,
  "loaded": 0,
  "total": 0,
  "response": null,
  "responseHeaders": "",
  "responseText": null,
  "responseXML": null
}
```

My export script logs the same object for every file that it tries to save.

#### Environment:

- OS: MacOS Tahoe 26.5.2 / Windows 11 (no access to that machine right now)
- Browser: Firefox 153.0 fails. Brave 1.93.129 (Chromium 151.0.7922.71) works.
- Violentmonkey Version: 2.46.0

The error also occurs on Windows, with the same script.

### Suspected cause

This looks like a regression of https://github.com/violentmonkey/violentmonkey/issues/2072. That issue used almost the same repro script and reported the same split between browsers: "Works as expected with Tampermonkey in Firefox, Violentmonkey in Chrome."

Commit `b8f2f90c` ("fix #2072: request local blobs in tab", 2024-04-10) fixed it. That commit added the `data:`/`blob:` branch and the `requestVirtualUrl()` helper to `src/injected/content/requests.js`. The helper read the blob in the content realm, where a page-origin `blob:` URL is valid.

Commit `db1b5c3e` ("fix #2605: reuse bg path for data/blob urls in gmxhr/gmdl", 2026-07-22, in 2.45.3) removed that branch and that helper:

```js
const isDataUri = sch === 'data:';
if (isDataUri || sch === 'blob:') {
  return requestVirtualUrl(msg, url, isDataUri, realm);
}
```

The `blob:` URL now goes to the background instead. A `blob:` URL is scoped to the context that created it, so I expect that the background cannot read it. This expectation matches the `status: 0` and the empty `finalUrl`. Chromium is unaffected, as in https://github.com/violentmonkey/violentmonkey/issues/2072.

The helper was in place for more than 2 years. I did not bisect the releases, so I cannot confirm 2.45.3 as the first affected release.

Issue https://github.com/violentmonkey/violentmonkey/issues/2605 asked for `data:` URI routing. I did not test `data:` URIs. `vetUrl()` passes a `data:` URI through unchanged, and a `data:` URI carries its own payload, so my reading suggests `data:` still works. A revert of the whole commit will also revert that fix.

I test the "browser download API" option from https://github.com/violentmonkey/violentmonkey/pull/2552 as well. But it does not help here. `downloadViaApi()` passes `opts.url` to `browser.downloads.download` unchanged, from the same background context.

Auxority:
Fixes #2624.

## Problem

In Firefox, `GM_download` of a `blob:` URL created by the page always fails. The callback reports `status: 0`, `finalUrl: ""` and `responseText: null`, and no file reaches the disk.

```js
const url = URL.createObjectURL(new Blob(['hi'], { type: 'text/plain' }));
GM_download({ url, name: 'x.txt', onload: () => console.log('OK'), onerror: e => console.log('FAIL', e) });
```

## Cause

A `blob:` URL is valid only in the context that created it, so bg and offscreen cannot re-fetch it. b8f2f90c ("fix #2072: request local blobs in tab") read such URLs in the content realm instead. db1b5c3e ("fix #2605: reuse bg path for data/blob urls in gmxhr/gmdl") removed that code, which brought #2072 back for Firefox.

## Change

This PR restores the content-realm read, but only for a `blob:` URL that carries a `fileName`, which means `GM_download`.

`injected/content/requests.js` reads the blob with the existing `importBlob` and hands it to bg: the `Blob` itself in Firefox, a `createObjectURL` string elsewhere. It then posts `readystatechange`, `load` and `loadend`, so the script callbacks run. `response` stays `null`, which matches the browser-downloads-API mode that db1b5c3e aligned it with. If the read fails, the code posts an `error` event. Previously the request stalled with no event at all.

`background/utils/requests.js` and `offscreen/index.js` restore the `DownloadBlob` command, wired `__.MV3 ? callOffscreen : downloadBlob` next to `RevokeBlob`. The anchor click stays in bg, which circumvents CSP in Firefox and keeps the single throttled download chain.

**The scope is deliberate, not a full revert.** `data:` URIs and plain `GM_xmlhttpRequest` stay on the background path from db1b5c3e, so #2605 remains fixed. The shortcut is entered only for a URL that starts with `blob:` and has a `fileName`.

## Verification

No unit test: the project has no harness for `GM_download`, `GM_xmlhttpRequest` or the request path, and the jest setup covers neither the content realm nor the web realm.

Instead I tested this branch end to end in real browsers with real builds. geckodriver drove Firefox against the MV2 build (2.46.0) as a temporary add-on, and Playwright drove Chrome for Testing 151 against the unpacked `dist-mv3`. One userscript covers five cases.

| case | Firefox before | Firefox after | Chromium before | Chromium after |
|---|---|---|---|---|
| `GM_download` `blob:` (the bug) | **fails**, `status: 0` | OK, file saved | OK | OK |
| `GM_download` `https:` | OK | OK | OK | OK |
| `GM_download` `data:` (#2605) | OK | OK | OK | OK |
| `GM_xmlhttpRequest` GET | OK | OK | OK | OK |
| `GM_xmlhttpRequest` `blob:` | fails | fails, unchanged | OK | OK |

The tests read every downloaded file from disk and compare its bytes, rather than trusting `onload`. `pnpm lint` and `pnpm test` (48 tests) pass.

## Two related points

1. Plain `GM_xmlhttpRequest` on a page `blob:` URL fails in Firefox too, as the last table row shows. The cause is identical, and this PR does not change it. A wider fix must send the blob contents back to the web realm. db1b5c3e changed that behavior on purpose, so I left it alone rather than guess at the intended semantics.
2. `saveAs` and `conflictAction` no longer apply to a `blob:` URL, because the request never reaches `downloadViaApi`. There is no alternative: `browser.downloads.download()` cannot resolve a page `blob:` URL from bg either. `data:` URIs, the case #2605 actually reported, still route through the browser API.

### DIFF
--- a/src/background/utils/requests.js
+++ b/src/background/utils/requests.js
@@ -1,5 +1,6 @@
 import { keepAlive, noop, sendTabCmd, string2uint8array } from '@/common';
 import { CHARSET_UTF8, FORM_URLENCODED, UA_PROPS, UPLOAD } from '@/common/consts';
+import { downloadBlob } from '@/common/download';
 import { deepEqual, forEachEntry, forEachValue } from '@/common/object';
 import { kGmDownloadViaApi } from '@/common/options-defaults';
 import { initXHR } from '@/offscreen/xhr';
@@ -69,6 +70,8 @@ addPublicCommands({
     )(opts, events, id, req, src, fileName).catch(cbError);
   },
   AbortRequest: abortRequest,
+  /** Used by injected/content for a page's `blob:` URL, which only it can read */
+  DownloadBlob: __.MV3 ? callOffscreen : args => downloadBlob(args[0], args[1]),
   // TODO: check if the content script can revoke it
   RevokeBlob: __.MV3 ? callOffscreen : URL.revokeObjectURL,
 });
--- a/src/injected/content/requests.js
+++ b/src/injected/content/requests.js
@@ -1,10 +1,11 @@
 import { U8_fromBase64, UA_PROPS, UPLOAD } from '../util';
 import * as bridge from './bridge';
-import { makeSafeBlob, sendCmd } from './util';
+import { createObjectURL, makeSafeBlob, sendCmd } from './util';
 
 const CHUNKS = 'chunks';
 const LOAD = 'load';
 const LOADEND = 'loadend';
+const READYSTATECHANGE = 'readystatechange';
 /** @type {{ [id:string]: GMReq.Content }} */
 const requests = createNullObj();
 let BlobProto, getArrayBuffer, getBlob, getBlobType, getTypedArrayBuffer;
@@ -58,6 +59,10 @@ bridge.addHandlers({
   async HttpRequest(msg, realm) {
     if (IS_FIREFOX) msg = nullObjFrom(msg); // copying into our realm to set its props freely
     else setPrototypeOf(msg, null);
+    const fileName = msg[kFileName];
+    if (fileName && msg.url::slice(0, 5) === 'blob:') {
+      return downloadPageBlob(msg, fileName, realm);
+    }
     const data = !IS_FIREFOX && msg.data;
     const uaData = getUAData && navigator::getUAData();
     requests[msg.id] = {
@@ -144,6 +149,49 @@ bridge.addBackgroundHandlers({
 function sendHttpRequested(msg, realm) {
   bridge.post('HttpRequested', msg, realm);
 }
+
+/**
+ * A `blob:` URL is only valid in the context that created it, so a page's blob can't be
+ * re-fetched in bg/offscreen (Firefox fails it with status 0). Only a content script can
+ * read it, hence this shortcut for `GM_download` of such URLs.
+ * @param {GMReq.Message.Web} msg
+ * @param {string} fileName
+ * @param {VMScriptInjectInto} realm
+ * @return {Promise<void>}
+ */
+async function downloadPageBlob(msg, fileName, realm) {
+  const { id, url } = msg;
+  const events = msg.events[0];
+  let blob;
+  try {
+    blob = await importBlob(url, true);
+  } catch (err) {
+    sendHttpRequested({
+      id,
+      type: ERROR,
+      data: null,
+      [ERROR]: [err && err.message || `${err}`, err && err.name],
+    }, realm);
+    sendHttpRequested({ id, type: LOADEND, data: makeVirtualResponse(url, 0) }, realm);
+    return;
+  }
+  // download in bg to a) circumvent CSP in Firefox and b) use a single throttled download chain
+  sendCmd('DownloadBlob', [IS_FIREFOX ? blob : createObjectURL(blob), fileName]);
+  for (const type of [READYSTATECHANGE, LOAD, LOADEND]) {
+    if (!(type === LOADEND/*to forget the request*/ || events[type]))
+      continue;
+    sendHttpRequested({ id, type, data: makeVirtualResponse(url, 200) }, realm);
+  }
+}
+
+/** `response` is null to match the browser downloads API mode and the spec, see db1b5c3e */
+const makeVirtualResponse = (url, status) => ({
+  finalUrl: url,
+  readyState: 4,
+  status,
+  [kResponse]: null,
+  [kResponseHeaders]: '',
+});
 
 /**
  * Only a content script can read blobs from an extension:// URL
--- a/src/offscreen/index.js
+++ b/src/offscreen/index.js
@@ -1,5 +1,6 @@
 import { leaseBlobUrl } from '@/common';
 import setClipboard from '@/common/clipboard';
+import { downloadBlob } from '@/common/download';
 import handlers from '@/common/handlers';
 import { sendCmdToSW } from '@/common/messaging-sw';
 import { initXHR, xhrs } from './xhr';
@@ -7,6 +8,7 @@ import { initXHR, xhrs } from './xhr';
 let autoCloseTimer;
 
 Object.assign(handlers, {
+  DownloadBlob: args => downloadBlob(args[0], args[1]),
   LeaseBlob: leaseBlobUrl,
   async Fetch([url, init, get = 'text']) {
     const req = await fetch(url, init);


Auxority:
Superseded by 2142cbcc, which also fixes GM_xmlhttpRequest on blob: URLs.

Issue #2403, opened 2025-12-21
[Feature] Add Function to add new domain easily


BlazeFTL:
**Motivation**
now a days link shortners rotate domains every 1-2 days to evade adblock or bypass and adding domains for them is very time consuming

**Proposed Solution**
1.Lets say i am on xyz.com page if i open violent money popup there should be a dedicated button(a plus(+) button) to add a new domain to active or disabled Userscripts or another button that would automatically detect domain and add the domain to that Userscript

2.A domain add button should be present in settings Userscript list so that i can just paste a domain that i copied without going inside code

3 // @match        https://*.xyz.com/* In code editor writing match even with auto complete is a pain(which doesn't reliably work in mobile phone either)
Why not let user type multiple domains just with a "," comma ?
Like -
 // @match        xyz.com,zyx.com,xzy.* 
Get rid of https://*. Part something similar is done for uBlockOrigin too as it just a extra hassle

**Use Cases**
Explained in point 1

Thank you for service nonetheless ❤️

BlazeFTL:
@tophf Should i make a PR for this ?..
I dont mind at all if you dont want it implemented currently...
Explained here https://github.com/BlazeFTL/ViolentMonkey_AddDomain

https://github.com/BlazeFTL/ViolentMonkey_AddDomain/actions/runs/29812937522/artifacts/8488175481

tophf:
Your UI looks neat but I'm not sure `+` should be in the script row and not inside the `...` menu.
@gera2ld?

gera2ld:
I think we shouldn't modify the original script; instead, we should simply modify `script.custom.match`. WDYT?
BTW, I also prefer the button inside `...` as there seems to be a lot on the small page.
Anyway, the idea looks good to me.

BlazeFTL:
Those sites that change domains.. they change domains every 1-2 days...this is specifically for someone who maintains or creates a Userscript not for a user of UserScript 

So for script.custom.match sharing/updating the script elsewhere won't carry the added domain.

For the button — Adding a domain is a frequent action burying it in an overflow menu adds a click and defeats the "quick add" purpose.

<details><summary>Details</summary>
<p>

<img width="702" height="1560" alt="Image" src="https://github.com/user-attachments/assets/9ae1e362-2461-40b3-8d72-458c6b021744" />

</p>
</details>

tophf:
[...]

PR #2628, opened 2026-08-04
Implement Domain Adding Button

BlazeFTL:
Implements Features From https://github.com/violentmonkey/violentmonkey/issues/2403#issuecomment-5177916349

I did hide the + button in popup ui by default and i would use
`.vm-add-match-btn { display: block !important; }` In Custom css..i hope that you guys will accept that🥹...if not either i can drop the + button in popup or make a entry in settings to show or hide it (both branch ready too)


tophf:
Wow, that's a lot of extra code, 10 times more than I expected based on the existing code for excludes, and my first reaction is that the vast majority of it is unnecessary, like toasts and other tweaks, I'll review in detail tomorrow. Ideally the feature should reuse the existing design and functionality, ideally the same `<div v-if="item.excludes"` block should be generalized to support both.

BlazeFTL:
About reusing existing design and functionality - excludes and match aren't the same data model. excludes writes to script.custom.exclude (VM settings, no source touch). Match-add edits the actual `@match` line in script metadata + runs expandMatchShorthand() on save. So the v-if block can maybe share layout/CSS, but save logic can't really be merged without forcing an awkward abstraction over two different write paths.

On diff size  most of the bulk is the popup+dashboard button repeated across two surfaces ( + different ui for table view) , plus the shorthand-expansion logic and its tests...

tophf:
Oops, the feature I and @gera2ld approved was adding the custom includes to `custom` object, not to the code of the script, but AFAICT it doesn't interest you that much so I guess we'll implement it ourselves.

BlazeFTL:
That said, if it still feels too big or out of scope for the project after review, I get it if you'd rather not merge it...i am more than happy with keeping a fork(its just that i have to click a few buttons to get updates 🤡..i am lazy)

BlazeFTL:
Both has its own positive and negative
The scope here is UserScript maintainer vs user

My reasoning for editing `@match` was portability: it survives export/import and moving to another userscript manager, since it's baked into the script source. 

But that only holds for your own scripts. For scripts installed from someone else's URL, an auto-update overwrites the source and wipes any added `@match` lines — custom.include survives that since it's stored separately from the script body.

tophf:
I'd rather add a new export mode to bake the custom matches/includes.

BlazeFTL:
How about this?
1. `+` In Popup is hidden by default..shown with a settings entry
2. Where the domain goes is decided from settings itself
Screenshot_20260806-101206_Spark Launcher

tophf:
No, I think it's too niche to justify adding all that code and Violentmonkey tries to be minimalistic conceptually. A more generally useful solution would be to add a function to apply the custom lists to the source code during export and/or a corresponding button in the editor's settings tab.

BlazeFTL:
I also think its too much code to merge and we shouldn't move forward..thank you anyway ❤️

PR #2672, opened 2026-09-11
Add Git-based sync (GitHub/Gitea) as an alternative to Google Drive

### LINKED ISSUE #2176: [Feature] Sync to a repository on GitHub and/or other `git` servers
**Motivation**

Currently there are several different options for syncing scripts. However, GitHub (or Gitlab, BitBucket, etc.) is not one of those options.

**Proposed Solution**

I propose adding the option to sync/backup scripts with a one of these platofrms, either with a source code repository, or even the option to sync with GitHub gist would suffice.

**Use Cases**

I think this would benefit both users and userscript developers alike.  Developers who publish their scripts undoubtedly push changes to a repository such as GitHub anyway, so it logically follows that GitHub would be a convenient option for developers.  Users, I'm sure, would not be against at the very least having the option to sync with a free platform as well (of course, vm provides alternatives already, though none specifically code-oriented).

CrazyWolf13:
Closes #2176.

That issue's been open for over a year and I finally got tired of needing a Google account just to sync my userscripts. This adds a sync provider that works against GitHub, GitHub Enterprise Server, Gitea, or Forgejo — anywhere that speaks the same "contents API" GitHub does. You point it at an existing repo + a personal access token and it syncs scripts as files, one commit per change.

A couple of deliberate scope calls:
- The repo has to already exist, this won't auto-create one for you, just fails clearly if it can't find it or the token can't see it.
- GitLab is not covered. Its API shape is different enough (separate tree-listing endpoint, different auth header) that it'd need its own provider rather than fitting into this one. Happy to see that added separately if someone wants it.

The drive-provider part of this lives in `usync-dev/usync` (npm scope `@usync/*`) rather than in this repo, so it's split across two PRs — this one wires it up on the Violentmonkey side, the other adds the actual `GitContents` adapter. Companion PR: usync-dev/usync#`<fill in once opened>`.

> [!NOTE]
> Full disclosure: I built this with a lot of help from Claude (Anthropic's AI) — most of the implementation was AI-generated, with me steering the design decisions and doing the actual testing. I'm not comfortable claiming I could've written all of this myself at this quality in a reasonable amount of time, and I'd rather say that upfront than have it come out in review.

> [!WARNING]
> Given the above, please review this more carefully than you normally would a human-written PR. I've tested it end-to-end (see below) and it works for my use case, but I may have missed edge cases an experienced maintainer would catch on sight.

## Testing

I've run this against a real GitHub repo: initial sync, editing a script and re-syncing, and a repo-not-found error case. All working.

<details>
<summary>How to test this yourself</summary>

This needs both repos since the sync-engine change here depends on the new provider in `usync`.

```bash
git clone https://github.com/CrazyWolf13/usync.git
cd usync
git checkout feat/gitcontents-sync-provider
pnpm install
pnpm --filter @usync/drive build

cd ..
git clone https://github.com/CrazyWolf13/violentmonkey.git
cd violentmonkey
git checkout feat/git-sync-gitcontents
pnpm install
```

Link the local `@usync/drive` build into Violentmonkey (needed since the package isn't published to npm yet) by adding this to `violentmonkey/pnpm-workspace.yaml` under its `overrides:` key:

```yaml
overrides:
  '@usync/drive': 'link:../usync/packages/drive'
```

Then:

```bash
pnpm install
pnpm build:mv3
```

Load `dist-mv3/` as an unpacked extension in Chrome (`chrome://extensions` → Developer mode → Load unpacked), or `dist/` in Firefox via `about:debugging` → Load Temporary Add-on.

In the sync settings, pick "GitHub / Gitea", point it at a repo you own with a PAT that has contents read/write access, and try a sync.

</details>

Let me know if you'd rather see this split differently or if there's a preferred pattern for host-configurable sync providers I should follow instead. I would really really love to get this merged, as I think a lot of technical folks would enjoy using a git repo for their userscripts.


gera2ld:
Tested locally against GitHub — sync works. Thanks for the contribution!

One required fix and two UX suggestions:

1. **Bump `@usync/drive` to `^0.3.1`** (package.json + pnpm-lock.yaml) — the code relies on newer APIs than the declared `^0.2.4` provides. I verified it works with 0.3.1.

2. **Rename the display name "GitHub / Gitea" to something more general.** The feature covers GitHub, GHES, Gitea and Forgejo, and the list will only grow. Note the transport is actually the GitHub "Repository Contents" REST API, not git itself, so a name like **"GitHub-compatible API"** is both host-agnostic and technically accurate (cf. the existing "S3 Compatible" naming). Host examples can live in the tooltip/placeholder.

3. **"Host type" is confusing** — it implies picking a product name, while it really only selects the request style used to *create* a file (updates use the same call on all hosts). Suggest:
   - Label: **"File creation method"** (matching the `createMethod` key)
   - Options name the method first, hosts as examples: `PUT — GitHub / GHES` (default) and `POST — Gitea / Forgejo`
   - Tooltip: *"If sync fails when creating the file for the first time, try the other method."*

CrazyWolf13:
Thanks, addressed the review feedback:

Bumped @usync/drive to ^0.3.1 in package.json/pnpm-lock.yaml
Renamed display name to "GitHub Compatible" 
Renamed "Host type" → "File creation method"; options now lead with the HTTP method.

One thing from the lockfile update: 0.3.1 pulls in @usync/util and isomorphic-git as new dependencies of @usync/drive. They shouldn't land in the built extension, right?

All tests pass locally

gera2ld (approved):
LGTM. Yes, isomorphic-git will not be imported. Thank you!

